PR #6, opened 2025-01-15
Draft request for new pytorch work

ccummingsNV:


tunabrain:
My main comment after having a read through is that a few features from the original were stripped out during the refactor, some critical. I don't if the plan is to readd them before merging, but I'll flag the ones I've spotted and the original motivation for including them:

- In `TorchAutoGradFunction`, `unpacked_args` / `unpacked_kwargs` are not purged of torch.Tensors before saving them in the context. This will leak GPU memory via a reference cycle (`torch.autograd.Function` -> `ctx` -> `unpacked_args` -> `torch.Tensor` -> `torch.autograd.Function` via `grad_fn`) that won't be cleaned up until the next gc pass. In a training loop you either run out of GPU memory or you have to manually call `gc.collect` every loop which slows things to a crawl :( . The current torch.py filters args so no tensors are referenced and stores them in `ctx.save_for_backward` only (internally, torch temporarily un-pythons tensors to be recreated later), remembers their paths and rebuilds the original args in backward()

- The buffer copy logic is removed and relies on SGLs import mechanism instead. This has a few issues: SGL ignores the strides of an ndarray, and just imports a contiguous segment of num_elements * element_size. If a tensor is non-contiguous (e.g. due to slicing) this will import the wrong memory segment and cause silent bugs, and it will even crash if the shape of a tensor is larger than its underlying memory (e.g. `a = torch.randn(512, 1); a = a.expand(512, 512);`). The safe solution is to either make tensors contiguous before passing to SGL or to copy the entire storage backing a tensor, but both have problems. Making a tensor that has been expanded contiguous dramatically inflates the memory it takes up and produces different outputs; and copying the whole storage is hugely inefficient if the tensor is a smaller view of a larger tensor. Both cases are pretty common, so it's a tough spot. The solution the current torch code takes is to first compute whether the viewable memory segment or the contiguous tensor size is smaller, then adjusts the internal strides to accomodate and copies the smallest amount needed. Long-term SGL needs a fix for the strides, but fundamentally it can't do any non-contiguous copies since this affects addressing outside of SGLs control. I've simplified the logic recently and it's waiting in an internal branch for the torch refactors.

- Out/inout values are gone and only a return value is supported now. How important that is depends; if you write new code to interface with torch, you will be able to avoid these (though multiple return values are harder), but with slangpy we promise that you can take your existing slang codebase and interface with python (and torch!) immediately, so users might might not always have the choice to write it correctly from scratch :) In-place operations are conditionally allowed in torch (and useful sometimes). 

- The cuda stream syncing is gone and now just syncs on the default cuda stream. I honestly don't know how common the non-default cuda stream is (probably just a multi-gpu thing and even then I'm not sure), but sync bugs are a bad time for end users, so better safe than sorry :)

The memory leak and SGL buffer copies are serious and would be a show stopper. The cuda streams are IMO not critical, but it looks like a low effort re-add (the collect_streams method is still around). Out/inout, depends. Beyond how important these are though, they were already implemented, merged and took a long time to figure out. I would push against removing them for a refactor, since we'll have to go through the same effort again if we want to readd them in the future.

Some other comments that are less critical:
- I don't know if making every `TorchFunction` a `torch.nn.Module` is a good idea. The purpose of a `torch.nn.Module` is that it tracks trainable parameters, which `TorchFunctions` won't have. Is the motivation to make it easier to turn functions into modules for using them with e.g. `torch.nn.Sequential`? We could add a trivial wrapper class for this like the `Lambda` class in this example: https://pytorch.org/tutorials/beginner/nn_tutorial.html#using-nn-sequential But generally this would also be easy to do for users
- Bigger picture comment---this is true for the current torch integration also---it feels unsatisfying that we need to rewrap every importable class to support torch. From the user's perspective, tensors are just another python class to pass to a slangpy function, but it needs rewrapping of every imported module. It's easy to forget and feels like an unnecessary nuisance. Internally, fixing this would involve either turning all modules into TorchModules if the torch extension is loaded, or automatically wrapping a regular Function if a torch.Tensor appears in the arguments during unpacking.
- I might have missed it, but it looks like right now slangpy will throw because 'torch.Tensor' is an unsupported python class if you accidentally pass it to a non-TorchFunction. Currently we catch that case and throw an exception to remind users to wrap in a `TorchFunction` first
- I think we might be able to get rid of a lot of complexity, since (without hooks) we can now modify the arguments. The original reason to have a `WrappedTensor` was that we needed to store additional information that wasn't available in the `torch.Tensor`. However, I think we can now just transform every `torch.Tensor` argument into a slangpy.`Tensor` and delete `wrappedtensor.py`. In the current implementation it tries to reuse a lot of the slangpy.`Tensor` logic anyway (though it's a struggle since we need to remember external state). This is not user-visible so this can be part of a future refactor

PR #9, opened 2025-01-23
Add diff-splatting example, fix issue with latest slang version, update group size for sample

saipraveenb25:


ccummingsNV:
Aside from comment about threads, code looks generally ok.

I'd like to ship examples with some documentation explaining what they do and what the user should learn from it. These are in the docs folder, and get built automatically on merge.

bmillsNV (commented):
LGTM

bmillsNV (approved):

PR #11, opened 2025-01-23
Tensor cleanup

tunabrain:
* Make tensor shape part of the interface
* Fix failing tensor test by tracking shape when slicing
* Remove workarounds for [ForceUnroll], extension
* Remove __subscript from interface as it triggers compiler corner cases and move it into extension instead

### DIFF
--- a/slangpy/builtin/tensor.py
+++ b/slangpy/builtin/tensor.py
@@ -209,7 +209,7 @@ class TensorMarshall(Marshall):
         primal_calldata = {
             'buffer': data.storage,
             'layout': {'offset': data.offset, 'strides': strides},
-            'shape': data.shape
+            '_shape': data.shape
         }
 
         if not self.d_in and not self.d_out:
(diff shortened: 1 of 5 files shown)


ccummingsNV:
Code in this looks valid (I think there's a general issue with linux CI crashing atm that isn't related). However the 'shape' variable looks a bit odd. I can't see where it's actually used in the shaders - is it needed? I ask because it's going to introduce real overhead both CPU and GPU side.

tunabrain:
Yes, it's a valid comment re: shape. I'm not sure myself.
It's not used currently by us, but it's a reasonable thing people would expect if they want to write less specialized code - the equivalent of `GetDimensions` for textures.

As you say, we can provide two separate tensor implementations, one storing shape and one not, and select one based on if the tensor is user visible or not. This could be done at a later point as a performance optimization and wouldn't affect end user code.

Alternatively we can delete it from the interface now, and bring it back if people ask for it later.

Deleting it or leaving it as is are both easy to do right now. Separate tensor implementations would likely have to wait a bit. I don't feel strongly either way.

PR #52, opened 2025-02-07
Signed distance field example

expipiplus1:


ccummingsNV:
Example looks really nice. I think to put it up as an example, the only change I'd make is to remove the use of cv2/PIL and use SGL's texture loading function. i.e. wherever possible we don't want examples to have dependencies.

If dependencies are a must, then we need a requirements.txt in there, and a note to run 'pip install -r requirements.txt' in the readme.md. But not having them is definitely preferrable.

PR #57, opened 2025-02-12
Swap struct/function lookup order

tunabrain:
This works around a slang bug that will allow looking up a function as a type if the function is generic with a type constraint. Trying to work with such functions will fail right now, since slangpy treats the function as a struct and subsequent operations on it fail.


ccummingsNV:
LGTM - merged.

PR #58, opened 2025-02-12
Fixes to hot reload

tunabrain:
This fixes a bunch of places in the codebase that break with hot reload.

Generally, we can't store raw TypeReflection/FunctionReflections, since those will go stale under reload. Instead, we have to use SlangFunction/SlangType, which will get updated during the hot reload hook.

This patch is being conservative and clears all caches during a hot reload. Some cached data might be salvageable, but I saw weird behavior when caches were left untouched, and it's better to take the performance hit on a reload which is already expensive.

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -106,17 +106,17 @@ class CallData(NativeCallData):
 
             # Perform specialization to get a concrete function reflection
             slang_function = specialize(
-                context, bindings, build_info.reflections, build_info.type_reflection)
+                context, bindings, build_info.functions, build_info.this_type)
             if isinstance(slang_function, MismatchReason):
                 raise ResolveException(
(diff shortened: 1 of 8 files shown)


tunabrain:
Yeah, this admittedly happens rarely, but this was to deal with the case when a hot reload would occur in the middle of the calldata constructor, which invalidates the reflection data stored in the build_info and you get an annoying error.
Unless you think SlangType/SlangFunction is a performance concern, I'd advocate for not storing raw SGL reflection to avoid hot reload bugs.

The current MR uses a pre-execution hook, but during development I was using an asynchronous reload loop instead that would check for file changes, and it could preempt at any point. Technically, any of the marshalls that get queried in the calldata constructor could call into SGL and unintentionally cause a cleanup as well, especially once we switch to automatic SGL garbage collection.
I don't think this can happen in the current codebase, but if it doesn't cost us much, it doesn't hurt to be defensive.

PR #59, opened 2025-02-12
Add Jupyter integration

tunabrain:
This patch adds Jupyter integration hooks that allow pretty printing of all SlangPy types, graphical display of bitmaps/textures, and auto-reload of shaders.

Jupyter presence is detected during device creation, and the hooks are installed if Jupyter is present. A light-weight check is used here that just peeks at the builtins that are injected; once we know Jupyter is available, it's safe to import `IPython`.


tunabrain:
Yes, __repr__ for all the types should be added, that will be a separate MR. I'm planning to copy SGLs existing model for __repr__. Unclear if we should add a pretty-print __str__ as well, that seems to be the Python way. SGL is verbose by default, but a short version for print() might be useful.

Either way, a dedicated Jupyter pretty print is useful to have for a bit of extra QOL like proper line breaks. There's some low-hanging fruit creature features we can add in the future, like turning the file paths into clickable links etc.

PR #101, opened 2025-03-04
Better errors for missing functions + made struct look like slang type

ccummingsNV:
Improved some error reporting, and added properties to Struct so it can duck type with SlangType. 

Main contentious change in terms of old code is that Struct.name now returns the same value as SlangType.name, where previously it returned SlangType.full_name.


tunabrain:
I think that's fine! I doubt there's much code relying on this, and it's an easy fix to use full_name instead.

PR #116, opened 2025-03-14
metal support

kaizhangNV:
Add support for Metal platform.

1. Change the entry point name from main->computeMain, because Metal doesn't support "main" as entry point name.
2. Change the way of copying data between GPU and CPU for structured data, because Metal has different alignment requirement then glsl/hlsl.
3. Change texture format from rgb to rgba because metal doesn't support rgb format texture.

### DIFF
--- a/examples/broadcasting/main.py
+++ b/examples/broadcasting/main.py
@@ -91,9 +91,9 @@ print("")
 
 # Create a sampler and texture
 sampler = device.create_sampler()
-tex = device.create_texture(width=32, height=32, format=sgl.Format.rgb32_float,
+tex = device.create_texture(width=32, height=32, format=sgl.Format.rgba32_float,
                             usage=sgl.ResourceUsage.shader_resource)
-tex.copy_from_numpy(np.random.rand(32, 32, 3).astype(np.float32))
+tex.copy_from_numpy(np.random.rand(32, 32, 4).astype(np.float32))
 
 # Sample the texture at a single UV coordinate. Results in 1 thread,
 # as the uv coordinate input is a single float 2.
--- a/examples/buffers/main.py
+++ b/examples/buffers/main.py
@@ -47,5 +47,5 @@ for x in range(16):
 
 # Or if installed, we can use tev to show the result (https://github.com/Tom94/tev)
 tex = device.create_texture(data=result.to_numpy(), width=16,
-                            height=16, format=sgl.Format.rgb32_float)
+                            height=16, format=sgl.Format.rgba32_float)
 sgl.tev.show(tex)
--- a/examples/type_methods/main.py
+++ b/examples/type_methods/main.py
@@ -31,7 +31,7 @@ result_cursor = particles.buffer.cursor()
 
 for i in range(particles.shape[0]):
     instance = result_cursor[i].read()
-    print('{} {}'.format(instance['position'], instance['velocity']))
+    print(f"{instance['position']} {instance['velocity']}")
 
 # Update the particles
 particles.update(0.1)
@@ -40,4 +40,4 @@ result_cursor = particles.buffer.cursor()
 # Print all the particles by breaking them down into groups of 6 floats
 for i in range(particles.shape[0]):
     instance = result_cursor[i].read()
-    print('{} {}'.format(instance['position'], instance['velocity']))
+    print(f"{instance['position']} {instance['velocity']}")
--- a/examples/type_methods/main.py
+++ b/examples/type_methods/main.py
@@ -27,10 +27,17 @@ particles.construct(
 )
 
 # Print all the particles by breaking them down into groups of 6 floats
-print(particles.to_numpy().view(dtype=np.float32).reshape(-1, 6))
+result_cursor = particles.buffer.cursor()
+
+for i in range(particles.shape[0]):
+    instance = result_cursor[i].read()
+    print('{} {}'.format(instance['position'], instance['velocity']))
 
 # Update the particles
 particles.update(0.1)
 
+result_cursor = particles.buffer.cursor()
 # Print all the particles by breaking them down into groups of 6 floats
-print(particles.to_numpy().view(dtype=np.float32).reshape(-1, 6))
+for i in range(particles.shape[0]):
+    instance = result_cursor[i].read()
+    print('{} {}'.format(instance['position'], instance['velocity']))
--- a/slangpy/core/utils.py
+++ b/slangpy/core/utils.py
@@ -8,7 +8,7 @@ from slangpy.backend import (DeclReflection, ProgramLayout,
                              DeviceType, Device)
 from slangpy.reflection import SlangType
 import builtins
-
+import sys
 
 def create_device(type: DeviceType = DeviceType.automatic, enable_debug_layers: bool = False,
                   adapter_luid: Optional[Sequence[int]] = None,
@@ -28,7 +28,8 @@ def create_device(type: DeviceType = DeviceType.automatic, enable_debug_layers: 
                 shaderpath,
             ]+list(include_paths),
         },
-        enable_cuda_interop=True,
+
+        enable_cuda_interop=sys.platform != "darwin",  # Disable CUDA interop on macOS
         enable_debug_layers=enable_debug_layers,
         adapter_luid=adapter_luid)
 
--- a/slangpy/core/utils.py
+++ b/slangpy/core/utils.py
@@ -12,7 +12,8 @@ import sys
 
 def create_device(type: DeviceType = DeviceType.automatic, enable_debug_layers: bool = False,
                   adapter_luid: Optional[Sequence[int]] = None,
-                  include_paths: Sequence[Union[str, PathLike]] = []):
+                  include_paths: Sequence[Union[str, PathLike]] = [],
+                  enable_cuda_interop: bool = False):
     """
     Create an SGL device with basic settings for SlangPy. For full control over device init, 
     use sgl.create_device directly, being sure to add slangpy.SHADER_PATH
@@ -29,7 +30,6 @@ def create_device(type: DeviceType = DeviceType.automatic, enable_debug_layers: 
             ]+list(include_paths),
         },
 
-        enable_cuda_interop=sys.platform != "darwin",  # Disable CUDA interop on macOS
         enable_debug_layers=enable_debug_layers,
         adapter_luid=adapter_luid)
 
--- a/slangpy/tests/generated_tests/polynomial_soa.slang
+++ b/slangpy/tests/generated_tests/polynomial_soa.slang
@@ -102,7 +102,7 @@ void _trampoline(out vector<float,3> _result, in vector<float,3> a, in vector<fl
 
 [shader("compute")]
 [numthreads(32, 1, 1)]
-void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
+void compute_main(uint3 dispatchThreadID: SV_DispatchThreadID)
 {
     if (any(dispatchThreadID >= call_data._thread_count)) return;
     int[1] call_id;
--- a/slangpy/tests/generated_tests/polynomial_soa.slang
+++ b/slangpy/tests/generated_tests/polynomial_soa.slang
@@ -102,7 +102,7 @@ void _trampoline(out vector<float,3> _result, in vector<float,3> a, in vector<fl
 
 [shader("compute")]
 [numthreads(32, 1, 1)]
-void main(uint3 dispatchThreadID: SV_DispatchThreadID)
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
 {
     if (any(dispatchThreadID >= call_data._thread_count)) return;
     int[1] call_id;
--- a/slangpy/tests/helpers.py
+++ b/slangpy/tests/helpers.py
@@ -7,13 +7,15 @@ from pathlib import Path
 from typing import Any
 
 import pytest
+import numpy as np
 
 from slangpy.core.calldata import SLANG_PATH
 
 import slangpy
 from slangpy import Module
 from slangpy.backend import (Device, DeviceType, SlangCompilerOptions,
-                             SlangDebugInfoLevel)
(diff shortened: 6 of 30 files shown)


slangpy/tests/test_differential_function_call.py
```diff
@@ -198,30 +198,32 @@ def test_vec3_call_with_buffers(device_type: DeviceType):
...
 
     res = res.with_grads()
-    res.grad.storage.copy_from_numpy(np.ones(32*3, dtype=np.float32))
+    helpers.write_ndbuffer_from_numpy(res.grad, np.ones(32*3).astype(np.float32), 3)
 
     kernel_eval_polynomial.bwds(a, b, res)
-    a_grad_data = a.grad.storage.to_numpy().view(np.float32).reshape(-1, 3)
-    b_grad_data = b.grad.storage.to_numpy().view(np.float32).reshape(-1, 3)
+
+    # TODO: Somehow the grads are not layout the same as input tensors.
```
ccummingsNV: Is this a new issue? We should get to the bottom of this and fix it before merging the PR if so. TODOs are fine if we know why they're happening, not if it's a bug we don't understand.

examples/broadcasting/main.py
```diff
@@ -91,9 +91,9 @@
 
 # Create a sampler and texture
 sampler = device.create_sampler()
-tex = device.create_texture(width=32, height=32, format=sgl.Format.rgb32_float,
+tex = device.create_texture(width=32, height=32, format=sgl.Format.rgba32_float,
```
ccummingsNV, line 94: Why the change from rgb32?

examples/buffers/main.py
```diff
@@ -47,5 +47,5 @@
 
 # Or if installed, we can use tev to show the result (https://github.com/Tom94/tev)
 tex = device.create_texture(data=result.to_numpy(), width=16,
-                            height=16, format=sgl.Format.rgb32_float)
+                            height=16, format=sgl.Format.rgba32_float)
```
ccummingsNV, line 50: Same question

slangpy/tests/generated_tests/polynomial_soa.slang
```diff
@@ -102,7 +102,7 @@ void _trampoline(out vector<float,3> _result, in vector<float,3> a, in vector<fl
 
 [shader("compute")]
 [numthreads(32, 1, 1)]
-void main(uint3 dispatchThreadID: SV_DispatchThreadID)
+void computeMain(uint3 dispatchThreadID: SV_DispatchThreadID)
```
ccummingsNV: SlangPy is following the SGL/python naming conventions, so this should be 'compute_main'

slangpy/tests/test_intances.py
```diff
@@ -37,6 +37,17 @@ def get_this(self) -> Any:
     def update_this(self, value: Any) -> None:
         self.update_called += 1
 
+def faltten_ndarray(data: np.ndarray) -> np.ndarray:
```
ccummingsNV: Typo - this should be 'flatten'

slangpy/tests/helpers.py
```diff
@@ -105,3 +107,30 @@ def create_function_from_module(
     if function is None:
         raise ValueError(f"Could not find function {func_name}")
     return function
+
+def read_ndbuffer_from_numpy(buffer: NDBuffer) -> np.ndarray:
```
ccummingsNV, line 111: I see the tests are now using this across the board. What's the reason for switching from just using the numpy array? Is it the alignment issue we discussed on slack? (if so, fair enough!)

ccummingsNV: We should add a function to read/write all elements at some point - no need to worry about it for now though.

slangpy/core/utils.py
```diff
@@ -28,7 +28,8 @@ def create_device(type: DeviceType = DeviceType.automatic, enable_debug_layers:
                 shaderpath,
             ]+list(include_paths),
         },
-        enable_cuda_interop=True,
+
+        enable_cuda_interop=sys.platform != "darwin",  # Disable CUDA interop on macOS
```
ccummingsNV: Suggest we pass in this as a bool that defaults to false and add it to release notes - forcing cuda by default has caused too many headaches.

slangpy/tests/test_differential_function_call.py
```diff
@@ -293,8 +295,11 @@ def test_vec3_call_with_buffers_soa(device_type: DeviceType):
     a_x_grad_data = a_x.grad.storage.to_numpy().view(np.float32).reshape(-1, 1)
     a_y_grad_data = a_y.grad.storage.to_numpy().view(np.float32).reshape(-1, 1)
     a_z_grad_data = a_z.grad.storage.to_numpy().view(np.float32).reshape(-1, 1)
+
     a_grad_data = np.column_stack((a_x_grad_data, a_y_grad_data, a_z_grad_data))
-    b_grad_data = b.grad.storage.to_numpy().view(np.float32).reshape(-1, 3)
+
+    # TODO: Somehow the grads are not layout the same as input tensors.
```
ccummingsNV: Same as above - we should work out and address this before merging.

slangpy/tests/test_intances.py
```diff
@@ -59,14 +70,20 @@ def test_this_interface(device_type: DeviceType):
...
+    data = helpers.read_ndbuffer_from_numpy(buffer)
+
     # position
-    assert data[0] == 1.0
-    assert data[1] == 2.0
-    # direction
-    assert data[2] == 3.0
-    assert data[3] == 4.0
+    positions = np.array([item['position'] for item in data])
+    assert np.allclose(positions, [1.0, 2.0])
```
ccummingsNV: Shouldn't these be np.all(positions==bla)? i.e. we expect them to be identical, not close.

slangpy/tests/test_pytorch.py
```diff
@@ -5,12 +5,20 @@
...
 import os
+import sys
 
 try:
     import torch
 except ImportError:
     pytest.skip("Pytorch not installed", allow_module_level=True)
 
+# Skip all tests in this file if running on MacOS
+@pytest.fixture(autouse=True)
```
ccummingsNV: Shouldn't this just be a module level skip (see line 13)? It shouldn't need to be a fixture unless you want logic to run that changes per test.

slangpy/tests/test_tensor_with_grads.py
```diff
@@ -6,6 +6,13 @@
 import numpy as np
 from typing import Any
 import os
+import sys
+
+
+@pytest.fixture(autouse=True)
```
ccummingsNV: See comments in torch file - this should be a module level skip

slangpy/tests/test_textures.py
```diff
@@ -9,6 +9,12 @@
 from slangpy.reflection import ScalarType
 from slangpy.builtin.texture import SCALARTYPE_TO_TEXTURE_FORMAT
 from slangpy.types.tensor import _slang_to_numpy
+import sys
+
+@pytest.fixture(autouse=True)
```
ccummingsNV: Module level skip again

slangpy/tests/test_textures.py
```diff
@@ -9,6 +9,12 @@
...
+def skip_for_macos():
+    if sys.platform == "darwin":
+        pytest.skip("Skipping on macOS: Waiting for slang-gfx fix for resource clear API")
```
ccummingsNV: Is the fix for resource clear coming soon? I don't think we can really say we support metal if we haven't tested textures.

slangpy/tests/test_torchintegration.py
```diff
@@ -23,6 +24,12 @@
     pytest.skip("Skipping pytorch tests as not in D3D", allow_module_level=True)
 DEVICE_TYPES = [DeviceType.d3d12]
 
+# Skip all tests in this file if running on MacOS
+@pytest.fixture(autouse=True)
+def skip_macos():
```
ccummingsNV: Use of fixture instead of module level skip.

examples/type_methods/main.py
```diff
@@ -27,10 +27,17 @@
 )
 
 # Print all the particles by breaking them down into groups of 6 floats
-print(particles.to_numpy().view(dtype=np.float32).reshape(-1, 6))
+result_cursor = particles.buffer.cursor()
```
ccummingsNV, line 30: This will need the documentation updating as well, as it refers to the example code.

examples/type_methods/main.py
```diff
@@ -27,10 +27,17 @@
...
+
+for i in range(particles.shape[0]):
+    instance = result_cursor[i].read()
+    print('{} {}'.format(instance['position'], instance['velocity']))
```
ccummingsNV: In SlangPy we use format strings (eg print(f'{instance['position']}, {instance['velocity']})

examples/type_methods/main.py
```diff
@@ -27,10 +27,17 @@
...
 
 # Update the particles
 particles.update(0.1)
 
+result_cursor = particles.buffer.cursor()
 # Print all the particles by breaking them down into groups of 6 floats
-print(particles.to_numpy().view(dtype=np.float32).reshape(-1, 6))
+for i in range(particles.shape[0]):
+    instance = result_cursor[i].read()
+    print('{} {}'.format(instance['position'], instance['velocity']))
```
ccummingsNV: Format string

examples/buffers/main.py
ccummingsNV: Cancel that question - I just re-read your comments. RGBA is fine.

ccummingsNV (changes requested):
Amazing work getting this all going, but there's a few queries / things to address before we merge it. The biggest concern for me is your TODO about gradient ordering. Gradients are pretty key to SlangPy, so if there's unclear problems popping up, we need to make sure we at least understand them thouroughly, and ideally fix them. I can help dig into these if need be!

slangpy/tests/test_differential_function_call.py
kaizhangNV: Yes, this is a new issue, I will create a task to track this.

examples/broadcasting/main.py
kaizhangNV: Oh, because Metal don't support rgb32 for texture, so I want to make this example runnable on all platforms.

slangpy/tests/generated_tests/polynomial_soa.slang
kaizhangNV: ok, will update all of them.

slangpy/tests/test_differential_function_call.py
kaizhangNV: This one now is tracked by https://github.com/shader-slang/slangpy/issues/118.
And I will fix soon after this PR get merge.

slangpy/tests/test_textures.py
kaizhangNV: Yes, https://github.com/shader-slang/slang/issues/6640 tracks this. It should be small fix to be done in half day. I will work on it after back from GDC.

This PR won't enable Metal until we can make a new release on Slang and SGL. This is just incremental change, so my next change will get smaller.

kaizhangNV: I'd like to embed those github issues in the comment, so will update later.

slangpy/tests/test_differential_function_call.py
ccummingsNV: I would rather this was fixed (or at least understood) before merging into main.

kaizhangNV: I figured out the reason for this issue.

sizeof(float3) is slang is not considering the alignment requirement for Metal, so we evaluate this compile time constant to 12, however, in the reflection API, we do consider the alignment, so the aligned size of float3 is 16.

kaizhangNV: So I think this should be fixed on Slang, it looks like there is nothing we can do here on slangPy.

kaizhangNV: This should not be problem in most of the case, however, we use `RWByteAddressBuffer` for the storage of `AtomicTensor`, that will bring us the problem for calling `atomicAdd`, because it needs byte address to access the data, sizeof(T) will make mistake on calculating the correct address.

ccummingsNV: Ok - good to know - I'm fine with this going in then (though update the comment accordingly so we don't forget). Does this need a slang side fix then (or potentially some more complex slangpy workaround in which we pass a stride in).

ccummingsNV (approved):

slangpy/tests/helpers.py
kaizhangNV: Yes, it's the alignment issue we discussed.

slangpy/tests/test_differential_function_call.py
kaizhangNV: The reason is that we use `RWByteAddressBuffer` in `AtomicTensor`, where the byte address is calculated by `sizeof(T) * index`, which is wrong because `sizeof(T)` is not equal to aligned stride in metal targets. But slang currently doesn't have such intrinsic to calculate the aligned stride. So we will have to keep this disabled, and we will figure out a way to add a intrinsic to calculate the aligned stride for a type.

kaizhangNV: We should provide an intrinsic to calculate the aligned stride, but until then, we probably need to provide the aligned stride for the element type to `AtomicTensor`.

kaizhangNV: sure, will update the comment.

slangpy/tests/test_intances.py
kaizhangNV: done.

slangpy/tests/test_pytorch.py
kaizhangNV: done.

examples/type_methods/main.py
kaizhangNV: done

slangpy/core/utils.py
kaizhangNV: done.

PR #127, opened 2025-04-01
Add a simplified version of the differentiable Gaussian splatting example, for tutorial purposes.

elcuppietan:


elcuppietan:
This should have come from my work account; closing and re-PRing momentarily.

PR #135, opened 2025-04-15
SlangPy NDBuffer/Tensor cleanup 

tunabrain:
This is the SlangPy side of changes for the cleanup of `NDBuffer`/`Tensor`.

Currently, `NDBuffer` and `Tensor` offer similar functionality, but often subtly different, or supported in one but not the other. This PR takes a step towards cleaning this up and aligning the two. This also fixes some latent bugs in the code and adds missing functionality of indexing and views.

This depends on [SGL PR 244](https://github.com/shader-slang/sgl/pull/244)

Overview of changes:

* New `StridedBufferView` base class that represents an SGL buffer with strides/view/offset applied.
* `to_numpy` now behaves identically between `NDBuffer` and Tensor. Before, `NDBuffer.to_numpy` would return a flat buffer of bytes. `Tensor.to_numpy` would try to return a buffer of scalars and return `None` on failure
  
  * The new `to_numpy` always returns a valid ndarray, with behavior depending on the slang dtype.
  * If the dtype is an array/vector (possible nested), its shape is appended to the shape of the `NDBuffer`/`Tensor`
  * If the innermost dtype is a scalar compatible with numpy, it is mapped to the equivalent numpy dtype
  * If the innermost dtype is not a scalar, it is converted to a 1D array of bytes and appended to the result shape
  * Examples:
    `NDBuffer` of dtype `float3` with shape `(4, 5)` -ndarray of dtype `np.float32` with shape `(4, 5, 3)`
    `NDBuffer` of dtype `struct Foo {...}` with shape `(5, )` -ndarray of dtype `np.uint8` with shape `(5, sizeof(Foo))`
  * `to_numpy` respects offsets and strides
* `to_torch` is now supported by both `NDBuffer` and `Tensor`, and its behavior now matches that of `to_numpy` (just in a different framework)
* `NDBuffer`/`Tensor` now support indexing. E.g. for an `NDBuffer` of shape `(64, 32)`, indexing `buffer[4]` will return an `NDBuffer` of shape `(32, )`, representing a 1D view of a slice of the original `NDBuffer`. Partial indexing, slicing and ellipses are supported.
* `Tensor`/`NDBuffer` now have a .view() method for creating new views of the same storage. This can be used to implement currently unimplemented operations
* `broadcast_to` is now supported by NDBuffer as well
* `Tensor.numpy` is deprecated in favor of `Tensor.from_numpy`

### DIFF
--- a/slangpy/slang/core.slang
+++ b/slangpy/slang/core.slang
@@ -260,14 +260,15 @@ public struct RWValueRef<T>
 public struct NDBuffer<T, let N : int>
 {
     public StructuredBuffer<T> buffer;
+    public int offset;
     public int[N] strides;
     public int[N] shape;
 
-    public T get(int[N] index) { return buffer[_idx(index, strides)]; }
-    //public T get(vector<int,N> index) { return buffer[_idx(index, strides)]; }
+    public T get(int[N] index) { return buffer[offset + _idx(index, strides)]; }
+    //public T get(vector<int,N> index) { return buffer[offset + _idx(index, strides)]; }
 
     //public __subscript(int[N] index)->T { get { return get(index); } }
-    public __subscript(vector<int,N> index)->T { get { return buffer[_idx_vec(index, strides)]; } }
+    public __subscript(vector<int,N> index)->T { get { return buffer[offset + _idx_vec(index, strides)]; } }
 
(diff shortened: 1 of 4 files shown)


ccummingsNV:
Closing this as re-merged with updated SlangPy in PR https://github.com/shader-slang/slangpy/pull/137,

PR #143, opened 2025-04-18
Fixes for windowed diff splatting example

elcuppietan:


experiments/diff-splatting/app.py
```diff
@@ -20,21 +20,17 @@ def __init__(self, title="BRDF Example", width=1024, height=1024, device_type=sg
...
-            image_count=3,
+        self.surface = self._device.create_surface(self._window)
+        self.surface.configure(width=self._window.width, height=self._window.height)
+
+        self._output_texture: 'sgl.Texture' = self.device.create_texture(
             width=self._window.width,
             height=self._window.height,
-            window=self._window,
-            enable_vsync=False,
+            format=sgl.Format.rgba16_float,
```
jkwak-work: Copilot pointed out that the format is changed from float32 to float16.
Is it intended?

elcuppietan: It wasn't intention-- thanks for the catch! Fixed.

jkwak-work:
How about resizing the window?

swoods-nv:
Opened #146

Issue #140, opened 2025-04-18
Slang-RHI update breaks most experiments, some examples


swoods-nv:
The update to Slang-RHI at commit (PR #136 ) breaks most of the samples in the "experiments" directory, and some in the "examples" directory.

Broken examples:
* simplified-splatting
* toy-restir

Broken experiments:
* brdf
* fwd-rasterizer
* neural-network
* soft-rasterizer
* window

swoods-nv:
PR #141 updates simplified splatting to use the new SGL APIs, but the example will still terminate with an error about not being able get get binding data.

gtong-nv:
> With [#141](https://github.com/shader-slang/slangpy/pull/141) , simplified-splatting works on Vulkan devices but not D3D12 devices on AMD integrated. On GTX 2070 super, Vulkan works, but d3d12 appears to stall after the first image with 100% GPU usage.

I tested it again with a 1660ti with both slangpy and SGL at TOT, with D3D12, it doesn't stall but just runs slow. Each iter takes about 20s. Similar perf with VK.

gtong-nv:
Based on the discussion, we are not going to addresss the perf issue in the naive implementation of the 'simplified splatting' sample.

With https://github.com/shader-slang/slang/pull/6446, a new member is added to the struct`s_bwd_prop_simpleSplatBlobs_Intermediates_0 ` and used to do array access directly. The additional member might results in excessive register usage, which is particularly bad for per-pixel thread with 200-loop in each of the thread.
But this should not affect more general / less pathological cases.

Besides, https://github.com/shader-slang/slangpy/pull/149 fixes the remaining tests with the slang-RHI backend change

PR #149, opened 2025-04-25
Fix examples after migrating to slang-rhi

gtong-nv:
Some backend interfaces have changed after migrating to slang-rhi. This change updates the examples and experiments to use the new APIs.
- device.run_garbage_collection() no longer needed
- change create_command_buffer() to create_command_encoder()

Fixes: #140


experiments/neuralnetwork/main.py
```diff
@@ -87,27 +87,24 @@ def training_main():
...
 
-        id = device.submit_command_buffer(cmd)
-        # Stall and wait, then garbage collect for a good interactive experience.
+        id = device.submit_command_buffer(command_encoder.finish())
+        # Stall and wait.
         # Will slow things down a lot though - headless training will run faster.
         # The interactive perf is not representative of the actual training cost
-        device.wait_command_buffer(id)
-        device.run_garbage_collection()
+        device.wait_for_submit(id)
```
gtong-nv, line 107: I'm not sure if I use the command_encoder API correctly. with this change, I still see some warning and not seeing anything presented 
```
Device does not support cooperative vector. Sample will run, but it will be slow
[WARN] (rhi) slang: (0): error 100: failed to load downstream compiler 'spirv-opt'
(0): note 99999: failed to load dynamic library 'slang-glslang'

[WARN] Slang compiler warnings:
eea8ee80b7a68c2e03e0b70675e7cc37d25b4037(16): warning 41018: returning without initializing out parameter 'value'
        }
...

Throughput: 0.00 MSamples/s Epoch time: 0.0ms
[WARN] Slang compiler warnings:
04cede19951a78f1b9f189b873b92c5f5243ab12(16): warning 41018: returning without initializing out parameter 'value'
        }
        ^
```

kaizhangNV: This is not a big issue, `value` in the generated shader may not be used at all.

gtong-nv: yea, the warning should be harmless. My main concern is "Throughput: 0.00 MSamples/s Epoch time: 0.0ms" indicating it's not working as expected.

swoods-nv:
Adding @tunabrain since I think the neural network sample where the comments above appear is originally his code.

Changes look fine to me, but either @tunabrain or @ccummingsNV should give the approval, since I'm less familiar with this codebase.

skallweitNV:
I took these changes here as the basis for https://github.com/shader-slang/slangpy-samples/pull/3

Experiments still need to be ported, but all examples work

PR #177, opened 2025-05-02
Fix metal tests failures

gtong-nv:
Cherry picked https://github.com/shader-slang/sgl/pull/208


src/sgl/device/shader_object.h
```diff
@@ -9,6 +9,8 @@
 #include "sgl/device/fwd.h"
 #include "sgl/device/shader_offset.h"
 #include "sgl/device/resource.h"
+#include "sgl/device/types.h"
```
skallweitNV: these includes seem to be unnecessary

src/sgl/device/shader_object.h
```diff
@@ -49,6 +51,7 @@ class SGL_API ShaderObject : public Object {
     virtual void get_cuda_interop_buffers(std::vector<ref<cuda::InteropBuffer>>& cuda_interop_buffers) const;
 
     rhi::IShaderObject* rhi_shader_object() const { return m_shader_object; }
+    ref<Device> get_device() const { return m_device; }
```
skallweitNV: in sgl we use just the property name for the getters in general .. so `device()` instead of `get_device()`

skallweitNV: also, you might just want to return `Device*` to not incur ref-counting overhead

skallweitNV (commented):
LGTM otherwise

PR #179, opened 2025-05-05
Notarize macos libs and bins for release

gtong-nv:


skallweitNV:
@gtong-nv this has been stale for a while so I'm closing this for now. Feel free to reopen, but it looks like additional notarization is not needed for the release pip packages to run?!

PR #184, opened 2025-05-05
Load PyTorch lazily

### LINKED ISSUE #153: Import `torch` only when needed
Currently, `import slangpy` always leads to `import torch` under the hood (if `torch` is available). Importing torch is quite slow, so we should really just import torch on demand.

skallweitNV:
This implements lazy loading of the `torch` module. Before this change, `import slangpy` on an environment with torch installed takes around 2.5 seconds. After this change, `import slangpy` takes around 0.2 seconds! This is a substantial quality of life improvement when iterating.

fixes #153

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,8 @@ SlangPy uses a `semantic versioning <http://semver.org>`__ policy for its API.
 Version 0.28.0 (TBA)
 ----------------------------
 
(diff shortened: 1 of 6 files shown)


ccummingsNV (approved):
Feel free to get this in - I think it's going to need work but this is still a step forwards from where we were

PR #188, opened 2025-05-06
add gcp runner to slangpy

aasgaonkar:


.github/workflows/ci.yml
```diff
@@ -86,16 +86,25 @@ jobs:
...
 
-      # Setup Python environment.
+      # Setup Python environment (GCP Windows)
       - name: Setup Python environment
+        if: matrix.runs-on.group == 'gcp' && runner.os == 'Windows'
         run: |
-          python -m pip install -r requirements-dev.txt
-          python -m pip install pytest-github-actions-annotate-failures
+          python --version
+          powershell -Command "Start-Process python -ArgumentList '-m pip install -r requirements-dev.txt pytest-github-actions-annotate-failures' -Verb RunAs -Wait"
```
jkwak-work, line 100: I am wondering why this is needed?
Doesn't GCP have git bash?

jkwak-work (commented):
Looks good to me.

.github/workflows/ci.yml
aasgaonkar: no, it doesn't at present. I don't like the setup as well, but need to get tests enabled asap. we would like to use some venv setup via conda or something similar.

Issue #192, opened 2025-05-07
NixOS Dev Environment Support


fangjunzhou:
Currently slangpy use vcpkg as the package manager which [doesn't work on NixOS](https://nix.dev/permalink/stub-ld):

```
-- Running vcpkg install
Could not start dynamically linked executable: /home/fangjun/personal/slangpy/external/vcpkg/vcpkg
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
-- Running vcpkg install - failed
-- Configuring incomplete, errors occurred!
```

nix flake is already supported in https://github.com/shader-slang/slang/pull/6635. I'm currently working on a similar dev shell flake for slangpy.

PR #193, opened 2025-05-07
Proposed Nix flake and uv Support

fangjunzhou:
This is a proposed PR to use nixpkgs as the package manager instead of vcpkg for slangpy. it adds a `flake.nix` with the devShell similar to https://github.com/shader-slang/slang/pull/6635

I've also added a CMake preset specifically for nix to avoid CMake for calling vcpkg binary which doesn't work on NixOS. Details are covered in this issue: https://github.com/shader-slang/slangpy/issues/192

I've also added the support for [uv](https://github.com/astral-sh/uv) as it's easier to work with compared to using raw requirements.txt

**To Use `uv`**

On non-nix machines, `uv sync` will create the `.venv` and source `.venv/bin/activate` will activate the virtual environment. The `uv.lock` will lock the version of all dependencies to guarantee reproducibility.

On nix machines, use `USE_NIX=1 uv sync` so that the setup.py will use nix preset to compile the project.

**Compile with CMake on Nix**

Activate nix devShell environment with `nix develop`. To use the nix cmake preset, configure the project with

```
cmake --preset nix-gcc
```

and compile with

```
cmake --build --preset nix-gcc-release
# cmake --build --preset nix-gcc-debug (for debug build)
```

**Known Issues**

Currently, all the CUDA tests would failed due to slang-rhi not able to find CUDA toolkit:

```
___________________________________________________________________ test_float64[DeviceType.cuda-ShaderModel.sm_6_1-uav] ____________________________________________________________________

device_type = DeviceType.cuda, shader_model = ShaderModel.sm_6_1, view = 'uav'

    @pytest.mark.parametrize("view", ["uav", "srv"])
    @pytest.mark.parametrize("shader_model", helpers.ALL_SHADER_MODELS)
    @pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
    def test_float64(device_type: spy.DeviceType, shader_model: spy.ShaderModel, view: str):
        if sys.platform == "darwin":
            pytest.skip("Not supported on macOS")
    
>       device = helpers.get_device(device_type)

tests/slangpy/device/slang/test_float64.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

type = DeviceType.cuda, use_cache = True

    def get_device(type: spy.DeviceType, use_cache: bool = True) -> spy.Device:
        if use_cache and type in DEVICE_CACHE:
            return DEVICE_CACHE[type]
>       device = spy.Device(
            type=type,
            enable_debug_layers=True,
            compiler_options={
                "include_paths": [SHADER_DIR],
                "debug_info": spy.SlangDebugInfoLevel.standard,
            },
        )
E       RuntimeError: Failed to create device!
E       /home/fangjun/personal/slangpy/src/sgl/device/device.cpp:112 in function Device

tests/slangpy/device/sglhelpers.py:54: RuntimeError
----------------------------------------------------------------------------------- Captured stdout call ------------------------------------------------------------------------------------
[INFO] (rhi) driver: [KNOBS]: All knobs on default.

[INFO] (rhi) driver: [DISKCACHE]: Opened database: "/var/tmp/OptixCache_fangjun/optix7cache.db"
[INFO] (rhi) driver: [DISKCACHE]:     Cache data size: "0 Bytes"
[INFO] (rhi) driver: [DISKCACHE]: Closed database: "/var/tmp/OptixCache_fangjun/optix7cache.db"
[INFO] (rhi) driver: [DISKCACHE]:     Cache data size: "0 Bytes"
----------------------------------------------------------------------------------- Captured stderr call ------------------------------------------------------------------------------------
[ERROR] (rhi) layer: Cannot find CUDA nvrtc. Please ensure that the CUDA Toolkit is installed.
Default search locations:
- CUDA_PATH environment variable
- /usr/local/cuda-x.y
- /usr/lib/x86_64-linux-gnu
```

This is because the cudatoolkit package on nix places `nvrtc` under `$CUDA_PATH/bin` instead of `$CUDA_PATH/bin64` which is the only path searched by slang-rhi. This issue is fixed and merged by this PR: https://github.com/shader-slang/slang-rhi/pull/342

Updating the slang-rhi submodule to the newest version will cause other issues as slang-rhi is adding bindless support: https://github.com/shader-slang/slang-rhi/issues/347. I'll check if this works again once the bindless branch in slangpy is merged.


ccummingsNV:
I don't think we're ready at this point to consider switching package manager without serious investigation into the options. Is there a reason you're looking to change it? I'm not precious over vcpckg, but it's not clear to me what we're solving here.

skallweitNV:
I agree with @ccummingsNV here, if you can keep this in your local branch that would be welcome.
Out of curiosity, what does FHS stand for?

fangjunzhou:
HFS stands for Hierarchical File System. Basically, NixOS doesn't have /usr/bin and /usr/lib globally. Instead, it creates dynamic links for all the binaries to /nix/store. This makes it fully declarative and reproducible: https://nixos.org/guides/how-nix-works/

However, this also makes it impossible to execute any ordinary Linux binary directly so I can't use vcpkg or Conan on NixOS directly.

For nix users, we have to figure out all the dependencies that a project really needs and configure it declaratively for in a nix derivation. This is pretty much what I did in flake.nix. But this also means it's impossible to miss any hidden dependencies a project requires. For example, imagine you installed some dependency with apt install on your Linux machine but you forget about it. Even if you use vcpkg, the project won't compile or might behave differently on other machines if they don't have the same package installed. In this case, CUDA version, Vulkan version, and some X11 libraries could infect the system, configuring it with nix lock those dependencies and make the entire project more reproducible.

It's understandable that this is probably unplanned and I'll keep it in my own fork.

Issue #199, opened 2025-05-09
Add More Option to `spy.create_device`


fangjunzhou:
I'm trying to use `sgl.device.print` as shown in the example `examples/print/print.py`. However, this requires creating the device with `spy.Device` manually currently. This makes it a little bit complicated to include the slangpy module which is done automatically in `spy.create_device`.

I think we should add more optional parameters like `enable_print` and `enable_hot_reload` to `spy.create_device` to expose these options to slangpy users.

PR #200, opened 2025-05-09
Added `enable_print` and `enable_hot_reload` options to `spy.create_device`

fangjunzhou:
Proposed improvement for #199

### DIFF
--- a/slangpy/core/utils.py
+++ b/slangpy/core/utils.py
@@ -23,7 +23,7 @@ def create_device(
     include_paths: Sequence[Union[str, PathLike[str]]] = [],
     enable_cuda_interop: bool = False,
     enable_print: bool = False,
-    enable_hot_reload: bool = False,
+    enable_hot_reload: bool = True,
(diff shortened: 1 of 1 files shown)


slangpy/core/utils.py
```diff
@@ -22,6 +22,8 @@ def create_device(
     adapter_luid: Optional[Sequence[int]] = None,
     include_paths: Sequence[Union[str, PathLike[str]]] = [],
     enable_cuda_interop: bool = False,
+    enable_print: bool = False,
+    enable_hot_reload: bool = False,
```
ccummingsNV: enable_hot_reload should default to True, to match the constructor of Device itself.

ccummingsNV (changes requested):

slangpy/core/utils.py
skallweitNV:
```suggestion
    enable_hot_reload: bool = True,
```

PR #223, opened 2025-05-19
Fix buffer cursor for metal

skallweitNV:
Check for metal device type for special handling of `boolX` values.

### DIFF
--- a/src/sgl/device/buffer_cursor.cpp
+++ b/src/sgl/device/buffer_cursor.cpp
@@ -2,6 +2,7 @@
 
 #include "buffer_cursor.h"
 
+#include "sgl/device/device.h"
 #include "sgl/device/shader_object.h"
 #include "sgl/device/resource.h"
 #include "sgl/device/cuda_interop.h"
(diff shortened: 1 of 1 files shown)


src/sgl/device/buffer_cursor.cpp
fangjunzhou, line 1: If we can get the device type with `m_buffer->resource()->device()->type()`, do we still need to guard the metal specific code with `#SGL_MACOS`?

src/sgl/device/buffer_cursor.cpp
```diff
@@ -323,20 +324,24 @@ template<>
 SGL_API void BufferElementCursor::set(const bool1& value)
 {
 #if SGL_MACOS
-    bool1 v = value;
-#else
-    uint1 v(value.x ? 1 : 0);
+    if (m_buffer->resource()->device()->type() == DeviceType::metal) {
```
tdavidovicNV, line 327: This needs to be done for CUDA as well (it also has 1B bool).
The if statement should instead read:
```
if (cursor.slang_type_layout()->getElementTypeLayout()->getSize() == 1) {
```
and not be tied to Mac only.

(I have a MR that should get most of the bools working, CUDA and Mac that's almost ready to go)

PR #229, opened 2025-05-21
Fix Inconsistant `ref<NativeCallData>` Usage in `NativeFunctionNode::call`

fangjunzhou:
I've noticed that all the NativeCallData generated from FunctionNode are using `ref` wrapper while the one fetched from cache under `NativeFunctionNode::call` is using raw pointer. This is probably a typo and should be fixed.


ccummingsNV (approved):
Strictly speaking this is fine, as the call data is cached at this point so the ref counting is just added cost. But it's negligible, and the PR is safer, so I'm happy to merge it.

PR #238, opened 2025-05-23
Fixed the shader cursor matrix alignment issue for 2x2 matrices

### LINKED ISSUE #227: `InstanceBuffer.construct` Failed with Matrix Type
**Problem Description**

Having the following `example.slang`:

```slang
struct Example {
  float2x2 mat;

  __init(float2x2 mat = float2x2(1, 0, 0, 1)) {
    this.mat = mat;
  }
};
```

Load it as a module and then call the constructor in a `InsatnceBuffer`

```python
example_buf = spy.InstanceBuffer(
    struct=module.Example.as_struct(),
    shape=(16,)
)
example_buf.construct(
    mat = spy.float2x2()
)
```

Will cause the following error:

```
RuntimeError: : Unsupported element type
[/Users/fangjun/Documents/stanford/bvhgs/external/slangpy/src/slangpy_ext/./device/cursor_utils.h:287](http://localhost:8888/external/slangpy/src/slangpy_ext/device/cursor_utils.h#line=286) in function operator()
[/Users/fangjun/Documents/stanford/bvhgs/external/slangpy/src/slangpy_ext/./device/cursor_utils.h:361](http://localhost:8888/external/slangpy/src/slangpy_ext/device/cursor_utils.h#line=360) in function write
```

**Environment**

slangpy custom fork from 0.23.0
macOS 15.3.2

**Problem Cause Analysis**

I've done some basic debug to troubleshoot this issue. After inserting the following `SGL_PRINT`:
```
SGL_PRINT((int)type->getScalarType());
SGL_PRINT(type->getRowCount());
SGL_PRINT(type->getColumnCount());
```
under this line: https://github.com/shader-slang/slangpy/blob/256576b6acdc304949986440774e0c83d902d21e/src/slangpy_ext/device/cursor_utils.h#L403

I got:
```
(int)type->getScalarType() = 7
type->getRowCount() = 2
type->getColumnCount() = 2
```

Scalar type 7 is `SLANG_SCALAR_TYPE_FLOAT16`, which explains the failure here as only `float32` matrix cursor are supported: https://github.com/shader-slang/slangpy/blob/256576b6acdc304949986440774e0c83d902d21e/src/slangpy_ext/device/cursor_utils.h#L337-L341

Note that manually specifying the dtype of `float2x2` with `np.float32` doesn't work either:

```python
example_buf.construct(
    mat = spy.float2x2(np.array([[1, 0], [0, 1]], dtype=np.float32))
) 
```

The root cause of this issue is still unclear.

fangjunzhou:
Fix #227 as discussed in the discord channel.

### DIFF
--- a/src/sgl/device/shader_cursor.cpp
+++ b/src/sgl/device/shader_cursor.cpp
@@ -568,7 +568,7 @@ void ShaderCursor::_set_matrix(
         ShaderOffset offset = m_offset;
         for (int row = 0; row < rows; ++row) {
             m_shader_object->set_data(offset, reinterpret_cast<const uint8_t*>(data) + row * row_size, row_size);
-            offset.uniform_offset += row_size;
+            offset.uniform_offset += static_cast<uint32_t>(row_size);
         }
     } else {
         m_shader_object->set_data(m_offset, data, size);
--- a/src/sgl/device/shader_cursor.cpp
+++ b/src/sgl/device/shader_cursor.cpp
@@ -568,7 +568,7 @@ void ShaderCursor::_set_matrix(
         ShaderOffset offset = m_offset;
         for (int row = 0; row < rows; ++row) {
             m_shader_object->set_data(offset, reinterpret_cast<const uint8_t*>(data) + row * row_size, row_size);
-            offset.uniform_offset += 16;
+            offset.uniform_offset += row_size;
         }
     } else {
         m_shader_object->set_data(m_offset, data, size);


csyonghe (approved):

src/sgl/device/shader_cursor.cpp
```diff
@@ -568,7 +568,7 @@ void ShaderCursor::_set_matrix(
         ShaderOffset offset = m_offset;
         for (int row = 0; row < rows; ++row) {
             m_shader_object->set_data(offset, reinterpret_cast<const uint8_t*>(data) + row * row_size, row_size);
-            offset.uniform_offset += 16;
+            offset.uniform_offset += row_size;
```
kaizhangNV: Note, this is not completely right.
for col == 2 and 4, it's correct.

But col == 3, it's not necessarily right.

Because metal will align it to 4.

So the correct implementation should be using reflection to find out the row alignment.
and do
```
auto rowAglinment = m_type_layout->getAlignment();
offset.uniform_offset += rowAglinment;
```

kaizhangNV: I revisit this code, and I noticed that the _set_matrix is called after matrix alignment is correctly handled, so this code is actually correct.

kaizhangNV:
Please fix the build error.

fangjunzhou:
It seems that the second row is still not written correctly on Windows. I don't have a windows machine by my hand now. I'll take a look at this issue latter.

kaizhangNV:
Close this PR now, because more thorough fix https://github.com/shader-slang/slangpy/pull/242 is already merged.

PR #239, opened 2025-05-25
Fixed reference to Bitmap in jupyter.py

p4vv37:
Fix for
NameError: name 'slangpy' is not defined
at texture preview in Jupyter Notebook


skallweitNV:
Thanks for the PR. Can you change `slangpy.Bitmap` to just `Bitmap` instead of adding `import slangpy`? This would make things more consistent.

p4vv37:
That's exactly what I did int this PR :) If I understood your comment correctly.

PR #242, opened 2025-05-27
fix matrix alignment issue during create_calldata

kaizhangNV:
When creating the calldata, we didn't calculate the size of the matrix correctly because there won't be alignment issue in some platforms, so we will need to add paddings to those cases. Similarly, we will need to removing the paddings when reading back from underlying buffer.

And this has to be done in create_calldata and read_calldata in valueref.py becuase those call will finally writting or reading the raw device buffer where has no layout info. Therefore, we have to organize the data layout correctly before writing and after reading.


src/sgl/device/buffer_cursor.cpp
```diff
@@ -194,7 +194,16 @@ void BufferElementCursor::_set_matrix(
 )
 {
     cursor_utils::check_matrix(m_type_layout->slang_target(), size, scalar_type, rows, cols);
-    write_data(m_offset, data, size);
+    size_t stride = slang_type_layout()->getStride();
+    if (stride != size) {
+        size_t rowStride = stride / rows;
```
csyonghe: I don't know what naming convention is used in this codebase, but it seems that local variables should be named `row_stride` here.

kaizhangNV: OK, will change the naming.

ccummingsNV (approved):

PR #248, opened 2025-05-30
Improve Tensor/Buffer apis + andd load from image

ccummingsNV:
- Brought the Tensor/NDBuffer apis for empty and zeros in line with each other
- Added a 'load_from_image' function to both


slangpy/types/buffer.py
```diff
@@ -100,6 +104,62 @@ def resolve_element_type(program_layout: SlangProgramLayout, element_type: Any)
...
+    path: Union[str, PathLike[str]],
+    flip_y: bool = False,
+    linearize: bool = False,
+    scale: float = 1.0,
+    offset: float = 0.0,
+    greyscale: bool = False,
+) -> np.ndarray[Any, Any]:
+    """
+    Helper to load an image from a file and convert it to a floating point tensor.
+    """ ""
```
skallweitNV, line 117: extra `""`

slangpy/types/buffer.py
```diff
@@ -100,6 +104,62 @@ def resolve_element_type(program_layout: SlangProgramLayout, element_type: Any)
...
+
+    # Load bitmap + convert to numpy array
+    bitmap = Bitmap(path)
+    data: np.ndarray[Any, Any] = np.asarray(bitmap)
```
skallweitNV: you are doing a copy here, you can add `copy=False` to avoid that

slangpy/types/buffer.py
```diff
@@ -100,6 +104,62 @@ def resolve_element_type(program_layout: SlangProgramLayout, element_type: Any)
...
+
+    # Validate array shape.
+    if data.ndim != 3:
```
skallweitNV: is this true for 2D/grayscale bitmaps too? don't remember if they are returned with dims=2

slangpy/types/buffer.py
```diff
@@ -100,6 +104,62 @@ def resolve_element_type(program_layout: SlangProgramLayout, element_type: Any)
...
+    ]:
+        data = data.astype(np.float32) / 255.0
+
+    # Optionally apply gammar correction.
+    if bitmap.srgb_gamma:
+        if linearize:
+            data = np.pow(data, 2.2)
+
+    # Optionally convert to single channel by averaging channels.
+    if greyscale:
```
skallweitNV: a lot of these conversion you can do with the `Bitmap` class
to go to grayscale you typically don't want just average but dot product with (0.2126, 0.7152, 0.0722) to get luminance

Issue #255, opened 2025-06-03
Re-enable tests and asserts in array.py


tdavidovicNV:
Had to remove the assert and corresponding test as it was not behaving correctly.

The failure case was binding `textures` in this type of struct:
```
struct Scene
{
     public Tensor<float, 3> textures[16];
}
```

See PR https://github.com/shader-slang/slangpy/pull/256 .

ccummingsNV:
Now redundant due to change in array behaviour

PR #256, opened 2025-06-03
Disabled assert and its test

tdavidovicNV:
Disabled the assert as it is interfering with existing binding.


tdavidovicNV:
This is to be reverted/fixed via https://github.com/shader-slang/slangpy/issues/255

PR #263, opened 2025-06-05
Bake objects

ccummingsNV:


src/slangpy_ext/device/cursor_utils.h
```diff
@@ -376,9 +379,21 @@ class WriteConverterTable {
...
-        // Check if nbval has a '.uniforms' function that can be called to get simple python object
-        if (nb::hasattr(nbval, "uniforms")) {
-            nbval = nb::getattr(nbval, "uniforms")();
+        // Special case for handling SlangPy PackedArg.
+        if (nb::isinstance<sgl::slangpy::NativePackedArg>(nbval)) {
+            auto packed_arg = nb::cast<sgl::slangpy::NativePackedArg*>(nbval);
+            if constexpr (requires { self.set_object(nullptr); }) {
+                self.set_object(packed_arg->shader_object());
+            } else {
+                SGL_THROW("Cannot set packed arg to shader object without set_object method.");
```
skallweitNV: not sure users will understand this .. this will happen for `BufferCursor` mostly, right?

src/slangpy_ext/utils/slangpy.h
```diff
@@ -444,6 +459,12 @@ class NativeBoundVariableRuntime : public Object {
     /// Set the shape being used for the current call.
     void set_shape(const Shape& shape) { m_shape = shape; }
 
+    /// Check if this is a parameter block.
+    bool get_is_param_block() const { return m_is_param_block; }
```
skallweitNV: We have the convention to name getters just the name, i.e. `is_param_block` and setters with `set_` prefix. It's a bit of an edge case here because this is binding wrappers only, so less important than sgl core API.

PR #265, opened 2025-06-06
enable more cuda tests

skallweitNV:
Enable CUDA tests now working after updating to latest slang with lots of CUDA fixes.


skallweitNV:
Looks like there is still an issue locating the `cuda_fp16.h` header file on linux systems.

Issue #72, opened 2025-02-12
Thread group control through implementation of Call group system


bmillsNV:
Proposal: Traditionally, compute shader dispatch follows a two-tier 3D grid model (ignoring clusters for now). We’re all familiar with this structure, as illustrated in this diagram: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/images/threadgroupids.png
I propose that, just as the traditional compute pipeline defines a thread-group, we introduce a corresponding concept: a call-group.
A call-group is an ND-shaped structure where all call-IDs within a call-group are guaranteed to execute within the same hardware thread-group.
This could be implemented very simply by just:
- Setting the thread group size to [1,1,full-call-group-size]
- Calculate call group id from dispatch group id
- Calculate call group thread id from dispatch group thread id
- Calculate call id by combining the 2
This approach integrates thread-groups pretty seamlessly with our new ND execution model. In traditional compute you can split your dispatch into 1D, 2D or 3D cells. Now for a 5D call, we can now say that group-shared memory exists in 5D cells.
We could also optimize where appropriate (or when hinted to by the user), and map call groups directly to thread groups where possible, which would simply reduce the need for arithmetic operations. i.e. if I specified a 3D call group size of [3,4,5], we can map that directly to a thread group size of [3,4,5] without needing to convert the linear representation to a grid one explicitly. 

<img width="667" alt="Image" src="https://github.com/user-attachments/assets/a5905c20-9f1d-4de5-a3d3-516dd6da7683" />

sricker-nvidia:
Implemented in https://github.com/shader-slang/slangpy/pull/267, closing

PR #267, opened 2025-06-06
Enhanced Kernel Dispatch Flexibility in SlangPy

sricker-nvidia:
Enhanced Kernel Dispatch Flexibility in SlangPy (#72)

This commit introduces enhanced kernel dispatch flexibility via
a new "call_group_shape()" function that enables more fine-grained
control over GPU thread organization and dispatch patterns. A simple
usage example might look something like:

`call_data = module.test2D.call_group_shape(Shape((4,8)))(spy.grid((32,64)), spy.thread_id(), _result='numpy')`

Where a shader named "test2D" is being given a call group shape of
4x8 (note that the ordering is [y,x] here), a call shape of 32x64
as a result of the spy.grid((32,64)) being passed in, as well as given
the thread_id. The mapping of call-id's to thread-id's from such a call
would look like:

[call-id[0] (y), call-id[1] (x), thread-id]

[   0    0    0]
[   1    0    1]
[   2    0    2]
...
[  61    0  229]
[  62    0  230]
[  63    0  231]
[   0    1    8]
[   1    1    9]
[   2    1   10]
...
[  61    1  237]
[  62    1  238]
[  63    1  239]

Running without the call_group_shape function:

`call_data = module.test2D.(spy.grid((32,64)), spy.thread_id(), _result='numpy')`

or alternatively:

`call_data = module.test2D.call_group_shape(Shape(()))(spy.grid((32,64)), spy.thread_id(), _result='numpy')`

results in SlangPy using it's standard linear dispatch scheme using call groups
containing 32 threads and the mapping from call-id's to thread-id's from such
calls would look like:

[   0    0    0]
[   1    0    1]
[   2    0    2]
...
[  61    0   61]
[  62    0   62]
[  63    0   63]
[   0    1   64]
[   1    1   65]
[   2    1   66]
...
[  61    1  125]
[  62    1  126]
[  63    1  127]

Conceptually, the default when not specifying a call group shape can be thought
of as running dispatches where each thread has its own call group, but in
actuallity, threads are still run in groups of 32. Thus a call like:

`call_data = module.test2D_thread.call_group_shape(Shape((1,1)))(spy.grid((32,64)), spy.thread_id(), _result='numpy')`

Will ultimately function differently than the two examples above and would likely
have worse performance. Similarly:

`call_data = module.test2D_thread.call_group_shape(Shape((1,32)))(spy.grid((32,64)), spy.thread_id(), _result='numpy')`

Can also function differently from the default case where call group shape is
not specified, as we can now run into alignment issues when the call shape is
not aligned to 32 threads in the "x" dimension (ex a 1x33 call group shape on a
32x64 call shape).

The new functionality also works with arbitrary dimensionalities:

`...test5D_thread.call_group_shape(Shape((2,2,2,2,2)))(spy.grid((2,2,2,32,64)), spy.thread_id(), _result='numpy')`

Using the new call_group_shape functionality, developers can now specify how
threads are arranged spatially, allowing for better uses of shared memory and
better control over memory access patterns.

In order to support the new functionality, the change extends the existing
dispatch system to support call_group_shape parameters with proper validation,
padding, and alignment handling, while also handling edge cases for misaligned
shapes and dimensions smaller than group sizes, while also providing error reporting
for invalid configurations.

The change includes unit tests that validate call group functionality across
1D through 5D dispatches to ensure mathematical consistency of the group ID
calculations and thread ID assignments. The tests cover both typical use
cases and edge cases and have been added under:

-slangpy/tests/slangpy_tests/test_call_groups.py

The change also includes integration tests to ensure call group functionality
works with SlangPy's existing major features such as tensor operations with
autodiff support, NDBuffer processing, texture sampling and filtering, and
transformation mapping operations. These integration tests have been added
under:

-slangpy/tests/slangpy_tests/test_call_group_integrations.py

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -88,6 +88,8 @@ class CallData(NativeCallData):
             positional_mapping = build_info.map_args
             keyword_mapping = build_info.map_kwargs
             type_conformances = build_info.type_conformances
+            if build_info.call_group_shape is not None:
+                self.call_group_shape = build_info.call_group_shape
 
             # Store logger from either function or module
             if build_info.logger is not None:
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -88,8 +88,7 @@ class CallData(NativeCallData):
             positional_mapping = build_info.map_args
             keyword_mapping = build_info.map_kwargs
             type_conformances = build_info.type_conformances
-            if build_info.call_group_shape is not None:
-                self.call_group_shape = build_info.call_group_shape
+            self.call_group_shape = build_info.call_group_shape
 
             # Store logger from either function or module
             if build_info.logger is not None:
--- a/slangpy/core/callsignature.py
+++ b/slangpy/core/callsignature.py
@@ -396,16 +396,100 @@ def generate_code(
 
     # Generate the header
     cg.add_import("slangpy")
+
+    call_data_len = context.call_dimensionality
+
+    # Get the call group size so we can see about using it when generating the
+    # [numthreads(...)] attribute. We use 1 as the default size if a call
+    # group shape has not been set, as we can use that to make things "linear".
+    # Note that when size is 1, we will still launch a group of 32 threads,
+    # but each thread is conceptually in its own group of size 1. This then
+    # leads to a linearly calculated call_id based on threadID.x and the call
+    # shape's size and strides.
+    call_group_size = 1
+    call_group_shape = build_info.call_group_shape
+    if call_group_shape is not None:
+        call_group_shape_vector = call_group_shape.as_list()
+
+        # Validate call_group_shape dimensions and values before using them
+        if len(call_group_shape_vector) > context.call_dimensionality:
+            raise KernelGenException(
+                f"call_group_shape dimension ({len(call_group_shape_vector)}) must be <= "
+                f"call_shape dimension ({context.call_dimensionality}). "
+                f"call_group_shape cannot have more dimensions than call_shape."
+            )
+
+        # Validate that all call_group_shape values are >= 1
+        for i, dim in enumerate(call_group_shape_vector):
+            if dim < 1:
+                raise KernelGenException(
+                    f"call_group_shape[{i}] = {dim} is invalid. "
+                    f"All call_group_shape elements must be >= 1."
+                )
+
+        # Calculate call group size as product of all dimensions
+        for dim in call_group_shape_vector:
+            call_group_size *= dim
+        # Check if call_group_size exceeds hardware limits
+        if call_group_size > 1024:
+            raise KernelGenException(
+                f"call_group_size ({call_group_size}) exceeds the typical 1024 maximum "
+                f"enforced by most APIs. Consider reducing your call_group_shape dimensions."
+            )
+
+    # Specify some global-scope static variables here to allow users to make calls
+    # to functions like get_call_id<N>() etc in order to access call_id,
+    # call_group_id, and call_group_thread_id.
+
+    # Note: Support for static global variables should be seen as a legacy feature
+    #       and its use is discouraged. As such, this may need to be reworked in
+    #       the future.
+    if call_data_len > 0:
+        # Context wants call_id as an int for some reason, so respect that here
+        # for now
+        cg.add_snippet("call_id", f"static int[{call_data_len}] call_id;\n")
+        cg.add_snippet("call_group_id", f"static uint[{call_data_len}] call_group_id;\n")
+        cg.add_snippet(
+            "call_group_thread_id", f"static uint[{call_data_len}] call_group_thread_id;\n"
+        )
+
+        # TODO: Ideally there would be a way to just return call_id here so we could avoid wasted effort itterating.
+        #       Not enough of a slang expert at the moment to know what that could/should look like.
+        cg.add_snippet(
+            "get_call_id",
+            f"export public static int[N] get_call_id<let N: int>() {{ int[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_id[i]; }} return ret; }};\n",
+        )
+        cg.add_snippet(
+            "get_call_group_id",
+            f"export public static uint[N] get_call_group_id<let N: int>() {{ uint[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_id[i]; }} return ret; }};\n",
+        )
+        cg.add_snippet(
+            "get_call_group_thread_id",
+            f"export public static uint[N] get_call_group_thread_id<let N: int>() {{ uint[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_thread_id[i]; }} return ret; }};\n",
+        )
+
     cg.add_import(build_info.module.name)
 
     # Generate constants if specified
     generate_constants(build_info, cg)
 
     # Generate call data inputs if vector call
-    call_data_len = context.call_dimensionality
     if call_data_len > 0:
-        cg.call_data.append_statement(f"int[{call_data_len}] _call_stride")
+        # We use the call shape dimensions to detect cases when the call shape
+        # and the call group shape are not aligned. When a thread's call id
+        # falls outside the call shape, we need it to return early. This is
+        # similar to the default linear case when the call shape size is not
+        # 32 thread aligned.
         cg.call_data.append_statement(f"int[{call_data_len}] _call_dim")
+        # A group can be thought of as a "window" looking at a
+        # portion of the entire call shape. Grid here refers to the
+        # N dimensional call shape being broken up into some number of N
+        # dimensional "window"s / groups.
+        cg.call_data.append_statement(f"uint[{call_data_len}] _grid_stride")
+        cg.call_data.append_statement(f"uint[{call_data_len}] _grid_dim")
+        cg.call_data.append_statement(f"uint[{call_data_len}] _group_stride")
+        cg.call_data.append_statement(f"uint[{call_data_len}] _group_dim")
+
     cg.call_data.append_statement(f"uint3 _thread_count")
 
     # Generate the context structure
@@ -483,23 +567,69 @@ def generate_code(
 
     # Generate the main function
     cg.kernel.append_line('[shader("compute")]')
-    cg.kernel.append_line("[numthreads(32, 1, 1)]")
-    cg.kernel.append_line("void compute_main(uint3 dispatchThreadID: SV_DispatchThreadID)")
+    if call_group_size != 1:
+        cg.kernel.append_line(f"[numthreads({call_group_size}, 1, 1)]")
+    else:
+        cg.kernel.append_line("[numthreads(32, 1, 1)]")
+
+    # Note: While flat_call_thread_id is 3-dimensional, we consider it "flat" and 1-dimensional because of the
+    #       true call group shape of [x, 1, 1] and only use the first dimension for the call thread id.
+    cg.kernel.append_line("void compute_main(uint3 flat_call_thread_id: SV_DispatchThreadID)")
     cg.kernel.begin_block()
-    cg.kernel.append_statement("if (any(dispatchThreadID >= call_data._thread_count)) return")
+    cg.kernel.append_statement("if (any(flat_call_thread_id >= call_data._thread_count)) return")
+
+    # Calculate the ids that we'll use to compute the call id. We'll also store the
+    # N-dimensional call_*_id's as global-scope static variables such that they can
+    # be "queried" in user shaders via getters provided by the slangpy module.
+    if call_data_len > 0:
+        # Calculate flattened id's that will be used to determine the N-dimensional
+        # id's.
+        cg.kernel.append_statement(f"uint flat_call_id = flat_call_thread_id.x")
+        # We use call_group_size here, which can be 1, because we want to allow for
+        # a default linear dispatch when the call group shape is not specified. When
+        # call_group_size is 1, each thread conceptually has its own group (even though
+        # they will stil be dispatched in groups of 32), where the call_group_id will match
+        # the call_id. This results in linear mapping from the flat thread id to the
+        # call id where call_id[i] = flat_call_id/call_stride[i] % call_dim[i].
+        cg.kernel.append_statement(
+            f"uint flat_call_group_id = flat_call_thread_id.x / {call_group_size}"
+        )
+        # Again, we want to use flat_call_thread_id.x for calcualting flat_call_group_thread_id
+        # rather than something like SV_GroupIndex because we want to be able to force
+        # flat_call_group_thread_id == flat_call_thread_id.x in the default case using
+        # linear dispatches. The same could be accomplished with SV_GroupIndex and
+        # [numthreads(1, 1, 1)], but this would likely be less performant.
+        cg.kernel.append_statement(
+            f"uint flat_call_group_thread_id = flat_call_thread_id.x % {call_group_size}"
+        )
 
     # Loads / initializes call id
-    context_args = "dispatchThreadID"
+    context_args = "flat_call_thread_id"
     if call_data_len > 0:
-        cg.kernel.append_line(f"int[{call_data_len}] call_id = {{")
+        cg.kernel.append_line(f"for (int i=0; i<{call_data_len}; i++)")
+        cg.kernel.append_line("{")
         cg.kernel.inc_indent()
-        for i in range(call_data_len):
-            cg.kernel.append_line(
-                f"(dispatchThreadID.x/call_data._call_stride[{i}]) % call_data._call_dim[{i}],"
-            )
+
+        # Calculate the N-dimensional id's
+        cg.kernel.append_statement(
+            "call_group_thread_id[i] = (flat_call_group_thread_id/call_data._group_stride[i]) % call_data._group_dim[i]"
+        )
+        cg.kernel.append_statement(
+            "call_group_id[i] = (flat_call_group_id/call_data._grid_stride[i]) % call_data._grid_dim[i]"
+        )
+        cg.kernel.append_statement(
+            f"call_id[i] = call_group_id[i] * call_data._group_dim[i] + call_group_thread_id[i]"
+        )
+        # In the event that the call shape is not aligned to the call group shape,
+        # we use an aligned call shape to calculate the number of threads that we
+        # need. However, that means that some threads will fall outside the call shape
+        # and as a result will need to return early and be wasted.
+        cg.kernel.append_statement("if (call_id[i] >= call_data._call_dim[i]) return")
         cg.kernel.dec_indent()
         cg.kernel.append_statement("}")
-        context_args += ", call_id"
+
+        context_args += f", call_id"
+
     cg.kernel.append_statement(f"Context context = {{{context_args}}}")
 
     # Call the trampoline function
--- a/slangpy/core/callsignature.py
+++ b/slangpy/core/callsignature.py
@@ -411,11 +411,11 @@ def generate_code(
     if call_group_shape is not None:
         call_group_shape_vector = call_group_shape.as_list()
 
-        # Validate call_group_shape dimensions and values before using them
+        # Validate call_group_shape dimensionality and values before using them
         if len(call_group_shape_vector) > context.call_dimensionality:
             raise KernelGenException(
-                f"call_group_shape dimension ({len(call_group_shape_vector)}) must be <= "
-                f"call_shape dimension ({context.call_dimensionality}). "
+                f"call_group_shape dimensionality ({len(call_group_shape_vector)}) must be <= "
+                f"call_shape dimensionality ({context.call_dimensionality}). "
                 f"call_group_shape cannot have more dimensions than call_shape."
             )
         elif len(call_group_shape_vector) < context.call_dimensionality:
@@ -453,39 +453,33 @@ def generate_code(
                 f"enforced by most APIs. Consider reducing your call_group_shape dimensions."
             )
 
-    # Specify some global-scope static variables here to allow users to make calls
-    # to functions like get_call_id<N>() etc in order to access call_id,
-    # call_group_id, and call_group_thread_id.
-
-    # Note: Support for static global variables should be seen as a legacy feature
-    #       and its use is discouraged. As such, this may need to be reworked in
-    #       the future.
-    if call_data_len > 0:
-        cg.add_snippet("call_id", f"static int[{call_data_len}] call_id;\n")
-        cg.add_snippet("call_group_id", f"static int[{call_data_len}] call_group_id;\n")
-        cg.add_snippet(
-            "call_group_thread_id", f"static int[{call_data_len}] call_group_thread_id;\n"
-        )
-
-        # TODO: Ideally there would be a way to just return call_id here so we could avoid wasted effort itterating.
-        #       Not enough of a slang expert at the moment to know what that could/should look like.
-        cg.add_snippet(
-            "get_call_id",
-            f"export public static int[N] get_call_id<let N: int>() {{ int[N] ret; [unroll] for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_id[i]; }} return ret; }};\n",
-        )
-        cg.add_snippet(
-            "get_call_group_id",
-            f"export public static int[N] get_call_group_id<let N: int>() {{ int[N] ret; [unroll] for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_id[i]; }} return ret; }};\n",
-        )
-        cg.add_snippet(
-            "get_call_group_thread_id",
-            f"export public static int[N] get_call_group_thread_id<let N: int>() {{ int[N] ret; [unroll] for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_thread_id[i]; }} return ret; }};\n",
-        )
-
     cg.add_import(build_info.module.name)
 
     # Generate constants if specified
     generate_constants(build_info, cg)
+
+    # Generate additional link time constants definition code. These are declared in callshape.slang
+    # and used to generated call_ids that can be queried by user modules.
+    cg.constants.append_statement(f"export static const int call_data_len = {call_data_len}")
+    cg.constants.append_statement(f"export static const int call_group_size = {call_group_size}")
+
+    # Also generate the call group shape and stride arrays as link time constants. Using constants
+    # should yield better performance than passing these in as uniforms.
+    cg.constants.append_line(f"export static const int[call_data_len] call_group_strides = {{")
+    cg.constants.inc_indent()
+    if call_group_size != 1:
+        for i in range(call_data_len):
+            cg.constants.append_line(f"{call_group_strides[i]},")
+    cg.constants.dec_indent()
+    cg.constants.append_statement("}")
+
+    cg.constants.append_line(f"export static const int[call_data_len] call_group_shape_vector = {{")
+    cg.constants.inc_indent()
+    if call_group_size != 1:
+        for i in range(call_data_len):
+            cg.constants.append_line(f"{call_group_shape_vector[i]},")
+    cg.constants.dec_indent()
+    cg.constants.append_statement("}")
 
     # Generate call data inputs if vector call
     if call_data_len > 0:
@@ -503,9 +497,6 @@ def generate_code(
         cg.call_data.append_statement(f"int[{call_data_len}] _call_dim")
 
     cg.call_data.append_statement(f"uint3 _thread_count")
-
-    # Generate the context structure
-    cg.context.type_alias("Context", f"ContextND<{context.call_dimensionality}>")
 
     # Generate call data definitions for all inputs to the kernel
     for node in signature.values():
@@ -594,70 +585,19 @@ def generate_code(
 
     # Loads / initializes call id
     context_args = "flat_call_thread_id"
+
+    # Call init_thread_local_call_shape_info to initialize the call shape info. See
+    # definition in callshape.slang.
     if call_data_len > 0:
+        cg.kernel.append_line(
+            f"""
+        if (!init_thread_local_call_shape_info(flat_call_group_thread_id,
+            flat_call_group_id, flat_call_thread_id, call_data._grid_stride,
+            call_data._grid_dim, call_data._call_dim))
+            return;"""
+        )
 
-        if call_group_size != 1:
-            # Calculate the N-dimensional id's based on call group shape.
-            cg.kernel.append_line(f"call_group_thread_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(
-                    f"(flat_call_group_thread_id/{call_group_strides[i]}) % {call_group_shape_vector[i]},"
-                )
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-
-            cg.kernel.append_line(f"call_group_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(
-                    f"(flat_call_group_id.x/call_data._grid_stride[{i}]) % call_data._grid_dim[{i}],"
-                )
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-
-            cg.kernel.append_line(f"call_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(
-                    f"call_group_id[{i}] * {call_group_shape_vector[i]} + call_group_thread_id[{i}],"
-                )
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-            # In the event that the call shape is not aligned to the call group shape,
-            # we use an aligned call shape to calculate the number of threads that we
-            # need. However, that means that some threads will fall outside the call shape
-            # and as a result will need to return early and be wasted.
-            for i in range(call_data_len):
-                cg.kernel.append_statement(f"if (call_id[{i}] >= call_data._call_dim[{i}]) return")
-        else:
-            # Default linear calculate of the N-dimensional id's.
-            # We can use _grid_dim and _grid_stride here as each thread technically has its own
-            # call group in this case, such that _grid_dim == _call_dim and _grid_stride == _call_stride
-            cg.kernel.append_line(f"call_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(
-                    f"(flat_call_thread_id.x/call_data._grid_stride[{i}]) % call_data._grid_dim[{i}],"
-                )
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-
-            cg.kernel.append_line(f"call_group_thread_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(f"0,")
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-
-            cg.kernel.append_line(f"call_group_id = {{")
-            cg.kernel.inc_indent()
-            for i in range(call_data_len):
-                cg.kernel.append_line(f"call_id[{i}],")
-            cg.kernel.dec_indent()
-            cg.kernel.append_statement("}")
-
-        context_args += f", call_id"
+        context_args += ", CallShapeInfo::get_call_id().shape"
 
     cg.kernel.append_statement(f"Context context = {{{context_args}}}")
 
--- a/slangpy/core/callsignature.py
+++ b/slangpy/core/callsignature.py
@@ -418,6 +418,16 @@ def generate_code(
                 f"call_shape dimension ({context.call_dimensionality}). "
                 f"call_group_shape cannot have more dimensions than call_shape."
             )
+        elif len(call_group_shape_vector) < context.call_dimensionality:
+            # Call group shape size is less than the call shape size so we need to
+            # pad the call group shape with 1's to account for the missing dimensions.
+            # However, inserting at the front of the list will be inefficient, so
+            # log a debug message, giving users a chance to correct their calls.
+
+            missing_dims = context.call_dimensionality - len(call_group_shape_vector)
+
+            # Pad with 1's at the beginning
+            call_group_shape_vector = [1] * missing_dims + call_group_shape_vector
 
         # Validate that all call_group_shape values are >= 1
         for i, dim in enumerate(call_group_shape_vector):
@@ -428,8 +438,14 @@ def generate_code(
                 )
 
         # Calculate call group size as product of all dimensions
-        for dim in call_group_shape_vector:
+        # Also grab the group strides here as that will allow us
+        # to use the group shape as constants to improve perf
+        call_group_strides = []
+        for dim in call_group_shape_vector[::-1]:
+            call_group_strides.append(call_group_size)
             call_group_size *= dim
+        call_group_strides.reverse()
+
         # Check if call_group_size exceeds hardware limits
         if call_group_size > 1024:
             raise KernelGenException(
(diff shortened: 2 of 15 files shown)


sam-test/call-id-test.py
```diff
@@ -0,0 +1,233 @@
...
+[DEBUG]   Call type: call
+[DEBUG]   Call shape: [32, 64]
+[DEBUG]   Call mode: prim
+[DEBUG]   Strides: [64, 1]
+[DEBUG]   Threads: 2048
+[DEBUG]   Call grid shape: [8, 8]
+[DEBUG]   Call grid strides: [8, 1]
+[DEBUG]   Call group shape: [4, 8]
+[DEBUG]   Call group strides: [8, 1]
+[[[   0    0    0]
```
kaizhangNV: I don't get how the shape impact the thread index layout here?

sam-test/call-id-test.slang
```diff
@@ -0,0 +1,165 @@
...
+    uint flat_call_id;
+
+    // 1-dimensional version of call_group_id
+    uint flat_call_group_id;
+
+    // 1-dimensional version of call_group_thread_id
+    uint flat_call_group_thread_id;
+}*/
+
+// can't seem to use int2 or int[2] anywhere
```
kaizhangNV: why?

slangpy/core/callsignature.py
```diff
@@ -394,18 +394,72 @@ def generate_code(
     """
     nodes: list[BoundVariable] = []
 
+    cg.header = ""
     # Generate the header
     cg.add_import("slangpy")
+
+    # Specify some global-scope static variables here
```
kaizhangNV: I think we should do that.
But I don't understand, why this has to be global static (thread-local) variable.
Why can't it be a shader parameter?

slangpy/core/callsignature.py
```diff
@@ -477,23 +531,40 @@ def generate_code(
...
+        cg.kernel.append_line("[numthreads(32, 1, 1)]")
+    # TODO: Should probably be consistent between camel case and underscore naming styles here.
+    # Note: While flatCallThreadID is 3-dimensional, we consider it "flat" and 1-dimensional because of the
+    #       true call group shape of [32, 1, 1] and only use the first dimension for the call thread id.
+    cg.kernel.append_line("void compute_main(uint3 flatCallThreadID: SV_DispatchThreadID, uint flatGroupThreadID: SV_GroupIndex)")
     cg.kernel.begin_block()
-    cg.kernel.append_statement("if (any(dispatchThreadID >= call_data._thread_count)) return")
+    cg.kernel.append_statement("if (any(flatCallThreadID >= call_data._thread_count)) return")
+
+    # Calculate the ids that we'll use to compute the call id. We'll also store these in
```
kaizhangNV: Though I'm 100% sure if it's possible, I really think all of these logic should be in a slang module with generic, and we can use link-time specialization to specialize these values.

IMO, we should always avoid use run-time code generation as much as possible unless we have to.

slangpy/slang/callidarg.slang
```diff
@@ -48,3 +48,16 @@ extension<T : __BuiltinIntegerType> VectorizeCallidArgTo<T, -1> {
 extension<T : __BuiltinIntegerType> VectorizeCallidArgTo<T, 1> {
     typealias VectorType = T;
 }
+
+// Sticking these here for now but they could be put into another
+// file if desired
+// vector<int, N> or int[N] for these???
+// Doesn't seem possible to get the generics to work
```
kaizhangNV: Have you figured out why the generic doesn't work?
Usually when your dimension bigger than 4, you can use array (int[N]), otherwise just use vector.
All of those generics need to be specialized during link time.

sam-test/call-id-test.py
sricker-nvidia: It probably helps to compare it to the above example that just uses the default "linear" dispatch style. Using a 2D example of 64x32 image (using [x,y] here), we'll move through that image in groups of 32 threads such that the first group goes through [0-31, 0], the second group goes through [31-63, 0], the next [0-31, 1] etc. Essentially each row of the image is broken up into two thread groups that each go through 32 pixels in a linear fashion. That's what gives us this sort of pattern:

 // [pixel x, pixel y, flat/1D thread id]

// These values correspond to the end of the first row of pixels
  [  61    0   61]
  [  62    0   62]
  [  63    0   63]

// These values correspond to the beginning of the second row of pixels
  [   0    1   64]
  [   1    1   65]
  [   2    1   66]

Specifying a thread/call group shape allows the groups of 32 to also be 2 dimensional. Having the same image but using a call group shape of [8,4] (x and y are flipped in the debug print outs list above) will instead have the first group of 32 threads go through pixels [0-7, 0], [0-7, 1], [0-7, 2], and [0-7, 3], the second [8-13, 0], [8-13, 1], [8-13, 2], [8-13, 3], and group 8 will end up accessing pixels [0-7, 4], [0-7, 5], [0-7, 6], [0-7, 7].

It might be easier to visualize with a smaller example. If I have an 8x4 image, groups of size 4 and a 2x2 group shape, the linear case for group 0 would like this:

(using "n" as the threads in the group currently accessing pixels and "o" as untouched pixels)

nnnnoooo
oooooooo
oooooooo
oooooooo

and the call group shape case looks like:

nnoooooo
nnoooooo
oooooooo
oooooooo

for group 4, linear looks like:

oooooooo
oooooooo
nnnnoooo
oooooooo

and for our call group:

oooooooo
oooooooo
nnoooooo
nnoooooo

sam-test/call-id-test.slang
sricker-nvidia: If I recall correctly I think this was related to a numpy limitation. As in using `RWNDBuffer<int2, 3>` as an output type would cause an error when trying to read back results, but having a shader function return int2 was fine.

I'm not sure if this is some known slangpy limitation or if I was just doing something wrong, but I moved on and just kept things simple in these initial test functions so I could have some minimal sanity checks.

slangpy/core/callsignature.py
sricker-nvidia: Correct me if I'm wrong here as I'm not a slang expert yet, but my understanding was that shader parameters are passed from the host application code into GPU code. The id's here need to be calculated based on a thread's flat/1D thread id at run time, so I don't think we can pass them in as shader parameters.

It might make more sense when I upload the next iteration of the patch where I've updated this and have the get_call_id() type functions working.

slangpy/core/callsignature.py
sricker-nvidia: I guess you could have a generic function in the slangpy module and pass a bunch of things in to do these computations there, but I'm not sure I understand how it would be cleaner or more performant.

If we really want to move a lot the code generation in this python function into a slang module I'd prefer to do it outside this change, but maybe others will feel different.

slangpy/slang/callidarg.slang
sricker-nvidia: yeah, I have it working in the next iteration that I'll be uploading later today. I'm not sure how "clean" that solution is, but maybe someone will see the intent of it and have a suggestion for improvement.

I think there was originally some other issue that I was hitting that I incorrectly assumed was related to  my attempts at using generics.

sam-test/call-id-test.slang
kaizhangNV: I see, can you try to rebase your tree. I think I resolve some of the issues.
Also, when you try to read back result, you need to use buffer cursor, instead of just to_numpy.
Check the code at  
```
slangpy/tests/slangpy_tests/helpers.py: write_ndbuffer_from_numpy and read_ndbuffer_from_numpy
```

slangpy/core/callsignature.py
kaizhangNV: You understanding is correct, and this helps me to understand more about the situation.
I think we should use link-time specialization in this case, but this is orthogonal to my original question (thread-local VS shader parameter).

sam-test/call-id-test.slang
kaizhangNV: There are some alignment issues in the numpy to GPU buffer.

Our NDBuffer doesn't handle alignments, the reason is that we allow an NDBuffer to bind to many different Slang types, like NDBuffer( shape(2, 2, 3, 4) ), can be float4 with shape (2, 2, 3), it can be float3x4 with shape(2,2), it can also be array float[4] with shape(2,2,3). But NDBuffer doesn't handle all of those. There is still room for the refactor on this. We can pass slangType to C++ binding when creating the NDBuffer, so when they create the GPU buffer, they will know exactly how large the buffer will be and handle the alignment issues more properly. But now, they are just always using raw buffer which is more efficient but can't handle alignment.

sricker-nvidia:
This PR is almost ready to be moved out of the "draft" state, I still need to double check some of the testing, remove some of the personal test files, updated the commit message etc.

I was able to test this against https://github.com/elcuppietan/slangpy-samples/tree/wave-ballot/experiments/balloted-splatting using the attached patch:
[call_group_shape_edits.patch](https://github.com/user-attachments/files/20664876/call_group_shape_edits.patch)

and things seem to work as expected (or at least run the same as the original sample). That said I don't notice a performance difference. I'm not sure if that's expected or not.

sam-test/call-id-test.py
kaizhangNV: I see, so the last element is still be the global thread id.

slangpy/core/callsignature.py
kaizhangNV: having everything in slang, and let compiler to do the specialization/code generation will always be our first option.
this run-time code generation will be the last option.

ccummingsNV: Whilst I am sure there is scope for use of specialization, we can't move thread count to link time - it's decided per dispatch, and we (deliberately) don't bind a generated kernel to a given dispatch size. i.e. it has to be a shader uniform.

slangpy/core/callsignature.py
```diff
@@ -477,23 +532,55 @@ def generate_code(
...
+    # global-scope static variables
+    if call_data_len > 0:
+        # Calculate flattened id's that will be used to determine the N-dimensional
+        # id's.
+        cg.kernel.append_statement(f"flat_call_id = flatCallThreadID.x")
+        cg.kernel.append_statement(f"flat_call_group_id = flatCallThreadID.x / {call_group_size}")
```
ccummingsNV: I still don't see why this (and the line below) aren't using the constants provided by the driver to get group id / group thread id.

slangpy/core/callsignature.py
```diff
@@ -394,18 +394,73 @@ def generate_code(
...
+        cg.add_snippet("flat_call_group_id", f"static uint flat_call_group_id = 0;\n")
+        cg.add_snippet("flat_call_group_thread_id", f"static uint flat_call_group_thread_id = 0;\n")
+
+        # TODO: Ideally there would be a way to just return call_id here so we could avoid wasted effort itterating.
+        #       Not enough of a slang expert at the moment to know what that could/should look like.
+        cg.add_snippet("get_call_id", f"export public static int[N] get_call_id<let N: int>() {{ int[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_id[i]; }} return ret; }};\n")
+        cg.add_snippet("get_call_group_id", f"export public static uint[N] get_call_group_id<let N: int>() {{ uint[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_id[i]; }} return ret; }};\n")
+        cg.add_snippet("get_call_group_thread_id", f"export public static uint[N] get_call_group_thread_id<let N: int>() {{ uint[N] ret;  for(int i=0; i<{call_data_len}; i++) {{ ret[i] = call_group_thread_id[i]; }} return ret; }};\n")
+
+        cg.add_snippet("get_flat_call_id", "export public int get_flat_call_id() { return flat_call_id; };\n")
```
ccummingsNV: Not sure about these flat options - they aren't provided in the classic dispatch grid, so it seems odd that we'd assume they're necessary for our version.

slangpy/core/callsignature.py
ccummingsNV: The use of thread-local is in response to user feedback, that plumbing the call id through multiple layers is creating a lot of friction. As such, we're aiming to shift to a global model for this small amount of data that identifies a thread.

slangpy/types/callgrouparg.py
```diff
@@ -0,0 +1,241 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from slangpy.bindings import (
```
ccummingsNV: I can see these have been useful in testing, but I'm really nervous about adding them as part of the main api. I think we should keep these guys in a branch, and focus on making the new global api good, otherwise it's just going to be confusing for users.

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -377,16 +378,73 @@ nb::object NativeCallData::exec(
...
     for (auto it = cs.rbegin(); it != cs.rend(); ++it) {
-        strides.push_back(total_threads);
-        total_threads *= *it;
+        strides.push_back(current_stride);
+        current_stride *= *it;
     }
     std::reverse(strides.begin(), strides.end());
 
+    // Get call group shape from build info
+    std::vector<int> call_group_shape;
```
ccummingsNV, line 405: Could you test out using short_vector for some/all of these. std::vector wasn't a great choice by me when there were just a couple of locals. Now there's a fair few, I think these allocations are going to start showing up on the profiler. short_vector with a pre-allocated buffer of (eg) 32 should be much more performant.

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -423,11 +485,22 @@ nb::object NativeCallData::exec(
...
         log_debug("Dispatching {}", m_debug_name);
         log_debug("  Call type: {}", command_encoder ? "append" : "call");
         log_debug("  Call shape: {}", call_shape.to_string());
         log_debug("  Call mode: {}", m_call_mode);
         log_debug("  Strides: [{}]", fmt::join(strides, ", "));
+        log_debug("  Call grid shape: [{}]", fmt::join(call_grid_shape, ", "));
+        log_debug("  Call grid strides: [{}]", fmt::join(call_grid_strides, ", "));
+        log_debug("  Call group shape: [{}]", fmt::join(call_group_shape, ", "));
+        log_debug("  Call group strides: [{}]", fmt::join(call_group_strides, ", "));
+        if (is_call_shape_unaligned) {
```
ccummingsNV, line 567: This has triggered a couple of alarm bells we need to think through. In the current system, with threads [1,1,32], you can at most waste 31 threads, regardless of call shape/size. Is that still the case?

I'm wondering if the scenario of a fixed group shape defined by the user does need treating differently to a 'null' group shape in which grouping is irrelevant.

slangpy/core/callsignature.py
ccummingsNV: fyi, we have discussed allowing the user to explicitly say "optimize for a fixed dispatch size", in which case certain things could be made constant (regardless of whether via code gen or link time specialization).

kaizhangNV: I'm pretty sure that we support something like this

```
extern static const uint numthread = 0;
[numthreads(numthread, 1, 1)]
```

where `numthread` is link-time constant.

But I'm not saying we will have to do this in current slangpy implementation. It's just FYI.

sam-test/call-id-test.py
sricker-nvidia: yep, this was one of my initial manual sanity tests and printing it out this way was just convenient.

sam-test/call-id-test.slang
sricker-nvidia: I plan on removing these files from the change, so I don't know if it's worth the effort. I originally included them in the draft PR so folks could have an idea of how things were intended to work, but that should also be apparent in the new pytest tests that I've added as well, so these are no longer really necessary (and I don't think we want to merge in a directory called sam-test anyway)

slangpy/core/callsignature.py
ccummingsNV: Ah - yeah for group size we could probably use link time constants I suspect, though not urgent. We do have a test somewhere in the core sgl code for that use case.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: The case in the current PR:

-For the default case of not specifying the call group shape, we fall back to linear dispatches that function the same as the current slangpy implementation. That is, you will waste as most 31 threads.

-For the case of call group shapes, the PR will currently detect if the call shape is not aligned to the call group shape and align up as necessary. In the worse case, you would have a call shape with a remainder of 1 after dividing by the call group shape in each dimension and the waste would get larger depending on the size of the call shape.

I'm not sure what a good alternative is to aligning up the number of threads. I don't think it makes sense to "tear" a call group with parts of it in one section of the call shape and other sections in a completely disjoint section of the call shape. The seems like it would defeat the purpose of memory coherency etc. Another option would be to enforce alignment between call group shapes and call shapes, but my thinking was that would reduce the number of use cases for the new functionality.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: Maybe easier to use std::vector::reserve() here? Since we know the dimensionality at this point we can just use reserve() to pre-allocate the exact size for all of these vectors and not worry about the underlying allocation implementation possibly doubling the size of the vector when we run out of space.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: Another idea might be to down align the call shape and then run some separate kernels where the group shape and call shape are modified to fit the trimmed off bits of the call shape, but that seems like it would get rather complicated fairly quickly. I think it would look something like:

````
for (int i= 0; i< call_shape.size(); i++)
{
    // we will create 1 new call shape and group shape per dimension
   // This currently assumes each dimension is unaligned.
    std::vector<int> new_call_shape;
    std::vector<int> new_call_group_shape;
    // The new shape/length for dimension "i" is the remainder
    // ex: for 66x32 and 4x4 the remainder for dim 1 is 2
    new_call_shape[i] =call_shape[dim] % call_group_shape[dim];
    // Match the new call shape for this dimension as it should be less
    // than the original group size.
    new_call_group_shape[i] = new_call_shape[i];
    // Now we need to set the length for the other dimensions of
    // this new call shape.
    for (int j= 0; j< call_shape.size(); j++)
    {
         if ( i == j)
             // Already set, move on
             continue;
         // Match the original dimensions for the other dimensions for the new call shape.
         new_call_shape[i] =call_shape[dim];
         // Match the original dimensions for the other dimensions for the new call group shape.
         new_call_group_shape[i] =call_group_shape[dim];
    }

     // Generate new kernel calls using
     // new_call_shape and new_call_group_shape
     // probably also need to pass some "offset"
}

// If more than one dimension was unaligned, there will be an overlapping
// trimmed area (think of the corner of a picture frame) that was not captured above.
// calculate one more call shape and group shape to handle this
// Again we are assuming unalignment of all dimensions for simplicity of this example.
std::vector<int> new_call_shape;
std::vector<int> new_call_group_shape;
for (int i= 0; i< call_shape.size(); i++)
{
    new_call_shape[i] = call_shape[i] % group_shape[i];
}

// Do something with new_call_shape and new_call_group_shape
// probably also need an offset of some sort
````

I "think" something along those lines would avoid wasting threads entirely and the number of dispatches would be equivalent to the number of dispatches made when aligning up.

File: slangpy/types/callgrouparg.py (hunk 8 above)
Comment (sricker-nvidia) [reply slangpy/types/callgrouparg.py]: Sure, I can remove these from the main change for now then.

File: slangpy/core/callsignature.py (hunk 7 above)
Comment (sricker-nvidia) [reply slangpy/core/callsignature.py]: I'll remove them for now then.

File: slangpy/core/callsignature.py (hunk 6 above)
Comment (sricker-nvidia) [reply slangpy/core/callsignature.py]: Our current "linear" flat thread id to N-dimensional mapping logic operates as if each thread conceptually has its own call group, even though they actually run in groups of 32.

To match the current mapping logic to the new calculations below, flat_call_group_id and flat_call_group_thread_id need to reflect the conceptual group size and not the actual group size, so I cannot use the driver variables SV_GroupID.x and SV_GroupThreadID.x (I assume these are the driver constants being referred to here) because these would reflect the actual group size of [32, 1, 1] rather than the conceptual size of [1, 1, 1]. The actual group size could be changed to [1, 1, 1] for this case such that SV_GroupID.x and SV_GroupThreadID.x could be used here, but I assume that would be far less performant as a result of increased dispatch overhead (more dispatch calls) and reduced parallelism.

Alternatively, we could use SV_GroupID.x and SV_GroupThreadID.x and [32, 1, 1] in the default case, when call group shape is not specified, but then we need to deal with cases where the call shape is not aligned to [32, 1, 1]. Currently with my changes, a 65x32 call shape using a group shape of 32x1 would result in padding the call shape to be 32x1 aligned giving a 96x32 call shape where 31*32 threads would be wasted, where as no threads are wasted when we do not specify call_group_shape.

**Comment by sricker-nvidia** [comment]
Noting that I see some warnings in the CI:
````
slangpy/tests/slangpy_tests/test_call_group_integrations.py ....::warning file=/opt/ci/actions-runner/_work/slangpy/slangpy/slangpy/tests/slangpy_tests/helpers.py,line=168::Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
Warning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
````

File: slangpy/core/calldata.py [hunk 11]
```diff
@@ -88,6 +88,8 @@ def __init__(
             positional_mapping = build_info.map_args
             keyword_mapping = build_info.map_kwargs
             type_conformances = build_info.type_conformances
+            if build_info.call_group_shape is not None:
```
ccummingsNV: I suspect the need for this check is that your nanobind binding doesn't support None. You've got 2 options here:
- make the nanobind binding support 'none' (I think it's .none() on the attribute - there's other examples) and make call_group_shape std::optional
- use the fact that shape supports being 'invalid' (I think!)

Either way, as general practice I've tried to make sure that a Python class fully initializes the native class it inherits from. If it's easy to do, it'd be preferable for this code to do the same.

slangpy/core/callsignature.py
```diff
@@ -396,16 +396,100 @@ def generate_code(
...
+        # for now
+        cg.add_snippet("call_id", f"static int[{call_data_len}] call_id;\n")
+        cg.add_snippet("call_group_id", f"static uint[{call_data_len}] call_group_id;\n")
+        cg.add_snippet(
+            "call_group_thread_id", f"static uint[{call_data_len}] call_group_thread_id;\n"
+        )
+
+        # TODO: Ideally there would be a way to just return call_id here so we could avoid wasted effort itterating.
+        #       Not enough of a slang expert at the moment to know what that could/should look like.
+        cg.add_snippet(
```
ccummingsNV: I'm surprised you can't just return the values directly, assuming they're exactly the same type. Ideally we solve this before shipping as it could have perf impacts. If there absolutely must be a loop, we should probably mark it as 'force unroll'.

slangpy/core/callsignature.py
```diff
@@ -396,16 +396,100 @@ def generate_code(
...
+        # and the call group shape are not aligned. When a thread's call id
+        # falls outside the call shape, we need it to return early. This is
+        # similar to the default linear case when the call shape size is not
+        # 32 thread aligned.
         cg.call_data.append_statement(f"int[{call_data_len}] _call_dim")
+        # A group can be thought of as a "window" looking at a
+        # portion of the entire call shape. Grid here refers to the
+        # N dimensional call shape being broken up into some number of N
+        # dimensional "window"s / groups.
+        cg.call_data.append_statement(f"uint[{call_data_len}] _grid_stride")
```
ccummingsNV: It's unclear to me what 'grid' refers to - is that the overall call shape?

slangpy/core/callsignature.py
ccummingsNV: I see, I think I'm with you. So the issue is that in the none-specific case in which we want to choose something automatically, the math would tie group shape to the right-most dimension? That is a bit surprising to me, but I've not worked out the math in detail myself - will take a peak today.

Assuming this is unavoidable, it seems to me that the correct fix then would be to generate different, simpler code in the situation that there was no explicit group shape. We could perhaps do that in a 2nd pass, along with the other optimization in which we explicitly use the hardware group if possible?

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: I think we'd be better off in practice using std::short_vector with a decent sized buffer. reserve is still a malloc, and these mallocs are starting to add up (during optimization of neural shading I've been progressively trying to reduce them)

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: My current thinking is:
- When user explicitly specifies a call group shape, it's their responsibility to avoid waste
- When user doesn't explicitly specify a call group shape, it's our responsibility to minimize waste

This one probably needs talking through together though. If you're confident you have something that works well, we can save it for phase 2.

slangpy/core/callsignature.py
```diff
@@ -483,23 +567,69 @@ def generate_code(
...
-    cg.kernel.append_statement("if (any(dispatchThreadID >= call_data._thread_count)) return")
+    cg.kernel.append_statement("if (any(flat_call_thread_id >= call_data._thread_count)) return")
+
+    # Calculate the ids that we'll use to compute the call id. We'll also store the
+    # N-dimensional call_*_id's as global-scope static variables such that they can
+    # be "queried" in user shaders via getters provided by the slangpy module.
+    if call_data_len > 0:
+        # Calculate flattened id's that will be used to determine the N-dimensional
+        # id's.
+        cg.kernel.append_statement(f"uint flat_call_id = flat_call_thread_id.x")
```
ccummingsNV: I agree with @kaizhangNV that this code is becoming complex enough that it might be worth implementing some slang side generic helpers that the generated code calls. However I don't think that's necessary for this PR.

slangpy/slang/callidarg.slang
```diff
@@ -48,3 +48,12 @@ extension<T : __BuiltinIntegerType> VectorizeCallidArgTo<T, -1> {
 extension<T : __BuiltinIntegerType> VectorizeCallidArgTo<T, 1> {
     typealias VectorType = T;
 }
+
+// Functions provided for use in user's shaders
+// Note: In the returned arrays, the first component represents the right-most dimension,
```
ccummingsNV: This statement seems incorrect. Generally, component 0 represents the left most dimension (with the highest stride), and this is reversed in the case of vectors.

If you want a vector version, we could supply get_call_id_v that returns vector<int,N>?

slangpy/core/callsignature.py
```diff
@@ -396,16 +396,100 @@ def generate_code(
...
+
+    # Specify some global-scope static variables here to allow users to make calls
+    # to functions like get_call_id<N>() etc in order to access call_id,
+    # call_group_id, and call_group_thread_id.
+
+    # Note: Support for static global variables should be seen as a legacy feature
+    #       and its use is discouraged. As such, this may need to be reworked in
+    #       the future.
+    if call_data_len > 0:
+        # Context wants call_id as an int for some reason, so respect that here
```
ccummingsNV: I think we should be consistent either way. I suggest we use int for all of them (to avoid you having to change existing systems and worry about breaking something) for now, and can look to changing all of them to uints later if we feel it's worth doing. In this case, I think 50/50 is worse than using int.

slangpy/tests/slangpy_tests/test_call_group_integrations.py
```diff
@@ -0,0 +1,906 @@
...
+import "slangpy";
+
+[Differentiable]
+float tensor_add_with_call_groups(
+    float a,
+    float b,
+    uint2 grid_cell
+) {
+    // Use call group functions in differentiable context
+    uint[2] call_group_id = get_call_group_id<2>();
```
ccummingsNV: This one's a little concerning - having to specify the dimensionality isn't great from a user perspective.

slangpy/core/callsignature.py
```diff
@@ -396,16 +396,100 @@ def generate_code(
...
+        # for now
+        cg.add_snippet("call_id", f"static int[{call_data_len}] call_id;\n")
+        cg.add_snippet("call_group_id", f"static uint[{call_data_len}] call_group_id;\n")
+        cg.add_snippet(
+            "call_group_thread_id", f"static uint[{call_data_len}] call_group_thread_id;\n"
+        )
+
+        # TODO: Ideally there would be a way to just return call_id here so we could avoid wasted effort itterating.
```
kaizhangNV: This can be done perfectly by using link-time constant.
assume `call_data_len` is a link-time constant., e.g.

```
extern static int call_data_len = 0;

static int[call_data_len] func() { return call_id;}
```
because slang support link-time sized array.

But you don't need to do in this PR as we discussed.

sricker-nvidia: Thanks Kai, I will make a follow up task to look into this. I suspected something like this might exist, but I had a lot of difficulty getting the syntax correct.

slangpy/core/callsignature.py
sricker-nvidia: Sure, I'll just swap them all to int for consistency for now.

slangpy/core/calldata.py
sricker-nvidia: I'll double check this one. I remember having issues previously that prevented me from removing the check, but I could have messed up the nanobind stuff.

slangpy/core/callsignature.py
sricker-nvidia: I think Kai may have provided a solve for this in another comment. I'll try to check that today. I suspect that the suggestion here is possible, but I don't know enough of slang's syntactic sugar yet.

slangpy/core/callsignature.py
sricker-nvidia: When we have a call group shape, that breaks the original call shape into some number of groups. The shape of these groups is what I refer to as the "grid" here.

ex: for a 8x8 call shape with a call group shape of 4x4, we get a grid shape of 2x2, as the call shape can be split in 2 groups horizontally and 2 groups vertically.

- group dim -> the N dimensional shape of threads within a group
- grid dim -> the N dimensional shape of groups within a call.

slangpy/core/callsignature.py
sricker-nvidia: Agreed, I can create some follow up tasks.

slangpy/slang/callidarg.slang
sricker-nvidia: To be honest the flipping directions of int[]s vs vectors has been greatly confusing me, but I'll try to explain and hopefully it makes sense.

In slangpy\src\slangpy_ext\utils\slangpy.cpp we set things like call_data._call_dim, call_data.grid_dim, etc based on our arrays like `const std::vector<int>& cs = call_shape.as_vector();`. Based on the fact that we need to traverse these arrays in reverse order to calculate strides (which we then reverse), I'm assuming that all of the arrays that we pass in to call_data are oriented such that component 0 is the left most dimension with the highest stride.

Meaning that the call_id calculations in this change are setting call_id[i] based on call_data.grid_dim etc and that call_id will then also have the "component 0 is the left most dimension with the highest stride" orientation.

The confusion here probably stems from my terminology, which itself stems from my confusion trying to nail down these orientations. When I refer to dimensions here, I'm thinking of dimensions in terms of [x, y, z, ...] and strides of [1, w, w*h, ...] as this is how I've always oriented things with graphics.

So when I say component 0 represents the left most dimension, I mean "y" or "the dimension with the highest stride".

slangpy/tests/slangpy_tests/test_call_group_integrations.py
sricker-nvidia: I might be helpful for me to see some basic example of a user shader where the call_id is used without knowing the dimension. I'm having trouble understanding why that would work. Not that it wouldn't, I just don't have enough experiment with slang shaders yet for this concept to feel intuitive.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: hmm, I'm not sure I follow then. Is std::short_vector not doing mallocs? I also didn't see a std::short_vector in c++ docs, but maybe I'm missing something?

I'm happy to change it, but I don't quite understand the rational yet.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: I believe that was my thinking as well. A responsible user can avoid waste by making sure the call shape has call group shape alignment (as in 65x32 call shape with 32x1 call shape group will be unaligned.). However, we will not block the user from making less optimal calls that waste threads, but we will print debug logs informing of the mis alignment and suggest fixing.

For the other bullet point, when call shape is not specified or is specified with Shape(()), the change defaults to the current way of doing things with linear dispatches of 32 threads and the amount of waste should be no different than what a user is currently experiencing.

The above would be a lot of additional work and better for a phase 2 if it's something that we really want. I'm also not sure how much performance it would recapture, so there would be some investigative / exploration type work to do there.

slangpy/core/calldata.py
sricker-nvidia: I was able to update this and remove the if check as suggested in the latest.

slangpy/core/callsignature.py
sricker-nvidia: Looked a bit more into this one, an issue seems to be that slang needs to know array sizes at compile time, so using the above and doing something like `int[2] call_id = func()" generates compile errors. There's a definite possibility that I'm still misunderstanding something here, but for now I'm not able to update this unfortunately.

slangpy/core/callsignature.py
sricker-nvidia: I tried out Kai's suggestion above, but I get lot of compile errors trying to use a link time constant to specify an array size. It may still be possible to fix this up, but I don't have time now to investigate further unfortunately.

sricker-nvidia:
Resolved the review comments that I could. I added another perf test under https://github.com/sricker-nvidia/slangpy/tree/call-id-improve-perf-test/slangpy/tests/call_groups that just tests the default linear case for the sake of being able to compare the PR against ToT. This is what some of the results looked like:

default only perf test
	-Note that this seems faster than the default case in call_group_performance_test.py
	 likely because get_call_id<2>() is slow and it is not used here (spy.call_id used instead)
	-PR
		Baseline Performance: 3.589 ± 0.369 ms
		Baseline Performance: 3.535 ± 0.305 ms
		Baseline Performance: 3.463 ± 0.046 ms
		Baseline Performance: 3.494 ± 0.188 ms
		Baseline Performance: 3.548 ± 0.319 ms

	-ToT
		Baseline Performance: 3.524 ± 0.211 ms
		Baseline Performance: 3.522 ± 0.173 ms
		Baseline Performance: 3.464 ± 0.083 ms
		Baseline Performance: 3.519 ± 0.223 ms
		Baseline Performance: 3.542 ± 0.289 ms

ToT seems to wins out slightly because in the PR we still populate call_group_id and call_group_thread_id,even in the default case leading to a small perf hit. However, my most recent runs are showing around 3.680 ± 0.440 ms, so it seems like I may have made an edit somewhere that reduced things back a bit. It could also just be noise/variance.

I also revisited the balloted-splatting example and had some interesting observations:
ballot splat perf
	-with PR
		-no use of call group shapes (default linear dispatch)
			-27 it/sec
		-using call group shapes
			-3.17it/sec
	-ToT
		-no use of call group shapes (default linear dispatch)
			-27 it/sec

So it looks like I was mistaken when previously I thought I saw no difference in perf in that sample, the call groups seem to be making things slower, even though the sample still has correct output. I don't fully understand what the sample is doing, so there could be something wrong with the edits in my patch adding in call groups.

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: std::short_vector is implemented in the slangpy code base. It contains local storage for <= N elements. So the creation of one on the stack with, say, storage for 16 elements will mean that any operation with less than 16 dimensions would result in no heap allocations.

slangpy/slang/callidarg.slang
ccummingsNV: Sure - I appreciate the confusion, but it's Python, and numpy and every ML system everywhere uses the standard that left most == index 0 == highest stride. We inverted it for vectors so that when interacting with, eg, graphics apis such as Texture.Load the classic logic still works.

The reason the statement seemed incorrect is that you say "the first component represents the right-most dimension", which is wrong (or should be!).

kaizhangNV:
I think the perf issue is just because we use the NDBuffer<float, 2> instead of 2D texture as the input.
1×32 groups will win, because it has the most cache friendly access pattern.
NDBuffer is actually still 1-dimensional buffer though its name is `ND`. But texture is different because the memory layout is designed to cache a 2/3-D block in the hardware.

sricker-nvidia:
Did some digging today. The perf issue on the balloted-splatting example seems to be the result of the newly added call-id calculations in the generated kernel code.

I tested by rewriting parts of the PR such that everything functions as before, but get_call_id will still return the newly calculated, non linear call id based on the given call group shape. That allowed me to tinker with using the pre computed call-ids and the call group shape call ids in the same function call. Using the value from get_call_id() gives the correct visual result, but slows things down. If the value from get_call_id() is used in a way that can be optimized out by the compiler, things speed up.

I haven't thought too deeply about possible solutions yet. One might be to try and pre calculate dispatch id's in NativeCallData::exec and send them in as a uniform, but that seems like it could have it's own perf impacts. Another idea is to somehow store the id's that threads calculate and reuse those if the same function call is made with the same call group shape. The most realistic idea I think of is to implement what was suggest on the issue, "We could also optimize where appropriate (or when hinted to by the user), and map call groups directly to thread groups where possible, which would simply reduce the need for arithmetic operations". I would expect that to get perf back on par, but would only be a good solution for up to 3D calls, but maybe that's ok for now?

Perf should in theory get worse with higher dimensionalities due to additional arithmetic operations required for each additional dimension. Not sure if there's a good way around that unless ids are precalculated and reused in multiple passes.

src/slangpy_ext/utils/slangpy.cpp
sricker-nvidia: I was able to update a few of these, but I couldn't change all of them due to shape.as_vector() returns a vector and short_vectors don't currently seem like they can be initialized with a vector.

slangpy/core/callsignature.py
sricker-nvidia: Revisited this and I'm more convinced now that link time constants could be used, but using them are currently blocked on issues like https://github.com/shader-slang/slang/issues/7509 and we'll need to update the version of slang that we use in slangpy in order to get the fixes.

sricker-nvidia:
Uploaded another update patch integrating changes from https://github.com/shader-slang/slangpy/compare/kai/dev/call-shape-link-time (thanks @kaizhangNV !). The perf issue on the balloted-splatting example has been resolved and was due to dynamic indexing in some for loops. Using [unroll] to force static indexing fixed the issue.

sricker-nvidia:
Also uploading a new patch for the balloted-splatting example using call group shapes
[call_group_shape.patch](https://github.com/user-attachments/files/20957473/call_group_shape.patch)

Performance for ToT balloted-splatting example with ToT slangpy: runs in 05:55
Performance for balloted-splatting example using call group shapes and current version of the PR: runs in 06:01

For additional comparison the updates_1 patch ran in 5:57

slangpy/core/callsignature.py
```diff
@@ -396,20 +396,107 @@ def generate_code(
...
     generate_constants(build_info, cg)
 
+    # Generate additional link time constants definition code. These are declared in callshape.slang
+    # and used to generated call_ids that can be queried by user modules.
+    cg.constants.append_statement(f"export static const int call_data_len = {call_data_len}")
+    cg.constants.append_statement(f"export static const int call_group_size = {call_group_size}")
+
+    # Also generate the call group shape and stride arrays as link time constants. Using constants
+    # should yield better performance than passing these in as uniforms.
+    cg.constants.append_line(f"export static const int[call_data_len] call_group_strides = {{")
```
ccummingsNV, line 468: We don't want to have to regenerate a kernel every time a tensor of a different shape is used - only when a kernel of different dimensions is used. Hence why the call dimensions can be 'hard coded' into the kernel, but actual call shape must be passed as uniforms.

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: Fair enough - we should probably clean that up at some point, but it's for  another PR another day :)

slangpy/core/callsignature.py
sricker-nvidia: This is the group shape though and not the call shape. The call shape is still a uniform below represented by _call_dim. Not sure if I'm missing something.

sricker-nvidia:
> It's looking good, but unless I misunderstand, the changes seem to have made the kernel dependent on precise call shape. i.e. for a user to change call shape, they have to compile a new kernel.

I don't think that should be the case, but feel free to correct me if I'm misunderstanding something. With this PR, the kernel is dependent on the call **group** shape, but not the call shape. Call shape has no impact on the **group** shape. However, changing the **group** shape or the **call** shape will have an impact on the **grid** shape, but like the call shape and call strides, grid shape and grid strides are passed in as uniforms.

For example, code like this:

`module.test2D(spy.grid((100,3)), _result='numpy')`
and
`module.test2D(spy.grid((500,5)), _result='numpy')`

would generate the same kernel as they have the same call group shape, despite having different call shapes. Code like this:

`module.test2D.call_group_shape(Shape((4,8)))(spy.grid((100,3)), _result='numpy')`
and
`module.test2D.call_group_shape(Shape((4,8)))(spy.grid((500,5)), _result='numpy')`

would also generate the same kernel, as again, they have the same call group shape, despite having different call shapes. Code like this however:

`module.test2D.call_group_shape(Shape((4,8)))(spy.grid((100,3)), _result='numpy')`
and
`module.test2D.call_group_shape(Shape((3,3)))(spy.grid((100,3)), _result='numpy')`

would generate different kernels because the call group shape changed.

If that's undesirable I can look into further updates to pass in call group shape and strides as uniforms similar to call shape, stride etc, but I think that would have a fairly negative impact on the current implementation.

For example, I don't think we would be able to use
`cg.kernel.append_line(f"[numthreads({call_group_size}, 1, 1)]")`

unless it's possible to do something like

`cg.kernel.append_line(f"[numthreads(call_data._call_group_size, 1, 1)]")`

but I'm not aware of such a possibility so I think we would be required to go back to using

`cg.kernel.append_line(f"[numthreads(32, 1, 1)]")`

by default. If that's the case, then using things like SV_GroupIndex (flat_call_group_thread_id, the 1D id of a thread within a group) no longer works. For example:

- With call shape -> 25x25 and group shape -> 5x5, gives a grid shape of 5x5
    - Case [numthreads(25, 1, 1)]
        - no wasted threads, SV_GroupIndex will give an accurate id in the range [0, 24]
    - Case [numthreads(32, 1, 1)]
        - SV_GroupIndex gives an id in the range [0, 31]
        - Some threads get the wrong call group thread id
            - Can return if id > 24, but that would result in 7 wasted threads per call group
            - Can manually calculate flat_call_group_thread_id = SV_DispatchThreadID.x % call_data._call_group_size}
                - Would still result in 17 wasted threads ((25x25)%32) and add extra calculations

sricker-nvidia:
Uploading another patch for the balloted-splatting example. It's not that much different than the previous except includes a new tweak for the surface format to correct a brightness / "tone" issue due to updates to slang-rhi in 

https://github.com/shader-slang/slang-rhi/commit/4706a4e764845135a114b9e45a5dbbab76caaf1d#diff-2bb77229b1029a456953e68507429bd41302d61641644a0735de1d51982b29f1

and I wanted to verify that the issue was not due to this PR.

[ballot-splat-call-group-shapes.patch](https://github.com/user-attachments/files/21026266/ballot-splat-call-group-shapes.patch)

PR #272, opened 2025-06-10
Add buffercursor ndarray write

tdavidovicNV:
Basic interface for writing multiple values into a buffer at once.

Takes a dictionary of ndarrays, where the first dimension of the ndarray must be the size of the buffer. The values are then rather blindly written into the buffer, with only a bytesize check.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -26,7 +26,7 @@ pre-commit install
 pre-commit run --all-files
 
 #Run unit tests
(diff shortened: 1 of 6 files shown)


ccummingsNV (approved):
Looks good, though if it's feasible to fix those warnings that seem to appear in the github log it'd be great. Not sure if they're new to this or not.

PR #282, opened 2025-06-12
Bitmap from non-contiguous array

skallweitNV:


src/slangpy_ext/core/bitmap.cpp
```diff
@@ -89,7 +89,46 @@ SGL_PY_EXPORT(core_bitmap)
...
+                            for (uint32_t x = 0; x < self->width(); ++x) {
+                                std::memcpy(
+                                    dst_row + x * dst_pixel_stride,
+                                    src_row + x * src_pixel_stride,
+                                    dst_component_size
+                                );
+                            }
+                        }
+                    } else if (data.ndim() == 3) {
+                        size_t src_component_size = data.dtype().bits / 8;
```
tdavidovicNV: You can use `data.itemsize()` instead (same code, but it rounds up)

src/slangpy_ext/core/bitmap.cpp
```diff
@@ -89,7 +89,46 @@ SGL_PY_EXPORT(core_bitmap)
                 if (is_ndarray_contiguous(data)) {
```
tdavidovicNV, line 89: I wanted to comment on this before, this is a bit of a misnomer. It checks for c-contiguous. It could be f-contiguous and it would still technically be contiguous. Importance depends on whether we ever see anything with the Fortran layout.. but bitmaps sometimes actually come in planes (all red, all green, all blues) which would be kinda contiguous but not in a useful way.

src/slangpy_ext/core/bitmap.cpp
```diff
@@ -89,7 +89,46 @@ SGL_PY_EXPORT(core_bitmap)
...
                     std::memcpy(self->data(), data.data(), self->buffer_size());
                 } else {
-                    SGL_THROW("data is not contiguous.");
+                    uint8_t* dst = reinterpret_cast<uint8_t*>(self->data());
+                    const uint8_t* src = reinterpret_cast<const uint8_t*>(data.data());
+                    size_t dst_component_size = DataStruct::type_size(self->component_type());
+                    size_t dst_row_stride = self->width() * self->bytes_per_pixel();
+                    size_t dst_pixel_stride = self->bytes_per_pixel();
+                    if (data.ndim() == 2) {
```
tdavidovicNV, line 100: Are there ever any 1D or 3D Bitmaps?

src/slangpy_ext/core/bitmap.cpp
```diff
@@ -89,7 +89,46 @@ SGL_PY_EXPORT(core_bitmap)
...
+                        size_t src_component_size = data.dtype().bits / 8;
+                        size_t src_row_stride = data.stride(0) * src_component_size;
+                        size_t src_pixel_stride = data.stride(1) * src_component_size;
+                        for (uint32_t y = 0; y < self->height(); ++y) {
```
tdavidovicNV, line 101: In theory, you could have contiguity in pixels but not in rows, so you could copy rows in a single memcpy.
Which is where asking for the `c` order could be more efficient. I'd at least add a comment for possible future improvement on this.

tdavidovicNV (approved):
LGTM, couple notes

src/slangpy_ext/core/bitmap.cpp
skallweitNV: Yes, this checks for c-contiguous but that's mostly what we care for in C code, no?

src/slangpy_ext/core/bitmap.cpp
skallweitNV: good point, fixed

src/slangpy_ext/core/bitmap.cpp
skallweitNV: same is true for pixel stride .. you could just have a row stride that doesn't match
if we start fixing this in other places we might want to come up with a more generic and optimized implementation, for now I guess this is fine

src/slangpy_ext/core/bitmap.cpp
skallweitNV: no

PR #283, opened 2025-06-12
Added ParameterBlock tests to ShaderCursor and fixed the writing code

tdavidovicNV:
- Added tests with ParameterBlock to test_shader_cursor.
- Fixed a bug where `ParameterBlock<T>` could only be written if `T` was a struct.
- Disabled the ParameterBlock tests, as Vulkan validation complains about the codegen (see https://github.com/shader-slang/slang/issues/7431) and the results are actually wrong (in variables that are not in the ParameterBlock).


tdavidovicNV:
This is still failing, will need closer review on MacOS

jhelferty-nv:
@tdavidovicNV Should this be moved to draft?

PR #286, opened 2025-06-13
Better handling of tensor bindings

ccummingsNV:
Changed tensor binding logic to reduce chance of attempts to bind the same resource as both d_in and d_out.

### DIFF
--- a/src/slangpy_ext/utils/slangpytensor.cpp
+++ b/src/slangpy_ext/utils/slangpytensor.cpp
@@ -129,18 +129,46 @@ void NativeTensorMarshall::write_shader_cursor_pre_dispatch(
         const ref<NativeTensor>& grad_in = primal->grad_in();
         const ref<NativeTensor>& grad_out = primal->grad_out();
 
-        if (!has_derivative()) {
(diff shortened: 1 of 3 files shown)


src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -129,18 +129,44 @@ void NativeTensorMarshall::write_shader_cursor_pre_dispatch(
...
-                write_shader_cursor_fields(context, binding, field["d_out"], grad_out.get(), read_back);
+                ShaderCursor d_out_field = field.find_field("d_out");
+                if (d_out_field.is_valid()) {
+                    bound_grad_out_buffer = grad_out->storage().get();
+                    write_shader_cursor_fields(context, binding, d_out_field, grad_out.get(), read_back);
+                }
             }
+
+            if (bound_primal_buffer == bound_grad_in_buffer || bound_primal_buffer == bound_grad_out_buffer) {
+                log_warn("Binding the same storage for primal and gradient on the same tensor. This will have serious "
```
skallweitNV, line 171: Isn't that just an error though?

PR #289, opened 2025-06-13
Draft: Reenabling bool tests

tdavidovicNV:
- Fixed a bug where call to _set_array would get type from reflection rather than ndarray
- Removed unnecessary striding in _set_matrix in BufferElementCursor.
- Added float4x3 tests to test_buffer_cursor
- Added faster memcpy codepath to shader_cursor when src and dst layout matches.
- Removed `get_scalar_type_size` as type sizes and strides need are not static (need to rely on reflection system anyway).
- Added handling for setting scalars, vectors and arrays of bools based on the backend reflection types.
- Added tests for bool2 buffers.
- Enabled bool cursor tests for CUDA and Metal.
- Added workaround for Slang CUDA-backend issue https://github.com/shader-slang/slang/issues/7441
- Added tests for bool2 reflection data, to catch when https://github.com/shader-slang/slang/issues/7441 is resolved.


tdavidovicNV:
Superceded by https://github.com/shader-slang/slangpy/pull/395

PR #290, opened 2025-06-13
Add _result='texture' and documentation

### LINKED ISSUE #69: Documentation for setting result type to 'texture'
We currently support + document various types for returning results, using '.return_type' or '_result='. Benedikt recently added 'Texture' to this, but we need to document it.

jhelferty-nv:
* Add 'texture' as possible return_type

* Add documentation page on how to use the _result argument to specify the return type for a non-void slang function.

* Add a test for _result='texture'

Fixes #69


ccummingsNV (approved):
Looks good to me - nice job.

PR #301, opened 2025-06-17
Add 'require()' to interface in print.slang

kaizhangNV:
Close #6764.

Slang require to put require() capability keyword on the interface, because we can not propagate the requirement from the implementation to interface. So the shader needs to be changed.

We need to develop mechanism for Metal specific to input the buffer length instead of using GetDimensions method to make this working on Metal.


ccummingsNV:
Change looks clean/simple enough but I don't fully understand the implications of 'require' as a keyword. Does this just cause metal to throw an error rather than something nastier when you try to use print?

kaizhangNV:
> nastier

Correct, it just report capability error when targeting Metal, no impact on other platforms.

skallweitNV:
I think this is a workaround and should be clearly marked as such. Slang should error out when using `GetDimensions` on Metal and not just silently emit that to MSL. I don't understand where exactly the problem is why the capability system cannot accomplish that right now, but that's clearly an issue in Slang IMO. So I would just ask to clearly state that in the comment, ideally with a link to the GitHub issue describing the problem. It feels like this is a substantial shortcoming of the capability system, any timeline of when this can be fixed?

skallweitNV:
Is there a ticket for the issue with the capability system? I can only find these two tickets:
- https://github.com/shader-slang/slangpy/issues/302
- https://github.com/shader-slang/slang/issues/6764

kaizhangNV:
The reason that current cap system cannot capture this is that capability check happens at type checking stage. But all the target related stuff happening at IR stages. Note we cannot check any target related stuff in type checking stage, because the AST has to be totally target independent, otherwise, slang module cannot be used at all.

But whether we can do this in IR? Maybe we can, but it will be very messy, because in IR, there is no concept of interface/struct, etc. And you will loss all of those requirement information, the only thing you can do will be some ad-hoc method to detect such kind of violation. So I think that is why we don't do that in IR at first place.

I don't think this PR is a WAR, it is just how people should use capability system. But this doesn't achieve what you need here I agree.

src/sgl/device/print.slang
```diff
@@ -43,6 +43,11 @@ interface IPrintOutput {
     /// \param total_data_count The total number of data elements in all arguments.
     /// \param[out] offset Offset into the output buffer. This needs to be passed to \c write_arg calls.
     /// \return True if the message was written successfully, false if there was not enough space.
+    /// \TODO: Note that the implementation of this interface is using RWByteAddressBuffer::GetDimensions which
+    /// has the requirement that the Metal is not supported. Slang requires the capability requirement on the interface,
+    /// so we will add this requirement here to prevent Slang generating invalid code for Metal.
+    /// See: https://github.com/shader-slang/slangpy/issues/302
+    [require(cpp_cuda_glsl_hlsl_spirv_wgsl, structuredbuffer_rw)]
```
jkwak-work, line 50: I am not sure why Metal doesn't have a function for the size, which seems useful.

Do we also remove `_metal` keyword from another function if it is not supported?
```
__intrinsic_op($(kIROp_StructuredBufferGetDimensions))
[require(cpp_cuda_glsl_hlsl_metal_spirv_wgsl, structuredbuffer_rw)]
uint2 __structuredBufferGetDimensions<T,L:IBufferDataLayout>(RWStructuredBuffer<T,L> buffer);
```

kaizhangNV: The reason metal doesn't have size is because Metal uses device pointer to represent the buffer resource type, unlike other platforms that use opaque handle.
So you cannot query the size of a pointer, while other platform can add such support in the opaque handle. This is a just language limit.

kaizhangNV: And yes, we should remove `_metal` from all of those `*GetDimensions` functions.

Issue #119, opened 2025-03-19
Cannot pass a python int to a struct `float`.


csyonghe:
```slang
// example.slang
struct Uniforms {
   float width, height;
}

float4 render(Uniforms u, int2 threadId)
{
    return 0;
}
```

```python
import slangpy as spy
import sgl
from app import App
from slangpy.types import call_id

app = App()
module = spy.Module.load_from_file(app.device, "example.slang")

u = {
   "width": app._window.width,
   "height": app._window.height
}
module.render(u, call_id(), _result=app.output)
```

Results in a marshalling error when passing `u.width` (a python int) to `Uniforms::width` (a slang float).

PR #306, opened 2025-06-25
Enable more flexible conversions from Python scalars to SlangPy arguments

venkataram-nv:
Addresses #119

**TLDR:** Allow passing Python `int`s as `float`s to slangpy (scalar parameters and fields of struct parameters) and vice versa.

From the SlangPy side this really only involves modifying the `IConvertibleFrom` specializations in `core.slang`. In addition to errors with passing a Python `int` to a SlangPy `float` field, there were likewise errors in passing a Python `float` to a SlangPy `int` field.

```slang
struct IntStruct {
        int x;
}

struct FloatStruct {
        float x;
}

float4 ok(float x)
{
        return x;
}

float4 error1(IntStruct s)
{
        return s.x;
}

float4 error2(FloatStruct s)
{
        return s.x;
}
```

```py
module = spy.Module.load_from_file(device, 'example.slang')

# Invocations
x = module.ok(x=512)
x = module.error1({ 'x': 512.0 })
x = module.error2({ 'x': 512 })
```

> [!IMPORTANT]  
> For Python `float` to SlangPy `int` field passing to work, there needs to be a slight correction in nanobind. Casting from int to float works already, but casting the other way (float to int) will result in a crash. I have a fix implemented in a fork on the current SlangPy nanobind fork, the diff can be viewed [here](https://github.com/skallweitNV/nanobind/compare/inheritance-fix...venkataram-nv:nanobind:casting?diff=split&w).

### DIFF
--- a/slangpy/tests/slangpy_tests/test_custom_types.py
+++ b/slangpy/tests/slangpy_tests/test_custom_types.py
@@ -453,7 +453,7 @@ int range_test(int input) {
 
 
 @pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
-@pytest.mark.parametrize("configuration", [(int, 'float'), (float, 'int')])
+@pytest.mark.parametrize("configuration", [(int, "float"), (float, "int")])
 def test_parameter_to_field_conversion(device_type: DeviceType, configuration: Tuple[type, str]):
 
     # Create function that just dumps input to output
@@ -473,7 +473,7 @@ int parameter_to_field_test(Uniform u) {{
 """,
     )
 
-    assert kernel_output_values({'x': ptype(512)}) == 512
+    assert kernel_output_values({"x": ptype(512)}) == 512
 
 if __name__ == "__main__":
     pytest.main([__file__, "-v"])
--- a/slangpy/tests/slangpy_tests/test_custom_types.py
+++ b/slangpy/tests/slangpy_tests/test_custom_types.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-from typing import Any, Callable
+from typing import Any, Tuple, Callable
 import numpy as np
 import numpy.typing as npt
 import pytest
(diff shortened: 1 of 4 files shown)


venkataram-nv:
FYI the currently failing test requires the changes I made to nanobind.

slangpy/tests/slangpy_tests/test_custom_types.py
```diff
@@ -448,10 +448,32 @@ def test_range(device_type: DeviceType):
     # Call function with 3D random arg
     res = kernel_output_values(range(10, 20, 2))
 
-    # Should get random numbers
```
jkwak-work, line 451: Let's keep this comment that seems to be still relevant.

slangpy/tests/slangpy_tests/test_vectorizing.py
```diff
@@ -66,7 +66,7 @@ def test_implicit_cast_no_vectorization(device_type: DeviceType):
     binding = call_data.debug_only_bindings.args[0]
     assert binding.vector_mapping.as_tuple() == ()
     assert binding.vector_type is not None
-    assert binding.vector_type.name == "float"
+    assert binding.vector_type.name == "int"
```
jkwak-work, line 69: Can you explain why this change is needed?
I am not sure how it is related to the rest of the changes

slangpy/tests/slangpy_tests/test_custom_types.py
venkataram-nv: For that test in particular we want the same values as the range, so its not random and comment is actually incorrect

slangpy/tests/slangpy_tests/test_vectorizing.py
venkataram-nv: The corresponding parameter in the slang function (`intfoo`, at the top of the file) has type `int`, and with these changes it is expected that the resulting binding type is the same (i.e. also `int`)

skallweitNV:
@venkataram-nv the nanobind PR needs to go upstream to the main nanobind repo at https://github.com/wjakob/nanobind

skallweitNV:
@venkataram-nv tests are failing on linux

PR #309, opened 2025-06-26
Enable Metal print support

### LINKED ISSUE #302: Implement print.slang for Metal
Background: https://github.com/shader-slang/slang/issues/6764

On Metal `RWByteAddressBuffer::GetDimensions` is not supported, because the buffer is mapped to a device pointer in Metal, it doesn't provide a way to query the size. Therefore, we will need another solution on Metal to pass the buffer size in.
This will involve adding the logic in
```
src/sgl/device/print.cpp 
src/sgl/device/print.h
src/sgl/device/print.slang
```

skallweitNV:
- Replace `buffer.GetDimensions` with a `buffer_capacity` uniform
- Add separate `buffer_size` buffer to hold atomic buffer size (needed for Metal)
- Use global shader variables instead of a ParameterBlock (improves compatibility with Metal running on virtual machines)
- Add missing buffer size clearing on initialization

Fixes #302

### DIFF
--- a/src/sgl/device/print.slang
+++ b/src/sgl/device/print.slang
@@ -373,7 +373,7 @@ struct DebugPrinter : IPrintOutput {
         uint size = (3 + arg_count + total_data_count) * sizeof(uint);
 
         // Reserve space for the message.
-        uint offset = buffer_size[0].add(size);
+        offset = buffer_size[0].add(size);
 
         if (offset + size > buffer_capacity - 4) {
             // Write sentinel value indicating that we have a buffer overlow.


src/sgl/device/print.slang
```diff
@@ -373,19 +368,23 @@ struct DebugPrinter : IPrintOutput {
...
-        buffer.GetDimensions(buffer_capacity);
-
         // Compute the size of the message.
         uint size = (3 + arg_count + total_data_count) * sizeof(uint);
 
         // Reserve space for the message.
+#ifdef __TARGET_METAL__
+        // Workaround for Slang issue https://github.com/shader-slang/slang/issues/7536
+        offset = buffer.Load(0);
+        buffer.Store(0, offset + size);
```
skallweitNV: This is currently needed as a workaround for the Metal atomics issue https://github.com/shader-slang/slang/issues/7536

Fixing this first is crucial because otherwise prints from multiple threads won't synchronize correctly.

src/sgl/device/print.slang
```diff
@@ -394,7 +393,12 @@ struct DebugPrinter : IPrintOutput {
         }
 
         buffer.Store(offset, size);
+#ifdef __TARGET_METAL__
+        // Workaround for Slang issue https://github.com/shader-slang/slang/issues/7534
+        buffer.Store(offset + 4, uint(getStringHash(fmt)));
```
skallweitNV, line 387: This is currently needed as a workaround for the Metal `RWStructureBuffer.Store` issue https://github.com/shader-slang/slang/issues/7534

src/sgl/device/print.slang
skallweitNV: Fixed by using separate `buffer_size` atomic buffer.

Issue #145, opened 2025-04-20
How to pass a matrix to slang via. torch and slangpy?


oliver-batchelor:
The only way I have made it work is using a `float[N][M]` as the slang type (see example below). It seems like it's meant to work, but I don't see an example - I tried messing with `MatrixType` where it seems to try to represent the matrix as an array of vectors, changing that to a matrix of scalars seemed to get further - but it didn't seem to play well with the slangpy prelude.

```
import math

from sgl import SlangCompilerOptions
import torch

from slangpy.backend import Device,  DeviceType
from importlib.resources import files
import slangpy as spy

slangpy_path = files('slangpy') / 'slang'

options = SlangCompilerOptions()
options.include_paths=[slangpy_path]

device = Device(
      type= DeviceType.automatic,
      compiler_options=options,
      enable_cuda_interop=True
)

source = """

[Differentiable]
float4x3 test(float4x3 a, float4x3 b) {
  return a * b;
}

[Differentiable]
float test2(matrix<float, 4, 3> a, vector<float, 3> c) {
  return 0.0;
}


[Differentiable]
float test3(float[4][3] a, vector<float, 3> c) {
  return 0.0;
}


"""

module = spy.TorchModule.load_from_source(device, "test", source)


a = torch.randn((16, 3, 4), dtype=torch.float32, device='cuda', requires_grad=True)
b = torch.randn((16, 3, 4), dtype=torch.float32, device='cuda', requires_grad=True)

# ValueError: Exception in kernel generation: Torch tensors do not support data type matrix<float,4,3>.
# result = module.test(a=a, b=b)


c = torch.randn((16, 3), dtype=torch.float32, device='cuda', requires_grad=True)

# ValueError: Exception in kernel generation: After implicit casting, no Slang overload found that matches the provided Python argument types. This typically suggests SlangPy selected an overload to call, but couldn't find a valid way to pass your Python arguments to it..
# result = module.test2(a=a, c=c)


# OK
result = module.test3(a=a, c=c)
```

bmillsNV:
@jarcherNV can you help to answer this?

jarcherNV:
Sorry it took my a while to get to this.

I'm not very familiar with torch so I could be wrong about this, but my current understanding is that this is currently only supported for arrays of scalars (in this case via the test3 function).

I thought perhaps you could use the floatMxN type in Slang and pass a numpy array to it (converted to numpy from torch) but that does not seem to work for me.

[...]

PR #314, opened 2025-06-26
Allow matrices to be passed to slangpy kernels

venkataram-nv:
Addresses #145

### DIFF
--- a/slangpy/slang/tensor.slang
+++ b/slangpy/slang/tensor.slang
@@ -114,6 +114,26 @@ public extension<T : IDifferentiable, let D : int, TensorType : ITensor<T, D>> T
             }
         }
     }
+
+    // TODO: row major/column major options
+    [Differentiable]
+    public void load<let M : int, let N : int>(ContextND<D - 2> context, out matrix <T, M, N> value)
+    {
+        int idx[D];
+        [ForceUnroll]
+        for (int i = 0; i < D - 2; ++i)
+            idx[i] = context.call_id[i];
+        [ForceUnroll]
+        for (int i = 0; i < M; i++) {
+            [ForceUnroll]
+            for (int j = 0; j < N; j++) {
+                idx[D - 2] = i;
+                idx[D - 1] = j;
+
+                value[i][j] = get(idx);
+            }
+        }
+    }
 }
 
 public extension<T : IDifferentiable, let D : int, TensorType : IRWTensor<T, D>> TensorType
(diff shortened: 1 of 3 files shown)


slangpy/slang/tensor.slang
```diff
@@ -171,6 +190,25 @@ public extension<T : IDifferentiable, let D : int, TensorType : IRWTensor<T, D>>
...
+        int idx[D];
+        [ForceUnroll]
+        for (int i = 0; i < D - 2; ++i)
+            idx[i] = context.call_id[i];
+        [ForceUnroll]
+        for (int i = 0; i < M; i++) {
+            [ForceUnroll]
+            for (int j = 0; j < N; j++) {
+                idx[D - 2] = i;
+                idx[D - 1] = j;
```
jkwak-work, line 205: Shouldn't this be
```
                idx[D - 1] = i;
                idx[D - 0] = j;
```
?

jkwak-work (approved):
Looks good to me.

slangpy/slang/tensor.slang
venkataram-nv: `idx[D - 0]` would be out of bounds, see line 196

kaizhangNV: No, idx[D-0] will be out of bound.

kaizhangNV:
we need more tests on this support, at least we will need a gradient tests.

slangpy/slang/tensor.slang
jkwak-work: Oh. Right. Dang.. what was I thinking... lol

venkataram-nv:
@jkwak-work @kaizhangNV ping for another check; all CI is passing now

slangpy/tests/slangpy_tests/test_tensor_with_grads.py
```diff
@@ -69,6 +69,37 @@ def test_differentiable_interface_parameters(device_type: DeviceType):
     compare_tensors(biases.grad_out.to_numpy(), bias_grad_ref)
 
 
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_differentiable_interface_parameters(device_type: DeviceType):
+    if device_type == DeviceType.cuda:
+        pytest.skip("CUDA backend generates invalid results")
```
jkwak-work, line 75: Is this expected that CUDA generates an invalid result? I wonder if it has something to do with the matrix row/column majorness memory layout.

jkwak-work (commented):
Looks good to me

slangpy/tests/slangpy_tests/test_tensor_with_grads.py
venkataram-nv: I'm not sure, that is from `test_differentiable_inference_parameters`

kaizhangNV: We know there are existing errors on cuda target, but not sure if this is the same issue.
Can you double check this will also fail on CUDA before skipping it?

venkataram-nv: Its not working on CI, but it keeps passing on my machine. Not sure if this is a CI problem or if I am missing something

kaizhangNV (approved):
LGTM

Issue #312, opened 2025-06-26
Build problem on Window when following the instruction


jkwak-work:
I followed the build instruction described in 
https://slangpy.shader-slang.org/en/latest/src/developer_guide/compiling.html#windows

And I got the following build errors.
```
D:\sbf\git\slang\slangpy>cmake --preset windows-msvc
Preset CMake variables:

  CMAKE_CXX_COMPILER="cl"
  CMAKE_C_COMPILER="cl"
  CMAKE_EXPORT_COMPILE_COMMANDS="ON"
  CMAKE_SYSTEM_VERSION="10.0.19041.0"
  CMAKE_TOOLCHAIN_FILE="D:/sbf/git/slang/slangpy/external/vcpkg/scripts/buildsystems/vcpkg.cmake"
  VCPKG_TARGET_TRIPLET="x64-windows-static-md-fix"

Preset environment variables:

  VCPKG_OVERLAY_TRIPLETS="D:/sbf/git/slang/slangpy/external/vcpkg-triplets"

-- Setting up mt-retry workaround.
-- Running vcpkg install
Detecting compiler hash for triplet x64-windows...
Compiler found: C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe
Detecting compiler hash for triplet x64-windows-static-md-fix...
Compiler found: C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe
All requested packages are currently installed.
Total install time: 545 us
-- Running vcpkg install - done
-- Platform: Windows
-- Architecture: x86_64
-- {fmt} version: 11.0.2
-- Build type:
-- Using Win32 for window creation
-- Populating slang ...
-- Populating slang_debug_info ...
-- Fetching DirectXShaderCompiler ...
-- Fetching WinPixEventRuntime ...
-- Fetching Agility SDK 1.616.0 ...
-- Fetching NVAPI ...
-- Fetching OptiX headers ...
-- Fetching vulkan headers ...
-- Copying shader file D:/sbf/git/slang/slangpy/build/windows-msvc/_deps/nvapi-src/nvHLSLExtns.h to nvapi
-- Copying shader file D:/sbf/git/slang/slangpy/build/windows-msvc/_deps/nvapi-src/nvHLSLExtnsInternal.h to nvapi
[...]

PR #315, opened 2025-06-26
Fix build problem when multiple VS toolsets are installed

jkwak-work:
When there are more than one visual studio toolset is installed, CMake may use an older version of the toolset while vcpkg always uses the latest one.

This PR adds a new CMake option to set which toolset version you want to use explicitly.
If it not explicitly set, it checks the environment variable, "VCToolsVersion", that gets set by the Visual Studio build environment command such as "x64 Native Tools Command Prompt for VS 2022".

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,19 @@ cd slangpy
 pip install .
 ```
 
+## Build Options
+
+When building from source, you can customize the build using the following CMake options (passed with `-D<OPTION>=<VALUE>`):
+
+| Option                          | Description                                                                 | Default |
+| ------------------------------- | --------------------------------------------------------------------------- | ------- |
+| `SGL_BUILD_PYTHON`              | Build the Python extension module.                                          | `ON`    |
+| `SGL_BUILD_EXAMPLES`            | Build the example applications.                                             | `ON`    |
+| `SGL_BUILD_TESTS`               | Build the test suite.                                                       | `ON`    |
+| `SGL_BUILD_DOC`                 | Build the documentation.                                                    | `OFF`   |
+| `SGL_USE_DYNAMIC_CUDA`          | Load the CUDA driver API dynamically at runtime.                            | `ON`    |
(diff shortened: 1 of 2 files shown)


jkwak-work:
Closes https://github.com/shader-slang/slangpy/issues/312

venkataram-nv (approved):
Looks good to me, might want another pair of eyes to look at this as well

README.md
```diff
@@ -41,6 +41,19 @@ cd slangpy
 pip install .
 ```
 
+## Build Options
```
skallweitNV: There is existing documentation for build options in the developer guide.
Can we keep that up-to-date instead of adding redundant documentation?

skallweitNV: documentation is here: https://github.com/shader-slang/slangpy/blob/main/docs/src/developer_guide/compiling.rst?plain=1#L171

skallweitNV:
@jkwak-work I had to revert this change as it prevented me from building slangpy locally.
Maybe a manual cmake configure within a native developer terminal works, but the configure process through VS Code and cmaketools doesn't work anymore, and I really want to preserve that workflow.

jkwak-work:
> @jkwak-work I had to revert this change as it prevented me from building slangpy locally. Maybe a manual cmake configure within a native developer terminal works, but the configure process through VS Code and cmaketools doesn't work anymore, and I really want to preserve that workflow.
> 
> Update: yes, when running cmake from the native tools command prompt it seems to work. So like:
> 
> ```
> mkdir build2
> cd build2
> cmake ../ --preset=windows-msvc
> ```
> 
> But again, we should make sure that people can just open the project in VS code and configure right from the IDE.

Sorry for the trouble I caused and thanks for reverting it before causing more problems on others.

But it seems strange that the existing flow would stop working with this PR.
This PR has no effect unless an environment variable is set.
And if it is set, it should give us more reliable result.

When you said "doesn't work", did you mean "VS Code" was having trouble parsing or using the file I modified?
Is this on Windows or MacOS?

PR #316, opened 2025-06-27
Make path tracer example work on CUDA

skallweitNV:
- update slang-rhi with a bunch of improvements for surface handling
- extend pathtracer example with ability to use raytracing pipelines (as CUDA/OptiX currently has no emulation for inline trace calls)
- extend blitter to optionally run with compute kernels


examples/pathtracer/pathtracer.cpp
```diff
@@ -30,6 +30,8 @@ SGL_EXPORT_AGILITY_SDK
 
 static const std::filesystem::path EXAMPLE_DIR(SGL_EXAMPLE_DIR);
 
+static const bool USE_RAYTRACING_PIPELINE = true;
```
tdavidovicNV, line 33: Can we have this as a parameter on the PathTracer?

examples/pathtracer/pathtracer.py
```diff
@@ -10,6 +10,8 @@
 
 EXAMPLE_DIR = Path(__file__).parent
 
+USE_RAYTRACING_PIPELINE = True
```
tdavidovicNV, line 13: Same question, would this make sense as in the .cpp, can we have this as not-global.

examples/pathtracer/pathtracer.slang
```diff
@@ -232,10 +216,114 @@ float3 envmap(float3 dir)
     // return env;
 }
 
+ParameterBlock<Scene> g_scene;
+RWTexture2D<float4> g_output;
+uniform uint g_frame;
+
+#if USE_RAYTRACING_PIPELINE
```
tdavidovicNV, line 223: This is mostly out of interest (could maybe use comment). If you're using the inline tracing (i.e., not PIPELINE), is there any cost in having the entry points in your code? As in, do you have to ifdef them out, or is that more or less optional?

examples/pathtracer/tone_mapper.slang
```diff
@@ -2,6 +2,7 @@
 
 float3 aces_film(float3 x)
 {
+    x *= 0.6;
```
tdavidovicNV, line 5: This looks weird, just checking it is not some debugging leftover.

skallweitNV: This is compensating for an exposure adjustment done in the fitted function.

examples/pathtracer/pathtracer.slang
skallweitNV: Not really, but it won't compile for CUDA then because the inline trace ray intrinsic is not available.

examples/pathtracer/pathtracer.cpp
skallweitNV: This is just example code, and I thought it's more obvious to have the thing named the same in both cpp/python code and the slang side.

PR #323, opened 2025-07-03
Initial pointer support

ccummingsNV:


slangpy/slang/core.slang
```diff
@@ -471,6 +471,9 @@ namespace impl
     // A python int can be passed to any slang integer scalar type
     extension<T : __BuiltinIntegerType> T : IConvertibleFrom<int> {}
 
+    // A python int (eg a buffer address) can be passed as a
```
skallweitNV:
```suggestion
    // A python int (e.g. a buffer address) can be passed to any slang pointer type
```

src/sgl/device/shader_cursor.cpp
```diff
@@ -479,6 +479,22 @@ void ShaderCursor::set_cuda_tensor_view(const cuda::TensorView& tensor_view) con
     }
 }
 
+void ShaderCursor::set_pointer(uint64_t pointer_value) const
+{
+    size_t pointer_size = m_type_layout->getSize();
+    if (pointer_size == 4) {
```
skallweitNV: is there HW that has 32-bit pointers?!

src/sgl/device/buffer_cursor.cpp
```diff
@@ -101,6 +101,22 @@ BufferElementCursor BufferElementCursor::find_element(uint32_t index) const
     return {};
 }
 
+void BufferElementCursor::set_pointer(uint64_t pointer_value)
+{
+    size_t pointer_size = m_type_layout->size();
+    if (pointer_size == 4) {
```
skallweitNV: same, do we really have 32-bit pointers anywhere?

slangpy/tests/slangpy_tests/test_pointers.py
```diff
@@ -0,0 +1,170 @@
...
+
+    res = function(100)
+
+    assert res == 100
+
+
+USAGES = [
+    BufferUsage.shader_resource,
+    BufferUsage.shader_resource | BufferUsage.unordered_access,
+    BufferUsage.shader_resource | BufferUsage.unordered_access | BufferUsage.shader_resource,
```
skallweitNV, line 48: this usage is the same as the one above .. you might want to test `unordered_access` as the only usage instead?!

src/sgl/device/shader_cursor.cpp
ccummingsNV: This was just me being safe for the sake of 4 lines of code :) - I can delete it if the answer's no

ccummingsNV: I'll change it to an asset

PR #328, opened 2025-07-03
Enable passing Python ints to Slang float fields

### LINKED ISSUE #119: Cannot pass a python int to a struct `float`.

venkataram-nv:
Fixes #119

The previous PR (#306) tried to additionally allow passing floats to int fields, but this behavior is discouraged and inconsistent with Python ([see this upstream PR, now closed](https://github.com/wjakob/nanobind/pull/1097#issuecomment-3033013135)). The changes in this PR are the reduced version of the older PR and require no upstream changes.

### DIFF
--- a/slangpy/slang/core.slang
+++ b/slangpy/slang/core.slang
@@ -485,8 +485,9 @@ namespace impl
     // If type B implements IConvertibleFrom<A>, then we can safely convert from A to B
     interface IConvertibleFrom<From> {}
 
-    // A python int can be passed to any slang integer scalar type
+    // A python int can be passed to any slang int/float scalar type
     extension<T : __BuiltinIntegerType> T : IConvertibleFrom<int> {}
+    extension<T : __BuiltinFloatingPointType> T : IConvertibleFrom<int> {}
 
     // A python int (e.g. a buffer address) can be passed to any slang pointer type
     extension<T> Ptr<T> : IConvertibleFrom<int> {}
--- a/slangpy/tests/slangpy_tests/test_custom_types.py
+++ b/slangpy/tests/slangpy_tests/test_custom_types.py
@@ -426,6 +426,26 @@ Particle rand_float_soa(Particle input) {
     assert np.all(pos >= -100.0) and np.all(pos <= 100.0)
     assert np.all(dir >= 0) and np.all(dir <= np.pi * 2)
 
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_parameter_to_field_conversion(device_type: DeviceType):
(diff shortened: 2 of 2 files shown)


slangpy/slang/core.slang
```diff
@@ -496,10 +497,18 @@ namespace impl
 
     // A python array can be passed to any slang array of the same length, as
     // long as their element types can be safely passed
-    extension<S, T : IConvertibleFrom<S>, let N : int, ArrT : ISizedArray<T, N>> ArrT : IConvertibleFrom<Array<S, N>> {}
+    extension<From : __BuiltinIntegerType, To : IConvertibleFrom<From>, let N : int, ArrTo : ISizedArray<To, N>>
```
ccummingsNV, line 500: Does this break any none-int/float array conversions?

slangpy/tests/slangpy_tests/test_custom_types.py
```diff
@@ -427,6 +427,28 @@ def test_rand_soa(device_type: DeviceType):
     assert np.all(dir >= 0) and np.all(dir <= np.pi * 2)
 
 
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_parameter_to_field_conversion(device_type: DeviceType):
```
ccummingsNV: Does the fix handle passing python ints to float arguments of functions? If so, we should have a test for that. If not, maybe it should - seems a bit inconsistent to support field conversions but not argument conversions.

These tests should probably go in their own tests file too - the custom types is to test things like the 'grid'

venkataram-nv: Passing ints to floats was already possible, but Ive added a test anyways. Both the additional and existing tests have been placed into `test_scalar_parameter_conversion.py`

slangpy/slang/core.slang
venkataram-nv: These changes will preserve conversions from int to float arrays, if that is what you mean. I have added a test for this (`test_array_conversion`)

kaizhangNV (approved):
lgtm

PR #336, opened 2025-07-09
Support using list for shape when creating NDBuffer

### LINKED ISSUE #107: Attempting to initialize a 'Shape' with a list throws poor error
Currently this code:

spy.NDBuffer(device, 'float3', shape=batch_shape)

Will raise an unhelful 'bad cast' error if batch_shape is a list. We should either support initialization of native Shape structures from lists, or implement a more helpful error message.

kaizhangNV:
Close #107

### DIFF
--- a/slangpy/tests/slangpy_tests/test_shapes.py
+++ b/slangpy/tests/slangpy_tests/test_shapes.py
@@ -16,20 +16,22 @@ from slangpy.types import floatRef
 from slangpy.types.buffer import NDBuffer
 from slangpy.types.valueref import ValueRef
 
+from typing import Union
+
 # First set of tests emulate the shape of the following slang function
 # float test(float3 a, float3 b) { return dot(a,b); }
 # Note that the return value is simply treated as a final 'out' parameter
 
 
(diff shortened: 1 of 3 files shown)


slangpy/tests/slangpy_tests/test_shapes.py
```diff
@@ -16,20 +16,22 @@
...
 
+from typing import Union
+
 # First set of tests emulate the shape of the following slang function
 # float test(float3 a, float3 b) { return dot(a,b); }
 # Note that the return value is simply treated as a final 'out' parameter
 
 
-def make_int_buffer(device_type: DeviceType, shape: tuple[int, ...]):
+def make_int_buffer(device_type: DeviceType, shape: Union[tuple[int, ...], list[int]]):
```
ccummingsNV: As you use this type hint so often, good practice would be to define it at the top (i.e. TTupleOrList), or potentially use a Sequence

ccummingsNV (approved):
Minor suggestion on neatness of type hints - your call about taking it.

kaizhangNV:
takes the suggestion about type hint, thanks.

PR #337, opened 2025-07-09
Change the _trampoline signature

### LINKED ISSUE #111: Generated trampoline function needs to use names with underscores, not 'data'
Currently the generated trampoline function uses fairly common names for arguments, including the word 'data'. If a user creates a function that also has a parameter called 'data', the names will collide and the kernel won't compile.

kaizhangNV:
Close #111

Change the _trampoline signature to exclude the CallData parameter, because there is already a ParameterBlock created for CallData.

In addition, since we choose a very common name for the parameter name of CallData - "data", this prevents user from writing function with the same parameter name.

### DIFF
--- a/slangpy/core/callsignature.py
+++ b/slangpy/core/callsignature.py
@@ -509,7 +509,7 @@ def generate_code(
     trampoline_fn = "_trampoline"
     if context.call_mode != CallMode.prim:
         cg.trampoline.append_line("[Differentiable]")
-    cg.trampoline.append_line(f"void {trampoline_fn}(Context context, CallData data)")
+    cg.trampoline.append_line(f"void {trampoline_fn}(Context context)")
     cg.trampoline.begin_block()
 
     # Declare parameters and load inputs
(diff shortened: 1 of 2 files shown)


skallweitNV:
Do I understand correctly that now `context` cannot be used as a parameter name by the user? Because that is a very common name as well. Maybe it should be `__context__` or even `__slangpy_context__` then?!

kaizhangNV:
no, context can still be used, but the `data` parameter in trampoline function is removed, now the trampoline function signature becomes
`void trampoline(Context context);`

previously is
`void trampoline(Context context, Call data);`

because `data` is a common name, and it's a global parameter, so it's no necessary to be a function parameter.

ccummingsNV (changes requested):
Change looks good, but I do agree with @skallweitNV - you may as well change 'context' to '__context__' or something whilst there. Otherwise we could hit the same issue - if the user writes a function that takes a variables called 'context', the trampoline function will have name collisions. You could of course make it global instead, but I think for this use case, local is neater.

PR #348, opened 2025-07-16
Added sanitization in case absolute module paths are used

tdavidovicNV:


ccummingsNV (approved):
Unless there's an issue with not escaping ':' in a regex, it looks good to me.

PR #350, opened 2025-07-16
Tweaks to doc index sidebar

### LINKED ISSUE #338: Clean up Reference section of slangpy documentation
https://slangpy.shader-slang.org/en/latest/src/api_reference.html

Currently generates Reference section, has API Reference and 3 sections for slangpy. Should be tidied up. (e.g., remove the 3 empty sections.)

### LINKED ISSUE #339: Update basic tutorials section
- Compute Shader section suggests this is easiest way, but it's not anymore. Reword to suggest it's another way to run compute shaders rather than the simplest/easiest way.
- Rename Tutorials -> Basic Tutorials to Graphics Tutorials. (remove subheading)
- Tidy any other bits you see

jhelferty-nv:
- Removes 3 empty Reference sections
- Renames Tutorials to Graphics Tutorials and removes subheading

Fixes #338
Fixes #339


docs/index.rst
jhelferty-nv, line 1: With these changes, only the Developer Guide remains in a folded subheading. Thoughts on whether that should also be unfolded? The main reason I left it like that is that on a 1080p display, the API Reference would then require scrolling down on the sidebar index to find, which seemed like bad UI. I could probably move the API Reference to the top, but that seemed a bit drastic? Thoughts?

kaizhangNV (approved):

Issue #187, opened 2025-05-05
Document external dependencies


skallweitNV:
We should document what external dependencies we rely on in slangpy. In particular this is:

- CUDA toolkit on Windows/Linux if using the CUDA backend (also, we might need a specific version)
- Metal SDK on macOS if using the Metal backend

Brought up in #186

jhelferty-nv:
Notes from scrum discussion w/ Yong:
- Test
  - CUDA versions:
    - version pytorch depends on
    - version in CI
    - current version
  - with:
    - all slang tests
    - all slangpy examples
- and document the version range based on the results.

jkwak-work:
Can you share an update for this?
You told me that you have done for CUDA 12.8 and 12.6 so far?
That leaves you:
- 12.8: Tested
- 12.7: Not tests
- 12.6: Tested
- 12.5: Not tests
- 12.4: Not tests
- 12.3: Not tests
- 12.2: Not tests
- 12.1: Not tests
- 12.0 ??

Is this right?

jkwak-work:
I don't know if it is necessary to test for all versions.
But I think it is expected work with 12.0 and even 11.

See a related PR before:
- https://github.com/shader-slang/slang/pull/7415/files

jhelferty-nv:
Confirmed cuda 11.8.0 works.

PR #354, opened 2025-07-16
Cleanups and doc updates

### LINKED ISSUE #226: SlangPy API reference contains some possibly stale SGL references
slangpy.math.transpose and slangpy.math.mul operations still reference sgl::math -- which shouldn't exist anymore I think?

jhelferty-nv:
Fixes #226

- Update repository name in developer guide
- Tweaks to Basics docs section
- Update firstfunctions samples documentation
- Fix some pydoc wrapper name resolution order issues
- Rename default window name to slangpy
- Document steps to regen py_doc.h, fix type
- Fix cmake rules for py_doc.h update for windows, update C++20 caveat comment
- Update py_doc.h and API Reference

### DIFF
--- a/docs/src/developer_guide/compiling.rst
+++ b/docs/src/developer_guide/compiling.rst
@@ -10,6 +10,11 @@ required:
 * Python ``>= 3.8``
 * git
 
+.. tip::
+
+    You may want to consider setting up and using a Python virtual environment
+    (e.g., ``venv``) to isolate your development activities.
+
 
 Cloning the repository
 ----------------------
@@ -176,7 +181,7 @@ can be specified on the command line when running CMake, for example:
 
 .. code-block:: bash
 
-    cmake --preset windows-msvc -DSGL_BUILD_DOCS=ON -DSGL_BUILD_EXAMPLES=OFF -DSGL_BUILD_TESTS=OFF
+    cmake --preset windows-msvc -DSGL_BUILD_DOC=ON -DSGL_BUILD_EXAMPLES=OFF -DSGL_BUILD_TESTS=OFF
 
 
 The following table lists the available configuration options:
@@ -198,7 +203,7 @@ The following table lists the available configuration options:
     * - ``SGL_BUILD_TESTS``
       - ``ON``
       - Build sgl tests
-    * - ``SGL_BUILD_DOCS``
+    * - ``SGL_BUILD_DOC``
       - ``OFF``
       - Build sgl documentation
     * - ``SGL_USE_DYNAMIC_CUDA``
@@ -216,6 +221,42 @@ The following table lists the available configuration options:
     * - ``SGL_ENABLE_HEADER_VALIDATION``
       - ``OFF``
       - Enable header validation
+
+
+
+Updating the API Reference
+--------------------------
+
+SlangPy uses ``pybind11_mkdoc`` to extract documentation strings from the C++
+source code. These comments are then used by ``nanobind`` to generate Python
+documentation comments. These comments are then used when building the API
+Reference document.
+
+To run ``pybind11_mkdoc``, specify the ``pydoc`` target when invoking cmake:
+
+.. code-block:: bash
+
+    # Install Python build prerequisites
+    pip install -r requirements-dev.txt
+
+    # Install Python documentation build prerequisites
+    pip install -r requirements-docs.txt
+
+    # Configure
+    cmake --preset windows-msvc
+
+    # Build with pydoc target
+    cmake --build --preset windows-msvc-release --target pydoc
+
+The generated API Reference page can then be updated by invoking the html build
+of the SlangPy docs. (It's regenerated as part of running ``sphinx-build``.)
+
+**Tested on:**
+
+* Windows 10 (build 19045)
+* Visual Studio 2022 (Version 17.13.6)
+* CMake 4.0.2
+* Ninja 1.12.1
 
 
 VS Code
--- a/docs/src/developer_guide/compiling.rst
+++ b/docs/src/developer_guide/compiling.rst
@@ -3,7 +3,7 @@
 Compiling
 =========
 
-In order to compile ``sgl`` from source, the following prerequisites are
+In order to compile ``slangpy`` from source, the following prerequisites are
 required:
 
 * A C++20 compliant compiler (tested with Visual Studio 2022, GCC 11 and Clang 14)
--- a/docs/src/developer_guide/compiling.rst
+++ b/docs/src/developer_guide/compiling.rst
@@ -3,7 +3,7 @@
 Compiling
 =========
 
-In order to compile ``slangpy`` from source, the following prerequisites are
+In order to compile SlangPy from source, the following prerequisites are
 required:
 
 * A C++20 compliant compiler (tested with Visual Studio 2022, GCC 11 and Clang 14)
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,0 +1,2 @@
+pybind11_mkdoc
+sphinx-build
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,2 +1,4 @@
 pybind11_mkdoc
-sphinx-build
+nbsphinx
+sphinx
+furo
--- a/src/sgl/app/app.h
+++ b/src/sgl/app/app.h
@@ -53,7 +53,7 @@ struct AppWindowDesc {
     /// Height of the window in pixels.
     uint32_t height{1280};
     /// Title of the window.
-    std::string title{"sgl"};
+    std::string title{"slangpy"};
     /// Window mode.
     WindowMode mode{WindowMode::normal};
     /// Whether the window is resizable.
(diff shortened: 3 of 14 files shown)


docs/src/developer_guide/compiling.rst
```diff
@@ -176,7 +181,7 @@ can be specified on the command line when running CMake, for example:
 
 .. code-block:: bash
 
-    cmake --preset windows-msvc -DSGL_BUILD_DOCS=ON -DSGL_BUILD_EXAMPLES=OFF -DSGL_BUILD_TESTS=OFF
+    cmake --preset windows-msvc -DSGL_BUILD_DOC=ON -DSGL_BUILD_EXAMPLES=OFF -DSGL_BUILD_TESTS=OFF
```
jhelferty-nv, line 184: In the cmake files, it's actually called `SGL_BUILD_DOC` instead of `SGL_BUILD_DOCS`. I wasn't sure if we had existing automation for the website that depended on this, so I updated the documentation instead of the cmake files.

docs/src/developer_guide/compiling.rst
```diff
@@ -218,6 +223,42 @@ The following table lists the available configuration options:
       - Enable header validation
 
 
+
+Updating the API Reference
+--------------------------
```
jhelferty-nv, line 228: In my experience on Windows, `-DSGL_BUILD_DOC=ON` isn't working, so I had to do a couple of manual steps. Arguably that should be addressed and we shouldn't need separate documentation, so I haven't bothered to make that section here.

Updating `src/slangpy_ext/py_doc.h`, which is arguably part of the base build and not strictly documentation, requires some additional packages be installed. For now I've documented the steps here, but long term, maybe this should be folded into the base build? The problem is that, up until now, there was a manual step to patch the python `clang` package for C++20. I didn't have this issue on my Windows machine with the latest version of `clang` and `pybind11_mkdoc`, but I'm not sure what's available on other platforms. Maybe I'll file a follow-up issue for this.

docs/src/developer_guide/compiling.rst
```diff
@@ -218,6 +223,42 @@ The following table lists the available configuration options:
...
+
+To run ``pybind11_mkdoc``, specify the ``pydoc`` target when invoking cmake:
+
+.. code-block:: bash
+
+    # Install Python build prerequisites
+    pip install -r requirements-dev.txt
+
+    # Install Python documentation build prerequisites
+    pip install -r requirements-docs.txt
```
jhelferty-nv, line 243: I've added these dependencies to a separate `requirements-docs.txt` file, because I'm not sure if we want all users to have to install `clang` and `sphinx`? Or what versions are needed? Again, maybe a topic for a follow-up task.

docs/src/developer_guide/compiling.rst
```diff
@@ -218,6 +223,42 @@ The following table lists the available configuration options:
...
+
+    # Configure
+    cmake --preset windows-msvc
+
+    # Build with pydoc target
+    cmake --build --preset windows-msvc-release --target pydoc
+
+The generated API Reference page can then be updated by invoking the html build
+of the SlangPy docs. (It's regenerated as part of running ``sphinx-build``.)
```
jhelferty-nv, line 252: So there's a step I was doing locally that isn't documented here, which is to run `pip install -e .` before running `sphinx-build` manually. (As mentioned earlier, the cmake docs build didn't work for me.) From looking at the scripts, I don't think it should be necessary, but noting here as a breadcrumb in case someone else runs into issues.

requirements-docs.txt
```diff
@@ -0,0 +1,4 @@
+pybind11_mkdoc
+nbsphinx
+sphinx
+furo
```
jhelferty-nv, line 4: sphinx is a dependency of `nbsphinx` and `furo` so doesn't need to be explicit, but.. I've included it in case the others get removed for some reason.

src/sgl/app/app.h
```diff
@@ -53,7 +53,7 @@ struct AppWindowDesc {
     /// Height of the window in pixels.
     uint32_t height{1280};
     /// Title of the window.
-    std::string title{"sgl"};
+    std::string title{"slangpy"};
```
jhelferty-nv, line 56: I have no idea if this is ever visible to the user. It does end up in the API Reference as the default, though, so I've changed it.

src/slangpy_ext/CMakeLists.txt
```diff
@@ -189,23 +189,35 @@ if(SGL_MASTER_PROJECT)
...
+        "${CMAKE_CURRENT_SOURCE_DIR}/../sgl/device"
+        "${CMAKE_CURRENT_SOURCE_DIR}/../sgl/math"
+        "${CMAKE_CURRENT_SOURCE_DIR}/../sgl/ui"
+        "${CMAKE_CURRENT_SOURCE_DIR}/../sgl/utils"
+    )
+    set(SGL_HEADER_FILES "")
+    foreach(sgl_dir IN ITEMS ${SGL_HEADER_DIRS})
+        file(GLOB FOUND_HEADERS "${sgl_dir}/*.h")
+        list(APPEND SGL_HEADER_FILES ${FOUND_HEADERS})
+    endforeach()
```
jhelferty-nv, line 204: This seems silly, but Windows doesn't do wildcards, so I had to get cmake to expand the wildcards..

src/slangpy_ext/device/reflection.cpp
jhelferty-nv, line 1: This fixes some instances of an issue only visible in `pydoc`. As mentioned in the commit comment, if you run python from the CLI and:

```
import slangpy
help(slangpy)
```
you'll see a bunch of `sgl::` showing up that don't show up when we gen the API Reference file. It's order of declaration related, unfortunately, and not all are fixable. These were the easy ones since they're in the same file.

src/slangpy_ext/slangpy_ext.cpp
jhelferty-nv, line 1: More fixes for CLI `pydoc` (by changing order things are wrapped).

docs/src/developer_guide/compiling.rst
```diff
@@ -3,13 +3,18 @@
 Compiling
 =========
 
-In order to compile ``sgl`` from source, the following prerequisites are
+In order to compile ``slangpy`` from source, the following prerequisites are
```
kaizhangNV: can we use SlangPy consistently through the document?

src/slangpy_ext/py_doc.h
```diff
@@ -3576,8 +3353,16 @@ static const char *__doc_sgl_FileSystemWatcher_m_output_delay_ms = R"doc(Delay b
 
 static const char *__doc_sgl_FileSystemWatcher_m_queued_events = R"doc(Events reported since last call to watch event callback.)doc";
 
+static const char *__doc_sgl_FileSystemWatcher_m_queued_events_mutex = R"doc(Mutex to protect the queued events.)doc";
+
+static const char *__doc_sgl_FileSystemWatcher_m_stop_thread = R"doc()doc";
```
kaizhangNV, line 3358: Is this intended to leave empty?

kaizhangNV (commented):
It looks good to me overall, but I'm not familiar with pydoc, and I don't know about the intention of few fix (e.g. why there is some empty R"doc()doc left  in py_doc.h)

So you will need Chris or Simon to take another look.

src/slangpy_ext/py_doc.h
jhelferty-nv: Are you asking about this line in particular, or more generally? These strings are extracted from the C++ source code by `pybind11_mkdoc`. If someone has added a documentation comment, it should show up here, otherwise the string is empty. The `D` macro defined in `src/slangpy_ext/nanobind.h` and used by the wrapper code in that subdir expands to use the comments from `py_doc.h` if an override string isn't otherwise specified.

docs/src/developer_guide/compiling.rst
jhelferty-nv: I'll take a look.

jhelferty-nv:
@ccummingsNV Mind taking a look at this one? Kai wasn't comfortable approving since he isn't as familiar with pydoc.

docs/src/developer_guide/compiling.rst
jhelferty-nv: Aside: I'm only updating the name here. Improving documentation of the dependencies is #187.

docs/src/developer_guide/compiling.rst
ccummingsNV: Good call - I'm not sure either so this is safer

docs/src/developer_guide/compiling.rst
ccummingsNV: I don't know if this impacts CI at all - the machine setups are automated, and probably involve pip insalling requirements-dev. They may need updating to install requirements-docs as well.

src/slangpy_ext/CMakeLists.txt
ccummingsNV: If it works, I trust you :)

jhelferty-nv:
Yeah, I don't think CI has any coverage, I've opened https://github.com/shader-slang/slangpy/issues/357 for follow-up discussion.

PR #361, opened 2025-07-21
Transition CUDA to use pointers for tensor + nd buffer data

ccummingsNV:


slangpy/slang/core.slang
```diff
@@ -312,7 +323,8 @@ public struct StridedLayout<let D : int>
 
 public struct NDBuffer<T, let N : int>
 {
-    public StructuredBuffer<T> buffer;
+    public StorageTraits<T>.BufferType buffer;
```
tdavidovicNV: I'd consider a `typealias` here

PR #362, opened 2025-07-21
Torch integration working with CUDA, Vulkan and D3D12

ccummingsNV:
This is the first version of rewritten torch integration, which includes fixed interop with vulkan/d3d, and direct device sharing with CUDA. Key additions:
- Lots of fixes for CUDA interop and CUDA backend
- Special 'torch module/function' removed - use of torch is now auto detected from parameters passed to a function
- Reduced complexity of autograd integrationn + moved some stuff native

What's missing and will be in next PR:
- Shifting more stuff to native land, such as the marshalling of TensorRef and aspects of the torch call
- A heap for interop buffers, to avoid constant re-allocation of buffers for passing in/out of torch

I'm still not entirely happy with the hoops we have to jump through for the auto-grad hook, but PyTorch has some pretty rock solid rules about what you can/can't do involving the storing/clearing/copying of tensors, and I can't find a simpler way to work around them.


slangpy/core/calldata.py
```diff
@@ -104,9 +111,31 @@ def __init__(
...
-            # Build 'unpacked' args (that handle IThis)
-            unpacked_args = tuple([unpack_arg(x) for x in args])
-            unpacked_kwargs = {k: unpack_arg(v) for k, v in kwargs.items()}
+            # Build 'unpacked' args (that handle IThis) and extract any pytorch
+            # tensor references at the same time.
+            tensor_refs = []
+            unpacked_args = unpack_refs_and_args(tensor_refs, *args)
+            unpacked_kwargs = unpack_refs_and_kwargs(tensor_refs, **kwargs)
+
+            # If we have tensor references, store whether we a call to this function
```
tdavidovicNV: typo: "whether we a call"

.github/workflows/ci.yml
```diff
@@ -91,8 +91,16 @@ jobs:
         run: |
           sudo apt update && sudo apt install -y libxinerama-dev libxcursor-dev xorg-dev libglu1-mesa-dev pkg-config
 
-      # Setup Python.
-      - name: Setup Python ${{ matrix.python }}
+      # Setup Python (no pip cache on unit test windows runners - massive slow down).
```
skallweitNV, line 94: Why can't we use the cache?

src/slangpy_ext/device/kernel.cpp
```diff
@@ -50,11 +52,21 @@ SGL_PY_EXPORT(device_kernel)
                     // bind globals
                     bind_python_var(cursor, vars);
                 };
-                self->dispatch(thread_count, bind_vars, command_encoder);
+                if (command_encoder) {
+                    SGL_CHECK(
+                        !cuda_stream.is_valid(),
+                        "Can not specify cuda stream if appending to a command encoder."
```
skallweitNV: nitpick: in general I think we should use `CUDA` all-caps in text/strings/output

src/slangpy_ext/device/kernel.cpp
```diff
@@ -40,6 +40,8 @@ SGL_PY_EXPORT(device_kernel)
                uint3 thread_count,
                nb::dict vars,
                CommandEncoder* command_encoder,
+               CommandQueueType queue,
```
skallweitNV, line 43: we don't support multiple queues, so is this needed?

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -36,6 +41,23 @@ void SignatureBuilder::add(const char* value)
 {
     add_bytes((const uint8_t*)value, (int)strlen(value));
 }
+void SignatureBuilder::add(const uint32_t value)
+{
+    uint8_t buffer[8];
+    for (int i = 0; i < 8; ++i) {
+        buffer[7 - i] = HEX_CHARS[(value >> (i * 4)) & 0xF];
```
skallweitNV, line 48: minor: there is `sgl::string::hexlify` but it returns a `std::string` so this has no allocation overhead. but we could also introduce a hexlify overload that takes an output buffer and use that.

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -1197,13 +1295,37 @@ SGL_PY_EXPORT(utils_slangpy)
...
+            &NativeCallData::_py_torch_call,
+            nb::arg("function"),
+            nb::arg("opts"),
+            nb::arg("args"),
+            nb::arg("kwargs"),
+            D_NA(NativeCallData, _py_torch_call)
+        )
         .def_prop_rw(
             "call_group_shape",
             &NativeCallData::get_call_group_shape,
```
skallweitNV: unrelated but this should be just `call_group_shape`

src/sgl/device/cuda_interop.h
```diff
@@ -9,6 +9,7 @@
 namespace sgl::cuda {
 
 struct TensorView {
+    int device_id;
```
skallweitNV, line 12: seems to not be used anywhere?!

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
         std::filesystem::create_directories(m_shader_cache_path);
     }
 
+    // Invalidate cuda interop if using cuda
```
skallweitNV:
```suggestion
    // Invalidate CUDA interop if using cuda
```

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+        log_warn("Device type is set to CUDA, but CUDA interop is requested. enable_cuda_interop will be ignored.");
+    }
+
+    // Invalidate use of adapter LUID if existing handles provided.
+    if (m_desc.adapter_luid.has_value()) {
+        for (const auto& handle : m_desc.existing_device_handles) {
+            if (handle.is_valid()) {
+                m_desc.adapter_luid.reset();
+                log_warn("Both adapter luid and existing handles have been provided, which are both ways to "
+                         "specify the device. Adapter luid will be ignored in favor of provided existing handles");
```
skallweitNV:
```suggestion
                log_warn("Both adapter LUID and existing handles have been provided, which are both ways to "
                         "specify the device. Adapter LUID will be ignored in favor of provided existing handles");
```

src/sgl/device/cuda_utils.h
```diff
@@ -44,23 +45,28 @@ SGL_API void destroy_external_semaphore(CUexternalSemaphore ext_sem);
...
     explicit Device(const sgl::Device* device);
+    explicit Device(CUcontext context);
+    explicit Device(CUdevice device);
     ~Device();
 
     CUdevice device() const { return m_device; }
     CUcontext context() const { return m_context; }
-    CUstream stream() const { return m_stream; }
+
+    AdapterLUID get_adapter_luid();
```
skallweitNV: should just be `adapter_luid()` and `adapter_name()` and probably should be `const`

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+            close();
+            SGL_THROW("Failed to initialize CUDA driver API.");
+        }
+
+        CUdevice cuda_device = -1;
+        CUcontext cuda_context = nullptr;
+        for (auto& handle : m_desc.existing_device_handles) {
+            if (handle.type() == NativeHandleType::CUdevice) {
+                cuda_device = handle.as<CUdevice>();
+                handle = {};
```
skallweitNV, line 125: why are we clearing these handles?

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+            } else if (handle.type() == NativeHandleType::CUcontext) {
+                cuda_context = handle.as<CUcontext>();
+                handle = {};
+            }
+        }
+        if (cuda_context) {
```
skallweitNV, line 131: shouldn't the device take precedence over the context here?

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+                "DeviceDesc. "
+                "The current primary CUDA context handles can be acquired with "
+                "slangpy.get_cuda_current_context_native_handles."
+            );
+        }
+    }
+
+    // If we now have a valid cuda device, use it to determine the adapter LUID.
+    if (m_cuda_device) {
+
```
skallweitNV: extra newline

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+                "CUDA interop is enabled, but no existing CUDA device or context is provided. To ensure the correct "
+                "device is selected, pass the CUDA device or context in as an existing device handle in the "
+                "DeviceDesc. "
+                "The current primary CUDA context handles can be acquired with "
+                "slangpy.get_cuda_current_context_native_handles."
+            );
+        }
+    }
+
+    // If we now have a valid cuda device, use it to determine the adapter LUID.
```
skallweitNV:
```suggestion
    // If we now have a valid CUDA device, use it to determine the adapter LUID.
```

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,97 @@ Device::Device(const DeviceDesc& desc)
...
+                    );
+                    found = true;
+                    break;
+                }
+            }
+        }
+
+        if (!found) {
+            close();
+            SGL_THROW("Unable to find matching adapter LUID or name for the provided CUDA device. ");
```
skallweitNV:
```suggestion
            SGL_THROW("Unable to find matching adapter LUID or name for the provided CUDA device.");
```

src/sgl/device/device.cpp
```diff
@@ -201,27 +292,37 @@ Device::Device(const DeviceDesc& desc)
...
     m_global_fence = create_fence({.shared = m_desc.enable_cuda_interop});
 
-    // Setup CUDA interop.
+    // Finalize CUDA interop.
     if (m_desc.enable_cuda_interop) {
-        SGL_CHECK(rhiCudaDriverApiInit(), "Failed to initialize CUDA driver API.");
-        m_cuda_device = make_ref<cuda::Device>(this);
-        m_cuda_semaphore = make_ref<cuda::ExternalSemaphore>(m_global_fence);
+
+        // If didn't create cuda device from existing handles, make it now that we
```
skallweitNV:
```suggestion
        // If didn't create CUDA device from existing handles, make it now that we
```

.gitignore
```diff
@@ -26,3 +26,5 @@ slangpy.egg-info/
 slangpy/tests/slangpy_tests/logs/
 
 *.nsys-rep
+
+*.sqlite
```
skallweitNV, line 30: where do we use sqlite?

src/sgl/device/cuda_utils.cpp
```diff
@@ -231,20 +232,79 @@ Device::Device(const sgl::Device* device)
...
+        SGL_CU_CHECK(cuDevicePrimaryCtxRelease(m_device));
+    }
+}
+
+AdapterLUID Device::get_adapter_luid()
+{
+    AdapterLUID luid;
+    memset(luid.data(), 0, luid.size());
+
+#if SGL_WINDOWS
```
skallweitNV, line 284: the LUID query code could be shared with what's in `find_device_by_luid`

src/slangpy_ext/utils/slangpy.h
```diff
@@ -758,9 +801,56 @@ class PyNativeCallDataCache : public NativeCallDataCache {
...
+    void set_id(int32_t id) { m_id = id; }
+
+    ref<TensorRef> grad_in() const { return m_grad_in; }
+    void set_grad_in(const ref<TensorRef>& grad_in) { m_grad_in = grad_in; }
+
+    ref<TensorRef> grad_out() const { return m_grad_out; }
+    void set_grad_out(const ref<TensorRef>& grad_out) { m_grad_out = grad_out; }
+
+    std::pair<AccessType, AccessType> last_access() const { return m_last_access; }
+    void set_last_access(const std::pair<AccessType, AccessType>& last_access) { m_last_access = last_access; }
```
skallweitNV, line 840: it looks a lot like these could all just be public fields instead of getter/setters

slangpy/core/function.py
```diff
@@ -132,6 +132,15 @@ def set(self, *args: Any, **kwargs: Any):
                 "Set requires either keyword arguments or 1 dictionary / hook argument"
             )
 
+    def cuda_stream(self, stream: NativeHandle) -> "FunctionNode":
+        """
+        Specify a CUDA stream to use for the function. This is useful for synchronizing with other
+        CUDA operations or ensuring that the function runs on a specific stream.
+        """
+        if stream.type != NativeHandleType.CUstream:
```
skallweitNV: you also do that check in `FunctionNodeCUDAStream` constructor

slangpy/tests/slangpy_tests/helpers.py
```diff
@@ -82,6 +80,17 @@ def get_device(
             "Please set use_cache=False if you want to use existing_device_handles."
         )
 
+    selected_adaptor_luid = None
```
skallweitNV, line 83: slightly different then in `sglhelpers.py`, maybe worth consolidating. i.e. we could maybe select the adapter in conftest.py?

slangpy/tests/slangpy_tests/helpers.py
```diff
@@ -121,10 +131,39 @@ def get_device(
...
+# This is useful for testing the torch integration.
+def get_torch_device(type: DeviceType) -> Device:
+    import torch
 
+    # For testing, comment this in to disable backwards passes running on other threads
+    torch.autograd.grad_mode.set_multithreading_enabled(False)
 
+    torch.cuda.init()
+    torch.cuda.current_device()
+    torch.cuda.current_stream()
```
skallweitNV, line 147: are these calls necessary?

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+
+from typing import Any, Optional, cast
+from numpy import ScalarType
+from slangpy import DataType, Device, BufferUsage, TypeReflection, DeviceType
+import torch
```
skallweitNV, line 7: are we guaranteed that this import is not run at `import slangpy` time?

skallweitNV (dismissed):
LGTM, some of the comments are on the nitpicky side, so feel free to ignore. I didn't super deep into the marshalling and pytorch autograd implementation but reading through it looked good to me.

.github/workflows/ci.yml
ccummingsNV: The cost of unzipping a zillion files is about 100x slower than the cost of realizing everything is already installed from the last run :) Without this, python process takes about 10mins total on windows, vs about 2s with it.

slangpy/tests/slangpy_tests/helpers.py
ccummingsNV: yeah unfortunately :) torch doesn't seem to reliably set a cuda context, even if you've initialized cuda, until you do that.

slangpy/torchintegration/torchtensormarshall.py
ccummingsNV: I'll double check, but torchtensormarshall.py shouldn't be imported by anything globally.

src/slangpy_ext/device/kernel.cpp
ccummingsNV: I'd argue we need to get multiple queues in, but more importantly, if we support it for submit, then we should support it for computekernel.dispatch, which is just a fancy way of calling submit.

src/sgl/device/device.cpp
ccummingsNV: If we don't they're passed through to the device init, which explodes because we're, eg, passing existing cuda device handles to a d3d12 rhi layer.

src/sgl/device/device.cpp
ccummingsNV: I concluded not, because the context is 'more precise' than the device. Our 'get current cuda context' returns existing handles that include both the device and context. If the user has supplied a context, we should use it - not create a new one. We could raise an error if they've requested a context that doesn't match the device though.

.gitignore
ccummingsNV: nsight systems spits them out in the current working directory. this isn't critical - I just didn't want to accidently check them in.

src/sgl/device/cuda_utils.cpp
ccummingsNV: I'll take a peak and see how easy it is to merge them in the next PR.

slangpy/tests/slangpy_tests/helpers.py
ccummingsNV: yeah I think in the next PR I'm going to try merging these 2 together a bit - it's silly how separate they are.

src/sgl/device/cuda_interop.h
ccummingsNV: Going to use this (probably) in next PR. Will remove if I don't.

src/slangpy_ext/utils/slangpy.h
ccummingsNV: TensorRef's going to have some serious work in next PR, so I'll review then

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: I'll check for next PR - I might end up deleting these anyway

src/sgl/device/device.cpp
```diff
@@ -91,6 +91,96 @@ Device::Device(const DeviceDesc& desc)
         std::filesystem::create_directories(m_shader_cache_path);
     }
 
+    // Invalidate CUDA interop if using CUDA
+    if (m_desc.type == DeviceType::cuda && m_desc.enable_cuda_interop) {
+        m_desc.enable_cuda_interop = false;
+        log_warn("Device type is set to CUDA, but CUDA interop is requested. enable_cuda_interop will be ignored.");
```
tunabrain, line 97: Should this be a warning? Nothing unexpected would happen if you request CUDA interop on a CUDA device that the user should be made aware of, right? It's just that the flag is not needed.

src/sgl/device/kernel.cpp
```diff
@@ -55,10 +49,20 @@ void ComputeKernel::dispatch(uint3 thread_count, BindVarsCallback bind_vars, Com
         pass_encoder->dispatch(thread_count);
         pass_encoder->end();
     }
```
tunabrain, line 51: Do you still need the nested braced block?

slangpy/torchintegration/autogradhook.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Clear all torch tensor references (PyTorch takes over at this point, and may
+        # want to allocate new ones, so holding on to them can just cause excess memory usage)
+        clear_tensor_refs(tensor_refs)
+
+        # Return the outputs, so they get hooked into the torch auto-grad graph
+        return tuple(primal_out_tensors)
+
+    @staticmethod
+    def backward(ctx: Any, *args: torch.Tensor):
```
tunabrain, line 129: We don't support double differentiation, and should decorate backward with `once_differentiable` to trigger the appropriate torch checks/errors.
https://docs.pytorch.org/docs/stable/generated/torch.autograd.function.once_differentiable.html#torch.autograd.function.once_differentiable

slangpy/torchintegration/autogradhook.py
```diff
@@ -0,0 +1,177 @@
...
+        if arg.id >= 0:
+            arg.tensor = all_tensors[arg.id]
+            if arg.last_access[0] in (AccessType.read, AccessType.readwrite):
+                arg.grad_out = TensorRef(-1, torch.zeros_like(arg.tensor))
+                grad_out_tensors.append(arg.grad_out.tensor)  # type: ignore
+            if arg.last_access[0] in (AccessType.write, AccessType.readwrite):
+                arg.grad_in = TensorRef(-1, grad_in_tensors.pop(0).contiguous())
+
+
+def alloc_gradients(args: list[TensorRef], tensors: list[Optional[torch.Tensor]]) -> Any:
```
tunabrain, line 73: Is this still used anywhere? If not, we can delete it.

slangpy/core/utils.py
```diff
@@ -71,6 +71,51 @@ def create_device(
     return device
 
 
+def create_torch_device(
```
tunabrain, line 74: Is this user-callable? If so, can we import it in `__init__.py`?

slangpy/core/utils.py
```diff
@@ -71,6 +71,51 @@ def create_device(
...
+
+    torch.cuda.init()
+
+    # Use current device if not specified
+    if torch_device is None:
+        torch_device = torch.cuda.current_device()
+
+    # Switch to the correct device then read cuda context
+    with torch.device(torch_device):
+        handles = get_cuda_current_context_native_handles()
```
tunabrain, line 105: It looks like this will throw (`No current CUDA context found.`) if the torch device has not been used yet. Minimal repro:
```
torch_device = torch.device('cuda:0')
torch.set_default_device(torch_device)
device = create_torch_device(spy.DeviceType.d3d12, torch_device)
```

It appears torch does not set the cuda context until you do something with it. In the code above, you can fix it by e.g. adding `torch.empty((1, ), device=torch_device)` before the `create_torch_device` call.

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
...
+        return result
+
+    def read_calldata(
+        self,
+        context: CallContext,
+        binding: "BoundVariableRuntime",
+        data: TensorRef,
+        result: Any,
+    ):
+        if context.device.info.type != DeviceType.cuda:
```
tunabrain, line 193: This should also check whether `binding.access[0]` is `write` or `readwrite`. Currently it will also copy back all read-only arguments.

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
...
+
+            primal.untyped_storage().copy_(interop_tensor.untyped_storage())
+            data.interop_buffer = None
+
+            if self.d_in is not None:
+                assert data.grad_in is not None
+                self.d_in.read_calldata(context, binding, data.grad_in, result["d_in"])
+            if self.d_out is not None:
+                assert data.grad_out is not None
+                self.d_out.read_calldata(context, binding, data.grad_out, result["d_out"])
```
tunabrain, line 214: For safety, it should probably also check the call mode and derivative access to see if `grad_out` needs to be copied. Looking at the code I'm not sure if there is a code path where `grad_out` can be set without being in `bwd` call mode, but better to be safe than overwrite a derivative when you shouldn't.

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
...
+                torch.Tensor,
+                data.interop_buffer.to_torch(type=data_type, shape=shape, strides=strides),
+            )
+
+            primal.untyped_storage().copy_(interop_tensor.untyped_storage())
+            data.interop_buffer = None
+
+            if self.d_in is not None:
+                assert data.grad_in is not None
+                self.d_in.read_calldata(context, binding, data.grad_in, result["d_in"])
```
tunabrain, line 211: Isn't `grad_in` read-only? I don't think we need to copy it back.

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
...
+
+            shape = tuple(primal.shape)
+            strides = primal.stride()
+            data_type = _torch_to_data_type[self.torch_dtype]
+            interop_tensor = cast(
+                torch.Tensor,
+                data.interop_buffer.to_torch(type=data_type, shape=shape, strides=strides),
+            )
+
+            primal.untyped_storage().copy_(interop_tensor.untyped_storage())
```
tunabrain, line 206: This will fail if the size of the storage != size of the tensor. In `create_calldata`, you create an interop buffer of size `primal.numel() * primal.element_size()`, but the size of `primal`'s storage can be different.

For example:
```
a = torch.empty((64, 64))
b = a[0]
print(b.numel() * b.element_size()) # 256
print(b.untyped_storage().nbytes()) # 16384
```

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -1 +1,272 @@
...
+        if context.device.info.type != DeviceType.cuda:
+
+            data_type = _torch_to_data_type[self.torch_dtype]
+            data.interop_buffer = get_storage(context, primal.numel(), primal.element_size())
+
+            interop_tensor = cast(
+                torch.Tensor,
+                data.interop_buffer.to_torch(type=data_type, shape=shape, strides=strides),
+            )
+            interop_tensor.copy_(primal)
```
tunabrain, line 134: This may copy wrong data or read out of bounds if `primal` has non-standard strides. The interop buffer is created assuming a dense layout (based on `numel` and `element_size`), but the `to_torch` call passes the `strides` of the original tensor, which might not be dense.

For example:
```
a = torch.randn((64, 64), dtype=torch.float)
b = a[::2]
```
Passing `b` to slangpy causes a `CUDA_ERROR_ILLEGAL_ADDRESS`

slangpy/tests/slangpy_tests/helpers.py
```diff
@@ -121,10 +131,39 @@ def get_device(
...
+TORCH_DEVICES: dict[str, Device] = {}
+
+
+# Helper that gets a device that wraps the current torch cuda context.
+# This is useful for testing the torch integration.
+def get_torch_device(type: DeviceType) -> Device:
+    import torch
 
+    # For testing, comment this in to disable backwards passes running on other threads
+    torch.autograd.grad_mode.set_multithreading_enabled(False)
```
tunabrain, line 143: Does torch integration work if `autograd` uses multithreading? If not, it should probably be disabled in `slangpy.core.utils.create_torch_device` as well.

slangpy/core/utils.py
tunabrain: After taking a quick look I haven't seen an obvious method in `torch.cuda` for setting the CUDA context. Weirdly,  `torch.cuda.init()` does not seem to do it.

Unfortunately the common practice of creating all devices up front before doing any work will trigger this issue. For safety, you could add a minimal torch cuda call like e.g. creating an empty 1x1 cuda tensor inside `create_torch_device`.

tunabrain:
Left a few comments for follow-up. I haven't dug deeply into the device/cuda changes, but the rest looks mostly good.

Main comments for a follow-up PR:
- There's redundant read-backs happening for read-only tensors that should be avoided
- The code becomes unstable (exceptions or CUDA memory access violations) if torch tensors are not contiguous or have non-zero offset. For now, I would add a check to disallow torch tensors with non-standard strides or non-zero offset, until we have a proper fix.
- What happens when the same torch tensor is passed to multiple arguments of the function? (e.g. `foo(a, a)`). There would be multiple `TensorRef`s with potentially different values of `last_access` referencing the same tensor. Is that a problem? I'm not sure how this interacts with autograd. If we're not confident this is fine, we should probably also check for this and disallow it.

PR #363, opened 2025-07-21
Add copy_from_torch() function

### LINKED ISSUE #113: Easy path to convert buffers of basic vector/matrix types to torch tensors
This is a common requirement, and currently forcing people into a path of always using buffers of floats, because they need to go to a torch tensor. Ideally there'd be code that could at least understand how to go from, eg, a buffer of float3s to an [N,3] tensor.

aidanfnv:
Closes #113

This change implements a function `copy_from_torch()`, for copying data from torch tensors to `NDBuffer`s and `Tensor`s, and also adds tests to verify that this `copy_from_torch()` functions as expected.


slangpy/tests/slangpy_tests/test_buffer_views.py
```diff
@@ -1,14 +1,23 @@
...
+from slangpy import DeviceType, BufferUsage
 from . import helpers
 from slangpy.types import NDBuffer, Tensor
 
 from typing import Any, Union, Type
 
 import numpy as np
+import sys
 
+try:
```
ccummingsNV: Many of the tests in this file don't require torch. You can either move the torch specific tests to, eg, test_dtypes which is torch only, or do the torch import + pytest.skip on a per test basis (prefable). Lots of tests here that are well worth Metal not breaking though :)

ccummingsNV (changes requested):
Tests look good but you can't skip all the tests in the file on Metal

slangpy/tests/slangpy_tests/test_buffer_views.py
aidanfnv: I have moved the skips into the tests to skip on a per test basis in https://github.com/shader-slang/slangpy/pull/363/commits/cefbf7267bb901339537d8683da9231cfd0e33ce

aidanfnv:
One of my previous commits caused the CI to hang during one of the tests, and I was not able to figure out exactly what the issue is, but I did rewrite my copy_from_torch() implementation to directly copy from torch to Slang buffer (with the torch-to-numpy-to-buffer implementation being used as a fallback), and now it seems to work.

ccummingsNV (approved):
Looks good to me.

Note: the file descriptor errors in tests are not related (and are fixed in recent PR). I've merged main back in and will re-run them.

ccummingsNV (approved):

PR #367, opened 2025-07-22
Moved debug_printer string registration to include strings in imported modules.

tdavidovicNV:
Moved the `debug_printer` string registration to `ShaderProgram::store_built_data`. In the previous location, it obtained only strings from modules directly loaded using `sgl`, not from modules that were imported using the Slang's `import` keyword.

Added a `test_print_module2.slang` so `test_print.slang` has a module it can `import` and we test that the strings are loaded correctly from it as well.

### DIFF
--- a/slangpy/tests/device/test_print.py
+++ b/slangpy/tests/device/test_print.py
@@ -22,6 +22,7 @@ def test_print(device_type: spy.DeviceType):
     # print("result:", result)
     expected = {
         spy.DeviceType.d3d12: """Hello World!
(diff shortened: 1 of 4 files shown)


tdavidovicNV:
@ccummingsNV the race condition and torch tests on Windows are failing on this. I don't think it could have possibly been triggered by my code, tho.

ccummingsNV:
Agreed - it's because pytorch is now on the machines. The fix is to disable the failing test - it's just a race condition that doesn't repro reliably

PR #368, opened 2025-07-22
Enable Claude Code Agent Action

szihs:


szihs:
Strange that the CI is broken. The changes should not have any impact on CI. @ccummingsNV is there some intermittent issue at play here?

tdavidovicNV:
@szihs the CI has an intermittent (and unfortunately rather frequent) errors in that test that @ccummingsNV is investigating.

szihs:
> @szihs what is this based off? Should we copy/paste the workflow across multiple repos like this or share it using a github action instead? And what is that workflow doing? Some description or references would be good.

@skallweitNV  - Most of the heavy work is already done by [using](https://github.com/marketplace/actions/claude-code-action-official) claude-code-action. You should be able to assign issues to @claude - for example 
https://github.com/szihs/slangpy/issues/2#issuecomment-3097565643

CLAUDE.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Build Commands
+```bash
+# Install from source (recommended)
+pip install .
+
+# Configure and build with CMake
+python tools/ci.py configure
+python tools/ci.py build
+```
```
jkwak-work, line 19: The document also mentions the following instruction.
https://slangpy.shader-slang.org/en/latest/
```
pip install -r requirements-dev.txt
```
Don't we need it here?

CLAUDE.md
```diff
@@ -0,0 +1,89 @@
...
+
+# Run example tests
+pytest samples/tests -vra
+
+# Run C++ unit tests
+python tools/ci.py unit-test-cpp
+
+# Run a single test
+pytest slangpy/tests/test_buffer.py::test_buffer_creation -v
+```
```
jkwak-work, line 36: The testing commands look a little different from how CI does.
https://github.com/shader-slang/slangpy/blob/main/.github/workflows/ci.yml

Are those instructions correct and functional?

jkwak-work (commented):
I think we need some instruction of how to clone slang repo as well.
When the issue is coming from slang rather than slangpy, LLM will have to deal with slang as well.

Or we can just clone slang as a part of the claude.yml??

CLAUDE.md
szihs: Added the proper build instructions.

szihs:
> @szihs what is this based off? Should we copy/paste the workflow across multiple repos like this or share it using a github action instead? And what is that workflow doing? Some description or references would be good.


Updated the PR - https://github.com/shader-slang/slang/pull/7811 should be utilized for common action

CLAUDE.md
szihs: This is standard `pytest` command.

jkwak-work (approved):
Looks good to me

skallweitNV:
@szihs updating branch is probably needed because we disabled the merge queue feature?! But I agree, if the change is fast-forward, we should just be able to merge without updating.

Issue #357, opened 2025-07-18
Docs-related CI coverage questions


jhelferty-nv:
#354 added a `requirements-docs.txt` with a few additional dependencies required to complete all steps of the documentation build. It's not clear to me how much of this should be covered by CI, or whether a comment propagation step (when editing C++ code) that's currently manual should become normative, so creating an issue here mainly as a place to centralize the discussion.

Aside: The list of dependencies in `requirements-docs.txt` is slightly different from the `requirements.txt` found in https://github.com/shader-slang/shader-slang.github.io

The SlangPy documentation build depends on:
- `pybind11_mkdoc` to generate the `py_doc.h` file that contains all the comments from the C++ `src/sgl` code. This code is then used by the `nanobind` bindings in `src/slangpy_ext` as python documentation comments. This step needs to be rerun whenever a change is made to the C++ code so that any comments get propagated to python.
  - Notably, this process currently needs to be manually invoked by specifying the `pydoc` cmake target. It looks like this choice was made because C++20 wasn't supported by earlier versions of the `pybind11_mkdoc` and `clang` python packages, so `clang` required a manual patch step. From my own experience on Windows, this manual step is no longer necessary.
- `sphinx` to generate the html. Our sphinx config file, `docs/conf.py`, invokes `docs/generate_api.py`, which extracts the python documentation and propagates it into `docs/generated/api.rst` so that `sphinx` will generate an html version of it for the website.
  - Note that this step (at least locally for me) generates a lot of errors. I'm not sure if this is a local python packaging versioning issue, or if the pages are just generally 'not clean'.
- `furo` sphinx theme
- `nbsphinx` for the Jupityr notebooks (used by the tutorial pages)

Summarizing, the points to consider:
- Should the cmake `pydoc` target to update binding comments become part of the normal build?
- Should the docs build have CI coverage?
- Should the `requirements-docs.txt` be folded into `requirements-dev.txt`?

jhelferty-nv:
I have since noticed that there's a `docs/requirements.txt` that includes all of these except `pybind11_mkdoc`. For now I'm going to update the documentation to point users at `docs/requirements.txt` instead of `requirements-doc.txt`, include a manual step to install `pybind11_mkdoc`, and delete `requirements-doc.txt`.

The question of how to deal with `pybind11_mkdoc` long term still needs to be addressed though.

bmillsNV:
@aidanfnv can you take a look at this?

PR #372, opened 2025-07-24
Update documented dependencies

jhelferty-nv:
- Adds explicit documentation of optional dependencies (cuda, pytorch, and Xcode for the metal compiler on macOS)
- Updates the macOS build steps, based on my recent experience installing from a clean system.
- Removes duplicate docs requirements file.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,4 +1,0 @@
-pybind11_mkdoc
-nbsphinx
-sphinx
-furo
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,7 +27,18 @@ See :ref:`developer guide <sec-compiling>` for more detailed information on how 
 Requirements
 ------------
 
-- **python >= 3.9**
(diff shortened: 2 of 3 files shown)


requirements-docs.txt
jkwak-work, line 1: These are not required?

jkwak-work (approved):
Looks good to me

requirements-docs.txt
jhelferty-nv: They were already present in `docs/requirements.txt` so this file (which I added in a previous commit) was redundant.

There was one difference; `requirements-doc.txt` included `pybind11_mkdoc`, but I've instead made installing that an explicit step in the documentation for how to build the API Reference.

In an ideal world, `pybind11_mkdoc` would be in the `requirements-dev.txt`, but I'd like to confirm that that isn't going to cause a bunch of dependency bloat (it depends on python `clang` package) before I do it. #357 to follow-up.

PR #374, opened 2025-07-27
Switch slangpy to Slang 2025.13.1

tdavidovicNV:
Moved to Slang 2025.13.1, to fix Slang bug with inheriting from IDifferentiable.

- Added pragma warning disable to newly identified places (consulted with the Slang team that the code itself was correct).
- The pragma does not fully work due to https://github.com/shader-slang/slang/pull/7942 so the warnings can still appear in the output for the time being.


ccummingsNV (approved):

ccummingsNV:
NOTE: Approval given, but bypassing tests check due to sporadic failures that are known to be unrelated.

PR #382, opened 2025-07-28
Hot reload fixes

tunabrain:
This fixes a couple of issues with hot reload:
1) Lookup modules are not reloaded after a hot reload, and trying to create an `NDBuffer` or `Tensor` after a hot reload has taken place will fail. The fix is to add a hot reload hook for lookup modules.

2) Changing a loaded module, then e.g. evaluating a cell in Jupyter will still run the old shader code. The hot reload trigger runs on dispatch, and the shader is reloaded only _after_ the next dispatch. This can be confusing when things don't run in a loop, e.g. in an interactive environment, as you need to run a cell twice to see the reloaded code. The fix is to re-add a jupyer hook that triggers a hot reload check before jupyter cells are evaluated.

This exposes hot reload functionality on the device. Input on how to expose this differently is welcome!


src/slangpy_ext/device/device.cpp
```diff
@@ -866,6 +867,17 @@ SGL_PY_EXPORT(device_device)
...
         "offset"_a
     );
+    device.def(
+        "set_hot_reload_delay",
+        [](Device* self, uint32_t delay_ms) { self->_hot_reload()->set_auto_detect_delay(delay_ms); },
+        "timeout_ms"_a,
+        D_NA(Device, set_hot_reload_delay)
+    );
+    device.def(
+        "hot_reload",
```
ccummingsNV: This check should be called be something like "hot_reload_update" or "hot_reload_check". i.e. it shouldn't imply a hot reload will occur when you call it, because it might do nothing.

ccummingsNV (dismissed):
Seems sensible - just a naming issue then should be good to go

src/slangpy_ext/device/device.cpp
tunabrain: Sounds good! PR adjusted to include this change.

skallweitNV (approved):

PR #391, opened 2025-07-30
Add synchronization before copy in copy_from_torch() tests

aidanfnv:
Related to #113 

The tests for `copy_from_torch()` that I added in #363 are causing intermittent failures the Windows CI tests.
The failure appears to be specifically from the first partial copy in `test_partial_torch_copy()` in `test_buffer_views.py`, and only in Vulkan when copying to an `NDBuffer` (and not a `Tensor`), where the results are zeros instead of the copied data.

I suspect this to be the result of a race condition, so this change adds `sync_to_cuda()` and `sync_to_device()` calls before the `copy_from_torch()` calls. I think that will fix the failures because in #363 I was seeing multiple of the partial copies with zeroes until I added the sync calls in between the `copy_from_torch()` calls, and I suspect that something (perhaps `torch.randn(...).cuda()`?) is not finished before the first copy but is finished before the second copy due to those sync calls from #363.

I will test the Windows CI checks repeatedly to verify that this eliminates the intermittent failures.

### DIFF
--- a/slangpy/tests/slangpy_tests/test_buffer_views.py
+++ b/slangpy/tests/slangpy_tests/test_buffer_views.py
@@ -205,6 +205,8 @@ def test_full_torch_copy(device_type: DeviceType, buffer_type: Union[Type[Tensor
     torch_ref = torch.randn(shape, dtype=torch.float32).cuda()
     usage = BufferUsage.shader_resource | BufferUsage.unordered_access | BufferUsage.shared
     buffer = buffer_type.zeros(device, dtype="float", shape=shape, usage=usage)
+    device.sync_to_cuda()
+    device.sync_to_device()
 
     buffer.copy_from_torch(torch_ref)
     device.sync_to_cuda()
@@ -256,6 +258,8 @@ def test_partial_torch_copy(
     torch_ref = torch.randn(shape, dtype=torch.float32).cuda()
     usage = BufferUsage.shader_resource | BufferUsage.unordered_access | BufferUsage.shared
     buffer = buffer_type.zeros(device, dtype="float", shape=shape, usage=usage)
+    device.sync_to_cuda()
+    device.sync_to_device()
 
     for i in range(shape[0]):
         buffer[i].copy_from_torch(torch_ref[i])
@@ -310,6 +314,8 @@ def test_torch_copy_errors(
 
     usage = BufferUsage.shader_resource | BufferUsage.unordered_access | BufferUsage.shared
     buffer = buffer_type.zeros(device, dtype="float", shape=shape, usage=usage)
+    device.sync_to_cuda()
+    device.sync_to_device()
 
     with pytest.raises(Exception, match=r"Tensor is larger"):
         tensor = torch.zeros((shape[0], shape[1] + 1), dtype=torch.float32)


aidanfnv:
Adding calls to `device.sync_to_cuda()` and `device.sync_to_device()` before the first torch copy results did indeed reintroduce the deadlock that I saw in my other PR, so now I am trying a different synchronization pattern, the one used in `test_torchbuffers.py`.
It does not deadlock, but I still need to test if the failures are intermittent or eliminated.

aidanfnv:
Currently the CI seems to fail almost every other run, but with this PR I was able to run the CI checks 8 times without any failures.

ccummingsNV:
Yeah:

```
void Device::sync_to_cuda(void* cuda_stream)
{
    // Signal fence from CUDA, wait for it on graphics queue.
    if (m_supports_cuda_interop) {
        SGL_CU_SCOPE(this);
        uint64_t signal_value = m_global_fence->update_signaled_value();
        m_cuda_semaphore->signal(signal_value, CUstream(cuda_stream));
        m_wait_global_fence = true;  <<<<<<< Just sets bool telling next submit to wait for the semaphore
    }
}
```

And

```
void Device::read_buffer_data(const Buffer* buffer, void* data, size_t size, size_t offset)
{
    SGL_CHECK_NOT_NULL(buffer);
    SGL_CHECK(offset + size <= buffer->size(), "Buffer read is out of bounds");
    SGL_CHECK_NOT_NULL(data);

    <<<<<< Goes straight to RHI, bypassing the submit process
    SLANG_RHI_CALL(m_rhi_device->readBuffer(buffer->rhi_buffer(), offset, size, data));
}
```

This would impact any api call that doesn't utilize slangpy submit, of which there are many. The fix would be to mimic what sync_to_device does, and immediately insert a fence wait into the gfx queue. Will take a look at this today and either do it, or feedback here how to do it.

ccummingsNV:
In fact, the same issue exists in reverse - sync_to_cuda only guaruntees the most recent submit is finished, so any operation that doesn't utilize submit will be bypassed.

aidanfnv:
Okay, it all makes more sense to me now.
I have reverted back to using `device.sync_to_cuda()` and `device.sync_to_device()`, but without the double syncs. I am also now using `get_torch_device`.

aidanfnv:
I fixed the race condition where the first copy is zeroes instead of copied data by waiting for the zeroing to finish before starting the copy I also un-skipped the CUDA device tests now that they should work, which revealed a CUDA-specific issue in my `copy_from_torch()` implementation, for which I also added the fix.

skallweitNV:
@aidanfnv I have merged a fix to the race condition and tests are working fine now. Maybe rebase this PR if there is anything else worth merging?

aidanfnv:
The secondary content of this PR, fixing copy_from_torch() for CUDA devices and un-skipping the relevant tests is still worth merging.

src/slangpy_ext/utils/slangpystridedbufferview.cpp
```diff
@@ -441,13 +438,16 @@ void StridedBufferView::copy_from_torch(nb::object tensor)
...
-        void* dst_data = reinterpret_cast<uint8_t*>(m_storage->cuda_memory()) + byte_offset;
-
         // Validate memory bounds - use accurate error message for direct tensor operations
         size_t buffer_size = m_storage->size() - byte_offset;
         SGL_CHECK(tensor_bytes <= buffer_size, "Tensor is larger than the buffer ({} > {})", tensor_bytes, buffer_size);
 
-        // Use proper CUDA device-to-device memory copy
+        // If we're not using a CUDA device, switch to the device's CUDA context
+        if (m_storage->device()->type() != DeviceType::cuda) {
+            SGL_CU_SCOPE(m_storage->device());
```
skallweitNV: This is popping the CUDA context too early. I think you can just drop the `if` statement here because you already are in a branch that checked for `m_storage->cuda_memory() != nullptr`.

src/slangpy_ext/utils/slangpystridedbufferview.cpp
```diff
@@ -441,13 +438,16 @@ void StridedBufferView::copy_from_torch(nb::object tensor)
...
+        }
```
skallweitNV:
```suggestion
        SGL_CU_SCOPE(m_storage->device());
```

skallweitNV (changes requested):

src/slangpy_ext/utils/slangpystridedbufferview.cpp
aidanfnv: I found that if I remove the `if` statement, the function results in a segfault. Moving `SGL_CU_SCOPE(m_storage->device());` any later also results in similar failures.

skallweitNV: Ok, that's very strange. I'd have to step through locally to figure out what's going on ..

skallweitNV: Pushed a fix for the context issue.

PR #392, opened 2025-07-31
Fix epic device sync race condition

ccummingsNV:


src/sgl/device/device.cpp
```diff
@@ -673,16 +666,60 @@ void Device::wait_for_idle(CommandQueueType queue)
...
+        uint64_t signal_value = m_global_fence->update_signaled_value();
+
+        // Signal it from CUDA.
+        m_cuda_semaphore->signal(signal_value, CUstream(cuda_stream));
+
+        // Wait for it on device.
+        rhi::IFence* fences[] = {m_global_fence->rhi_fence()};
+        uint64_t fence_values[] = {signal_value};
+        rhi::SubmitDesc submit_desc{
+            .commandBuffers = nullptr,
```
skallweitNV, line 682: you could rely more on the default values of `SubmitDesc` and just specify fence related fields, just a suggestion

skallweitNV (approved):

PR #394, opened 2025-07-31
Improve testing stability

skallweitNV:


slangpy/tests/device/test_buffer.py
```diff
@@ -53,7 +53,7 @@ def test_buffer_init_data(device_type: spy.DeviceType):
         "structured_buffer_uint",
     ],
 )
-@pytest.mark.parametrize("size_MB", [128, 1024, 2048, 3072, 4096])
+@pytest.mark.parametrize("size_MB", [128, 1024, 2048, 3072, 4094])
```
tdavidovicNV, line 56: 4094 is a weird number? The (deleted) note below says VULKAN supports 4GB-1B, so this test 4GB-2MB, to give it a sane buffer?

slangpy/tests/device/test_device.py
```diff
@@ -114,7 +114,7 @@ def test_hot_reload_event(device_type: spy.DeviceType):
     # Load a shader
     program = device.load_program(
         module_name="test_shader_foo.slang",
-        entry_point_names=["main_a", "main_b", "main_vs", "main_fs"],
+        entry_point_names=["main_a"],
```
tdavidovicNV, line 117: Why is this reduced?

tdavidovicNV (approved):
LGTM, just confused about two test reductions.

slangpy/tests/device/test_buffer.py
skallweitNV: It actually didn't. It supports 4094 MB according to validation layers. We need to add a way to determine the max buffer size more reliable, but that's for later.

slangpy/tests/device/test_device.py
skallweitNV: Because it's not valid to have two compute entry points (same shader stage) in a single linked program. Validation layers complained.

PR #395, opened 2025-07-31
Refactored the bool setters and getters

tdavidovicNV:
This is another stab at refactoring the getters and setters so things work for various strides and sizes (mostly bool).

We cannot establish stride of the element on the stride nor size of the vector, because on Metal int3 is 16B size and stride. What we do instead is we use the knowledge of the bug, and if the type is cuda bool2-4 and the total size of the type is `sizeof(int) * dimension`, we just set the stride to 4, because we know the backend implementation.
Hopefully this will allow seamless transition to the new Slang code that implements `boolX` as a struct with `X` bools.

`bool1` is still skipped, because it is just implemented very differently (as `bool` rather than `int1`) so even the type size and alignment do not match the reflection data.

Also fixed a bug where the source scalar type was actually deduced from the Slang type layout, rather than from the ndarray data source.


slangpy/tests/device/test_shader_cursor.py
```diff
@@ -35,26 +35,22 @@ class TypeInfo:
...
     "int16_t": TypeInfo(size=4, struct="hxx", dtype=np.int16),
     "uint16_t": TypeInfo(size=4, struct="Hxx", dtype=np.uint16),
     "float16_t": TypeInfo(size=4, struct="exx", dtype=np.float16),
-    "int64_t": TypeInfo(size=8, struct="q", dtype=np.int64),
-    "int64_t": TypeInfo(size=8, struct="Q", dtype=np.uint64),
-    "float64_t": TypeInfo(size=8, struct="d", dtype=np.float64),
+    ## Unused for now
+    # "int64_t": TypeInfo(size=8, struct="q", dtype=np.int64),
+    # "int64_t": TypeInfo(size=8, struct="Q", dtype=np.uint64),
+    # "float64_t": TypeInfo(size=8, struct="d", dtype=np.float64),
```
ccummingsNV, line 50: Does this mean we support these types but don't test them? If so, maybe just add them to the tests?

ccummingsNV (approved):
LGTM - just one query about the tests

slangpy/tests/device/test_shader_cursor.py
tdavidovicNV: I just commented them out for now, because this is the only part of the whole test infrastructure that can "handle" them.

The Slang code cannot bring back value of 8B types (it does `asuint` cast to bring back exact bit image of the on-device memory, and `__asuint64` doesn't seem to be supported). So to me this is just a bit of confusing futureproofing. 

We do test `int64` and `uint64` in the `test_buffer_cursor`, but there the way the test is verified uses the cursor to also bring the values back (rather than write them to `RWStructuredBuffer<uint32_t>`).

PR #396, opened 2025-07-31
Fix exception for multi-dimensional indexing on buffers

tunabrain:
Buffer indexing fails when indexing with multiple indices (e.g. `buffer[3, 4]`), because `stridedbufferview` expects these to be passed as `*args`. However, python groups them into a tuple instead, and nanobind passes it through as `nb::args(nb::tuple(3, 4))` causing an exception during indexing.

This adds a fix, modifies the tests to catch this and adds a better error message.


skallweitNV (approved):
LGTM

PR #397, opened 2025-08-04
Fix CUDA race condition

skallweitNV:
Based on #392 with addition fixes in the affected python tests.

- `Device::sync_to_cuda` and `Device::sync_to_device` now synchronize immediately instead of preparing a sync for the next submission
- Do not free external memory
- Fix `test_buffer_views` CUDA interop tests


tdavidovicNV (approved):
LGTM

PR #399, opened 2025-08-04
Fixed the way entry points are looked up to avoid exceptions

tdavidovicNV:
Entry points are no longer looked up using exceptions, but instead by searching the list of entry points.


skallweitNV (approved):
LGTM

PR #410, opened 2025-08-08
Enable shapes and instances tests

### LINKED ISSUE #407: Tensor.with_grads should default to zeroing the grads
The `zero` option currently defaults to `False` but it would be safer to set it to `True`.

At least on CUDA, with simple slangpy scripts users will get back zeros, but then as memory is reclaimed on longer runs, they will not get zeros back if not explicitly enabled with the flag. We should try to be safe by default and opt for performance if required.

When changing the default we should make sure that existing calls `with_grads(zero=True)` are replaced with just `with_grads()`. There are a bunch in the tests at least.

ccummingsNV:
Enable shapes + instances tests that were broken due to uninitialized buffers
Add extra uninitialized buffer fixes

Fixes #407

### DIFF
--- a/slangpy/tests/slangpy_tests/test_custom_types.py
+++ b/slangpy/tests/slangpy_tests/test_custom_types.py
@@ -363,8 +363,9 @@ void add_to_bucket(int id, RWByteAddressBuffer bucket, float value) {{
     buckets = NDBuffer(element_count=bucket_size, device=device, dtype=int)
     buckets.clear()
 
-    # Run bucketer with 1B random floats
-    count = 1 * 1000 * 1000 * 1000
+    # Run bucketer with 1B random floats (takes ages on CUDA so making it just 1M)
+    # count = 1 * 1000 * 1000 * 1000
+    count = 1 * 1000 * 1000
     bucket_values(grid((count,)), buckets.storage, rand_float())
 
     # Verify the distribution of values in range [0,1) is roughly even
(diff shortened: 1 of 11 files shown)


slangpy/tests/slangpy_tests/test_custom_types.py
```diff
@@ -363,8 +363,9 @@ def test_rand_float_uniformity(device_type: DeviceType):
     buckets = NDBuffer(element_count=bucket_size, device=device, dtype=int)
     buckets.clear()
 
-    # Run bucketer with 1B random floats
-    count = 1 * 1000 * 1000 * 1000
+    # Run bucketer with 1M random floats
+    # TODO: Find out why this took insanely long to run on CUDA with 1B floats
```
skallweitNV, line 367: Is this true? On my machine this wasn't slow IIRC.

ccummingsNV: Weird - yeah on my machine it takes multiple minutes

skallweitNV: I can check again once this is in.

ccummingsNV: around 90s, same for both gpus.

skallweitNV (approved):
LGTM

ccummingsNV:
Removed zero=True in with grads calls + fixed extra test post approval.

PR #418, opened 2025-08-11
Fix the aglinment issue for tensor

kaizhangNV:
When copy data to GPU Tensor, if the element type of the Tensor is vector type, we have to special handle the alignment issue because Metal has different alignment with other platforms.


jkwak-work:
Windows Debug CI was failing on "Unit test (C++)".
I restarted the test to see if it is intermittent.

skallweitNV:
I think this needs more work. The current `copy_from_numpy` implementation doesn't do ANY checks on the shape or type of the passed in tensor. All it does is checking that the memory fits. So the initial implementation probably needs some work to begin with. The added padding logic can probably go wrong in so many weird ways (if shape or dtype doesn't match) that it can be very confusing for users.

@ccummingsNV can you take a look at this when you're back?

Issue #420, opened 2025-08-12
Support bindless storage buffers


venkataram-nv:
Currently slangpy supports bindless textures but not not bindless storage buffers. The slang-rhi does implement this feature, so this should be part of slangpy as well.

PR #421, opened 2025-08-12
Add support for bindless storage buffers

venkataram-nv:
Addresses #420

### DIFF
--- a/src/sgl/device/resource.cpp
+++ b/src/sgl/device/resource.cpp
@@ -229,6 +229,24 @@ NativeHandle Buffer::shared_handle() const
     return NativeHandle(rhi_handle);
 }
 
+DescriptorHandle Buffer::descriptor_handle_ro(Format format, BufferRange range) const
+{
+    rhi::DescriptorHandle rhi_handle = {};
+    rhi::Format rhi_format = format == Format::undefined ? static_cast<rhi::Format>(m_desc.format) : static_cast<rhi::Format>(format);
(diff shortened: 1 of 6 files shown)


src/sgl/device/resource.cpp
```diff
@@ -229,6 +229,24 @@ NativeHandle Buffer::shared_handle() const
     return NativeHandle(rhi_handle);
 }
 
+DescriptorHandle Buffer::descriptor_handle_ro(Format format, BufferRange range) const
```
kaizhangNV: why do we need this "ro"/"rw" suffix? The access permission should be able to query from the buffer type itself.

venkataram-nv: Textures have a similar interface, so Ive ported that to the buffers for consistency. Also, even if we can query this, couldn't there be a case where a rw buffer is being used ro only?

ccummingsNV:
This looks fairly valid (other than formatting - worth enabling the pre-commit hooks). @skallweitNV is this how you envisioned the basic bindless api working? I think we should make it more automatic when using, eg, buffer or shader cursor, but that can be another PR

src/sgl/device/resource.cpp
skallweitNV: In D3D12 for example you need to create different descriptors for SRV (read-only) and UAV (read-write) and this is independent of the buffer usage flags (shader-resource, unordered-access) because as @venkataram-nv mentioned, a buffer with unordered-access can also be bound as an SRV.

PR #422, opened 2025-08-13
Fix slang warnings using a workaround

skallweitNV:


tdavidovicNV (approved):

skallweitNV:
For some reason this consistently breaks on macOS debug only.

PR #440, opened 2025-08-22
Initial setup for slangpy benchmarks

skallweitNV:
Add a minimal system for GPU benchmarking of slangpy code:
- Benchmarks are in `slangpy/benchmarks`
- Benchmarks are run with `pytest slangpy/benchmarks`
- At the end of pytest execution, a table summary of all the benchmarks is shown and a `benchmark_report.json` is dumped

Add three benchmarks using slangpy `Tensor` in `test_benchmark_tensor.py`. The `test_tensor_sum_indirect` benchmark reproduces the Vulkan perf issue.

### DIFF
--- a/slangpy/benchmarks/bench/__init__.py
+++ b/slangpy/benchmarks/bench/__init__.py
@@ -1,6 +1,5 @@
 # # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-from .report import Report
 from .benchmark import benchmark, BenchmarkFixture
 
-__all__ = ["Report", "benchmark", "BenchmarkFixture"]
+__all__ = ["benchmark", "BenchmarkFixture"]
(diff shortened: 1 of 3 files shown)


slangpy/benchmarks/helpers.py
```diff
@@ -0,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
ccummingsNV, line 1: I feel horrific that we're creating a 3rd semi-copy-and-paste of these helpers. But I guess we don't want to conflate fixing this with helper changes

ccummingsNV (approved):
More copies of helpers.py makes me sad, but otherwise looks good

slangpy/benchmarks/helpers.py
skallweitNV: yeah, we should consolidate this at some point

PR #449, opened 2025-08-25
Add `link` modules of a Module are now linked for layout. Added option to relink.

tdavidovicNV:
The `link` modules are added to the `combined_program` of `Module` on creation, so the contained data types can be found in the Module's layout.

This is needed when the Module depends on other code-generated modules it itself cannot `import`.

Added `relink_if_different` function that relinks the program and clears the caches, if the newly provided link list is different from the one already in the module.

The compilation of a program to get a layout should be replaced by using `createCompositeComponentType` instead of compiling all the way down to a program.

### DIFF
--- a/slangpy/core/module.py
+++ b/slangpy/core/module.py
@@ -62,8 +62,8 @@ class Module:
         self.slangpy_device_module = device_module.session.load_module("slangpy")
 
         # Extract linked modules
-        self.link = [x.module if isinstance(x, Module) else x for x in link]
-        self.link.sort(key=id)
+        self.link_set = set([x.module if isinstance(x, Module) else x for x in link])
+        self.link = list(self.link_set)
 
         #: Reflection / layout information for the module.
         # Link the user- and device module together so we can reflect combined types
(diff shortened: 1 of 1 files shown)


slangpy/core/module.py
```diff
@@ -61,17 +61,20 @@ def __init__(
         #: The slangpy device module.
         self.slangpy_device_module = device_module.session.load_module("slangpy")
 
+        # Extract linked modules
+        self.link_set = set([x.module if isinstance(x, Module) else x for x in link])
```
skallweitNV: `self.link_set` is not needed

tdavidovicNV: It makes checking if I need to recreate the module or not easier.

Otherwise I have to do something like `material_modules.issubset(set(module.link))`. I can do that, if you think that's fine

skallweitNV: I think that's better than storing redundant information in the module.

slangpy/core/module.py
```diff
@@ -61,17 +61,19 @@ def __init__(
         #: The slangpy device module.
         self.slangpy_device_module = device_module.session.load_module("slangpy")
 
+        # Extract linked modules
+        self.link = list(set([x.module if isinstance(x, Module) else x for x in link]))
```
skallweitNV, line 65: @ccummingsNV why do we filter out `SlangModule` here?

ccummingsNV: Not sure - looks like a bug.

tdavidovicNV: `link_program` requires `Sequence[SlangModule ]` so I guess it is just frontloading that?

PR #461, opened 2025-08-27
Support Attribute reflection

TH3CHARLie:
While working on my project I found the User-Attribute related APIs were commented out and they are quite useful tools for reflection. This PR adds basic support for Attribute reflection.

### DIFF
--- a/src/sgl/device/reflection.cpp
+++ b/src/sgl/device/reflection.cpp
@@ -47,6 +47,7 @@ namespace detail {
     SGL_FROM_SLANG(VariableLayoutReflection);
     SGL_FROM_SLANG(EntryPointLayout);
     SGL_FROM_SLANG(ProgramLayout);
+    SGL_FROM_SLANG(Attribute);
 
 #undef SGL_FROM_SLANG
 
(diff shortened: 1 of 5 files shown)


src/sgl/device/reflection.cpp
```diff
@@ -161,6 +162,20 @@ ref<const DeclReflection> DeclReflection::find_first_child_of_kind(Kind kind, st
...
+    str += fmt::format("  name={},\n", name());
+    uint32_t count = argument_count();
+    str += fmt::format("  argument_count={},\n", count);
+    for (uint32_t i = 0; i < count; i++) {
+        str += fmt::format("  argument_{}={},\n", i, argument_type(i));
+    }
+    str += ")";
+    return str;
+}
+
```
ccummingsNV, line 181: Can you see if this can be done more cleanly with fmt - it has a lot of tools for building large strings. I think it might even be able to handle the loop, but if not this should be a single call to format for all the rest of the text, and a single call per argument. See the other 'to_string' functions in code for examples.

ccummingsNV (changes requested):
Only required change is to try and clean up that format function a bit.

If you were feeling awesome, you could add an attribute list mechanism, similar to how the other list types work, but not required - maybe one for another PR :)

ccummingsNV (approved):
Nice job - LGTM (I'll push it assuming merge + tests pass)

TH3CHARLie:
The PR should be ready to merge with tests passed

PR #463, opened 2025-08-28
Add some string conversion functions

ccummingsNV:


ccummingsNV:
[self-approve: just string conversion functions, no real functionality changes]

PR #464, opened 2025-08-29
Option to print generated shaders to TTY

ccummingsNV:
Mainly for AI development, adds env variable based option to slangpy to have generated shaders dumped to the TTY, so the agent can see what the code gen is producing.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -31,7 +31,10 @@ Testing:
     - The Python testing system uses PYTEST
     - The C++ testing system uses doctest
     - Always build before running tests.
-    - To run all Python tests, run "pytest slangpy/tests"
+    - To run all Python tests, run "pytest slangpy/tests -v"
+    - An example of running a specific set of tests in a file is: "pytest slangpy/tests/slangpy_tests/test_shader_printing.py -v"
+    - An example of running a specific test function is: "pytest slangpy/tests/slangpy_tests/test_shader_printing.py::test_printing -v"
+    - To log any generated shaders in a test, set the SLANGPY_PRINT_SHADERS environment variable to "true". For example: "$env:SLANGPY_PRINT_SHADERS=1"true"; pytest slangpy/tests/slangpy_tests/test_shader_printing.py -v" (powershell syntax)
 
 C++ Code style:
     - Class names should start with a capital letter.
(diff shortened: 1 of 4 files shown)


.github/copilot-instructions.md
```diff
@@ -31,7 +31,10 @@ Testing:
     - The Python testing system uses PYTEST
```
skallweitNV, line 31: Unrelated but copilot now also picks up `AGENTS.md` in the root, so we might start consolidating the agent instructions into a shared file. I think CLAUDE still doesn't pick it up, but many other agents do.

slangpy/core/calldata.py
```diff
@@ -33,6 +33,11 @@
 
 _DUMP_GENERATED_SHADERS = False
 _DUMP_SLANG_INTERMEDIATES = False
+_PRINT_GENERATED_SHADERS = os.environ.get("SLANGPY_PRINT_SHADERS", "false").lower() in (
```
skallweitNV: should the env variable be named `SLANGPY_PRINT_GENERATED_SHADERS` to be more consistent with the `set_print_generated_shaders`?

skallweitNV: also for completeness, should we also have `SLANGPY_DUMP_GENERATED_SHADERS` and `SLANGPY_DUMP_SLANG_INTERMEDIATES` env variables?

slangpy/tests/slangpy_tests/test_shader_printing.py
```diff
@@ -0,0 +1,135 @@
...
+    # Enable shader printing
+    set_print_generated_shaders(True)
+
+    # Capture stdout to verify the shader is printed
+    captured_output = io.StringIO()
+    with redirect_stdout(captured_output):
+        # Just verify it can be called with no exceptions (like test_call_function)
+        function(5, 10)
+
+    output = captured_output.getvalue()
```
skallweitNV, line 36: In the past I found that capturing is not super reliable. We can enable these tests, but lets quickly skip them in case they are flakey.

slangpy/tests/slangpy_tests/test_shader_printing.py
```diff
@@ -0,0 +1,135 @@
...
+        else:
+            os.environ["SLANGPY_PRINT_SHADERS"] = original_env
+        set_print_generated_shaders(False)
+
+
+def test_environment_variable_parsing():
+    """Test that various environment variable values are parsed correctly."""
+
+    # Test various true values
+    true_values = ["true", "1", "yes", "TRUE", "True", "YES"]
```
skallweitNV: this feels a bit crazy to me .. i would just support `0` and `1` TBH

ccummingsNV: I'll compromise on using tolower and only being "true" or "1" :)

The number of times I've confused myself with env variables not sure if true==actually true or is just being treated as not 1...

skallweitNV: yeah, that's why I stick to `0/1` :) I just think a test for this is kinda too much

ccummingsNV: yup - removed the test

ccummingsNV:
[self-approve: minor changes to address PR]

PR #465, opened 2025-08-29
Benchmarks

ccummingsNV:


slangpy/benchmarks/conftest.py
```diff
@@ -1,3 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+# Importing these here means we don't have to explicitly import them into every benchmark
```
skallweitNV, line 3: True, but you still import the typings so IMO importing both is probably cleaner, but not fussy

slangpy/benchmarks/test_benchmark_interop.py
```diff
@@ -0,0 +1,449 @@
...
+    BUFFER_SIZE = 10
+    import torch
+
+    buffer0 = torch.randn((BUFFER_SIZE,), dtype=torch.float32, device="cuda")
+    buffer1 = torch.randn((BUFFER_SIZE,), dtype=torch.float32, device="cuda")
+
+    def tensor_addition(a: torch.Tensor, b: torch.Tensor):
+        res = a + b
+
+    benchmark_python_function(device, tensor_addition, a=buffer0, b=buffer1)
```
skallweitNV, line 121: Wouldn't it be better to not have to pass a device here? It's not used for the actual benchmark so it would be cleaner to not having to pass it.

slangpy/testing/benchmark/__init__.py
```diff
@@ -1,5 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# pyright: reportUnusedImport=false
```
skallweitNV, line 2: wouldn't it be better to use `__all__` as before then you can remove this and also `from benchmark import *` works

slangpy/testing/benchmark/fixtures.py
```diff
@@ -4,20 +4,62 @@
...
+class ReportFixture:
 
     def __init__(self, config: pytest.Config, node: Any):
         super().__init__()
         self.config = config
         self.node = node
 
+    def __call__(
+        self,
+        device: spy.Device,
```
skallweitNV: Should the report fixture not take a `device` in but an additional `meta` parameter?

slangpy/testing/benchmark/fixtures.py
```diff
@@ -48,31 +90,105 @@ def __call__(
...
+
+class BenchmarkPythonFunction:
+
+    def __init__(self, report_fixture: ReportFixture):
+        super().__init__()
+        self.report_fixture = report_fixture
+
+    def __call__(
+        self,
+        device: spy.Device,
```
skallweitNV: We should remove this IMO.

skallweitNV (changes requested):

slangpy/benchmarks/conftest.py
skallweitNV: Hm, this avoids having to do the `# type: ignore` workaround so might still be worth it.

ccummingsNV: The issue is that because the benchmark fixture used the report fixture, the tests had to ALSO import 'report', because of how pytest works, even though they didn't use it at all, which is really ugly.

slangpy/testing/benchmark/__init__.py
ccummingsNV: I hate '__all__' personally :) 

Maybe I've misunderstood, but as far as I can tell it's just a way of overriding what the module exports by default. But as it stands, it is just code duplication, as we export everything it imports.

ccummingsNV: Or does from benchmark import * not work at all without __all__?

slangpy/testing/benchmark/fixtures.py
ccummingsNV: I've got scenarios where the adapter is relevant to the test, so I'll pass it in, but make it optional. i.e. just because a test is running a python function doesn't mean the python function isn't doing work that's affected by the gpu in any way.

slangpy/testing/benchmark/fixtures.py
ccummingsNV: For now, that'd make life harder, as I have functions using the report functionality that want to submit the device info, but I can start with it optional.

ccummingsNV:
[self-approve: Minor changes for PR]

PR #466, opened 2025-08-29
Allows overriding of the TRACKING defines from cmake

tdavidovicNV:
The goal is to allow enabling tracking without directly modifying the used SlangPy version.

### DIFF
--- a/src/sgl/core/object.h
+++ b/src/sgl/core/object.h
@@ -21,7 +21,9 @@ static_assert(sizeof(Py_ssize_t_) == sizeof(size_t));
 /// Enable/disable object lifetime tracking.
 /// When enabled, each object derived from Object will have its
 /// lifetime tracked. This is useful for debugging memory leaks.
+#ifndef SGL_ENABLE_OBJECT_TRACKING
 #define SGL_ENABLE_OBJECT_TRACKING 0
+#endif // SGL_ENABLE_OBJECT_TRACKING
 
 /// Enable/disable reference tracking.
 /// When enabled, all references to an object that has reference tracking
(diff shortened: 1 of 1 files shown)


src/sgl/core/object.h
```diff
@@ -21,15 +21,19 @@ static_assert(sizeof(Py_ssize_t_) == sizeof(size_t));
 /// Enable/disable object lifetime tracking.
 /// When enabled, each object derived from Object will have its
 /// lifetime tracked. This is useful for debugging memory leaks.
+#ifndef SGL_ENABLE_OBJECT_TRACKING
 #define SGL_ENABLE_OBJECT_TRACKING 0
+#endif // SGL_ENABLE_OBJECT_TRACKING
```
skallweitNV:
```suggestion
#endif
```

src/sgl/core/object.h
```diff
@@ -21,15 +21,19 @@ static_assert(sizeof(Py_ssize_t_) == sizeof(size_t));
...
 
 /// Enable/disable reference tracking.
 /// When enabled, all references to an object that has reference tracking
 /// enabled using set_enable_ref_tracking() are tracked. Each time the reference
 /// count is increased, the current stack trace is stored. This helps identify
 /// owners of objects that are not properly releasing their references.
 /// Optionally all references can be tracked by setting SGL_TRACK_ALL_REFS.
+#ifndef SGL_ENABLE_REF_TRACKING
 #define SGL_ENABLE_REF_TRACKING 0
+#endif // SGL_ENABLE_REF_TRACKING
```
skallweitNV:
```suggestion
#endif
```

skallweitNV (dismissed):
I'd remove those trailing comments, they are useful when the `#endif` is far apart from the `#if` but that's not the case here

PR #472, opened 2025-09-01
Re-enable tests, disable torch and add trimmed mean

ccummingsNV:
Tweaks for benchmarking
- Re-enabled a load of the tests in interop
- Not using get_torch_device when not needed, and disabled tests that do need it
- Use trimmed mean 
- Much larger test set

### DIFF
--- a/slangpy/benchmarks/test_benchmark_interop.py
+++ b/slangpy/benchmarks/test_benchmark_interop.py
@@ -106,6 +106,7 @@ void compute_main(uint3 tid: SV_DispatchThreadID, StructuredBuffer<float> a, Str
 """
 
 
+@pytest.mark.skip(reason="See if pytorch causes perf issues")
 def test_pytorch_tensor_addition_cpu(benchmark_python_function: BenchmarkPythonFunction):
     device = helpers.get_device(spy.DeviceType.cuda)
 
@@ -196,6 +197,7 @@ def test_slangpy_tensor_addition_appendonly_cpu(
     benchmark_python_function(device, tensor_addition, a=buffer0, b=buffer1, res=resbuffer)
 
 
+@pytest.mark.skip(reason="See if pytorch causes perf issues")
 def test_pytorch_tensor_addition_gpu_est(report: ReportFixture):
     device = helpers.get_device(spy.DeviceType.cuda)
 
--- a/slangpy/benchmarks/test_benchmark_interop.py
+++ b/slangpy/benchmarks/test_benchmark_interop.py
@@ -106,7 +106,7 @@ void compute_main(uint3 tid: SV_DispatchThreadID, StructuredBuffer<float> a, Str
 """
 
 
-@pytest.mark.skip(reason="See if pytorch causes perf issues")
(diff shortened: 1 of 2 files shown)


slangpy/benchmarks/test_benchmark_interop.py
```diff
@@ -108,8 +106,9 @@
 """
 
 
+@pytest.mark.skip(reason="Pytorch integration is affecting benchmarks")
 def test_pytorch_tensor_addition_cpu(benchmark_python_function: BenchmarkPythonFunction):
-    device = helpers.get_torch_device(spy.DeviceType.cuda)
+    device = helpers.get_device(spy.DeviceType.cuda)
```
skallweitNV, line 111: I think we shouldn't create a device at all here.

slangpy/benchmarks/test_benchmark_interop.py
```diff
@@ -127,7 +126,7 @@ def tensor_addition(a: torch.Tensor, b: torch.Tensor):
 def test_compute_addition_noalloc_cpu(
     device_type: spy.DeviceType, benchmark_python_function: BenchmarkPythonFunction
 ):
-    device = helpers.get_torch_device(device_type)
+    device = helpers.get_device(device_type)
```
skallweitNV, line 129: Technically, all these tests are not CUDA interop tests. So we should probably reorganize things a bit at some point.

slangpy/testing/benchmark/fixtures.py
```diff
@@ -67,8 +76,8 @@ def __call__(
         self,
         device: spy.Device,
         function: Union[spy.Function, FunctionNodeBwds],
-        iterations: int = 200,
-        warmup_iterations: int = 10,
+        iterations: int = 2000,
```
skallweitNV: We should probably have `DEFAULT_ITERATIONS` and `DEFAULT_WARMUP_ITERATIONS` to share between the functions. Also, doesn't 2000 TDR on stock settings?

slangpy/benchmarks/test_benchmark_interop.py
ccummingsNV: I agree - will do that soon.

slangpy/testing/benchmark/fixtures.py
ccummingsNV: TDR only happens if we do all dispatches in 1 submit

slangpy/benchmarks/test_benchmark_interop.py
ccummingsNV: This one's now disabled, but when we enable it, we need some means of reporting device info - i.e. it needs to report the gpu etc that the machine has. Simplest would be to call get_torch_device and use it

skallweitNV: We currently assume that the machine only has one GPU, that's the one being used :)

skallweitNV: This is a good test case for not reporting a device which the benchmark app needs to support.

skallweitNV (approved):

PR #473, opened 2025-09-01
Write timestamp improvements

skallweitNV:
- add `PassEncoder::write_timestamp`
- add ability to write timestamps in `ComputeKernel::dispatch`

### DIFF
--- a/src/sgl/device/kernel.cpp
+++ b/src/sgl/device/kernel.cpp
@@ -39,14 +39,27 @@ ComputePipeline* ComputeKernel::pipeline() const
     return m_pipeline;
 }
 
-void ComputeKernel::dispatch(uint3 thread_count, BindVarsCallback bind_vars, CommandEncoder* command_encoder)
+void ComputeKernel::dispatch(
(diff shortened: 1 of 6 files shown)


src/sgl/device/kernel.cpp
```diff
@@ -39,14 +39,27 @@ ComputePipeline* ComputeKernel::pipeline() const
     return m_pipeline;
 }
 
-void ComputeKernel::dispatch(uint3 thread_count, BindVarsCallback bind_vars, CommandEncoder* command_encoder)
+void ComputeKernel::dispatch(
+    uint3 thread_count,
+    BindVarsCallback bind_vars,
+    CommandEncoder* command_encoder,
+    QueryPool* query_pool,
```
ccummingsNV, line 46: I'd suggest it might be worth us wrapping the query parameters in a little structure, as it doesn't make sense to supply only some of them.

skallweitNV: I'll keep it like that for now as discussed

PR #475, opened 2025-09-01
Added SGL_CHECK_EQ and SGL_CHECK_NE

tdavidovicNV:


src/sgl/core/error.h
```diff
@@ -103,6 +103,8 @@ namespace detail {
 /// Helper for throwing an exception if a pointer is null.
 #define SGL_CHECK_NOT_NULL(arg) SGL_CHECK(arg != nullptr, "\"{}\" must not be null", #arg)
 
+#define SGL_CHECK_EQ(arg, value) SGL_CHECK(arg == value, "\"{}\" must be equal to {}", #arg, value)
+#define SGL_CHECK_NE(arg, value) SGL_CHECK(arg != value, "\"{}\" must be different from {}", #arg, value)
```
ccummingsNV: Nit picking a bit, but to be consistent this should say 'not equal to'

ccummingsNV (dismissed):

src/sgl/core/error.h
tdavidovicNV: I was debating that. The consistency is twofold. With the exception of the not null, all of the others tell you what the condition *must* be (and isn't). So this one would be the odd one out. But I was like 50/50 on it, so if you think `must not be equal` is the more consistent, I will go with that.

skallweitNV: IMO `must not be equal` is more consistent

skallweitNV (approved):

PR #476, opened 2025-09-01
Update rhi

ccummingsNV:


ccummingsNV:
[self-approve: just updating rhi sub repo]

PR #479, opened 2025-09-02
Handle change in heap report api

ccummingsNV:


docs/changelog.rst
```diff
@@ -7,6 +7,11 @@ Changelog
 
 SlangPy uses a `semantic versioning <http://semver.org>`__ policy for its API.
 
+Version 0.33.2 (Unreleased)
```
skallweitNV: next version should be `0.34.0` not sure why we started doing patch releases

src/slangpy_ext/device/device.cpp
```diff
@@ -239,6 +239,13 @@ SGL_PY_EXPORT(device_device)
 
     nb::class_<ShaderHotReloadEvent>(m, "ShaderHotReloadEvent", D(ShaderHotReloadEvent));
 
+    nb::class_<HeapReport>(m, "HeapReport", D_NA(HeapReport))
```
skallweitNV, line 242: I wondered about this when adding the compilation reports and didn't come to a conclusion.
It feels like it might be more useful to just return a dict here. This will be easier to print and work with on the python side. But it's a bit of a separation between the sgl C++ and Python side. Not sure how you feel about this. But like with this API `print(device.report_heaps())` will be very not useful. Then you can implement `__repr__` but then while you can print you still can't dump it to json for example.

ccummingsNV: I'd rather not do this just for HeapReport - if we want to go down that path we should do 1 PR / release with this done across the code to any relevant types rather than have a half/half api.

PR #481, opened 2025-09-02
Support passing call data as entry point parameter, and use it on CUDA

ccummingsNV:


slangpy/core/callsignature.py
```diff
@@ -509,7 +512,14 @@ def generate_code(
...
         cg.trampoline.append_line("[Differentiable]")
-    cg.trampoline.append_line(f"void {trampoline_fn}(Context __slangpy_context__)")
+
+    # For entry point mode, add CallData as a parameter to the trampoline function
+    if is_entry_point:
+        cg.trampoline.append_line(
+            f"void {trampoline_fn}(Context __slangpy_context__, CallData __calldata__)"
+        )
+    else:
+        cg.trampoline.append_line(f"void {trampoline_fn}(Context __slangpy_context__)")
```
tdavidovicNV, line 522: Could this be something like:

```
cg.trampoline.append_line(f"void {trampoline_fn}(Context __slangpy_context__{[", CallData __calldata__" if is_entry_point else ""]})")
```

(Same question goes for all the other if/else where it is mostly the same, but differs in one thing)

slangpy/core/callsignature.py
```diff
@@ -518,11 +528,18 @@ def generate_code(
...
+                data_name = (
+                    f"_param_{x.variable_name}"
+                    if x.create_param_block
+                    else f"__calldata__.{x.variable_name}"
+                )
+            else:
+                data_name = (
+                    f"_param_{x.variable_name}"
+                    if x.create_param_block
+                    else f"call_data.{x.variable_name}"
```
tdavidovicNV, line 541: Can we instead rename the ParameterBlock from `call_data` to `__call_data__` and leave this code the same? (The danger of name clash is actually larger on the global one than the entrypoint parameter)

slangpy/core/callsignature.py
```diff
@@ -560,11 +577,18 @@ def generate_code(
...
+                )
```
tdavidovicNV, line 591: another if based solely on different name as global and as entry point parameter

slangpy/core/callsignature.py
```diff
@@ -581,10 +605,16 @@ def generate_code(
...
-        "void compute_main(int3 flat_call_thread_id: SV_DispatchThreadID, int3 flat_call_group_id: SV_GroupID, int flat_call_group_thread_id: SV_GroupIndex)"
-    )
+    if is_entry_point:
+        cg.kernel.append_line(
+            "void compute_main(int3 flat_call_thread_id: SV_DispatchThreadID, int3 flat_call_group_id: SV_GroupID, int flat_call_group_thread_id: SV_GroupIndex, CallData call_data)"
+        )
+    else:
+        cg.kernel.append_line(
+            "void compute_main(int3 flat_call_thread_id: SV_DispatchThreadID, int3 flat_call_group_id: SV_GroupID, int flat_call_group_thread_id: SV_GroupIndex)"
+        )
```
tdavidovicNV, line 615: Maybe instead of ternary operator a variable like `extra_params` that is either empty or `, CallData call_data`

tdavidovicNV (approved):
LGTM. I think the codegen got less readable than it needs to be, and could be made a bit tighter with more abuse from the string formatting

slangpy/core/callsignature.py
ccummingsNV: it's a bit more complex than that, as the entry point and global are both called 'call_data', but the argument to the trampiline function is the __call_data__ one, as it's the only point at which user defined names could overlap with generated ones.

slangpy/core/callsignature.py
ccummingsNV: heeh - I find this version less readable :) I'll do another pass over this code soon though and try and tidy it up as a whole - I agree it's starting to suffer from 'organicness'

PR #484, opened 2025-09-04
Lock GPU clocks for running benchmarks

skallweitNV:


tools/ci.py
```diff
@@ -202,6 +213,14 @@ def benchmark_python(args: Any):
                 raise
             # Otherwise, continue with other devices
 
+    # Unlock GPU clocks
```
ccummingsNV: This should really be in some form of 'finally' block I'd say - or tests breaking will leave your clock locked right?

ccummingsNV (changes requested):
I'd strongly suggest we try and avoid leaving the machine in a 'locked clock' state in the event of test failing

tools/ci.py
skallweitNV: fixed

ccummingsNV (approved):

PR #485, opened 2025-09-04
Added caching of generated ref<SlangModule>s in SlangSession

tdavidovicNV:
When SlangSession sees a SlangModuleDesc request that it has already processed, it returns the already existing SlangModule, rather than asking `slang::ISession` for reprocessing it to a new `slang::IModule`


src/sgl/device/shader.h
```diff
@@ -336,22 +365,17 @@ class SGL_API SlangSession : public Object {
...
+    SessionModuleCache m_session_module_cache;
+    /// Maps module to an iterator in the cache for fast module deletion.
+    std::map<SlangModule*, SessionModuleCache::iterator> m_session_module_cache_reversed;
+
     void update_module_cache_and_dependencies();
     bool write_module_to_cache(slang::IModule* module);
     void create_session(SlangSessionBuild& build);
 };
 
-struct SlangModuleDesc {
```
skallweitNV, line 344: Would be good to keep the definition here, fits the context better. Use a fwd decl instead.

src/sgl/device/shader.h
```diff
@@ -311,7 +340,7 @@ class SGL_API SlangSession : public Object {
     // Internal functions to link programs+modules to their owning session
     void _register_program(ShaderProgram* program);
     void _unregister_program(ShaderProgram* program);
-    void _register_module(SlangModule* module);
+    void _register_module(SlangModule* module, const SlangModuleDesc& desc);
```
skallweitNV: Why do we need to pass the desc here, isn't it stored on the module?

src/sgl/device/shader.h
```diff
@@ -261,6 +263,33 @@ struct SlangSessionData : Object {
...
+    /// Required module name
+    std::string module_name;
+
+    /// Optional module source. If not specified slang module resolution is used.
+    std::optional<std::string> source;
+
+    /// If source specified, additional path for compilation.
+    std::optional<std::filesystem::path> path;
+
+    bool operator==(const SlangModuleDesc& rhs) const
```
skallweitNV: I'd prefer keeping the desc structs free from member functions in general. Do we need all of this?

src/sgl/device/shader.h
tdavidovicNV: I've tried. And on linux it won't compile, because the map in SlandSession cannot work on forward declared struct. (In MSVC it is fine)

src/sgl/device/shader.h
tdavidovicNV: Good point, can grab it from the pointer.

src/sgl/device/shader.h
tdavidovicNV: You can't use it as a key in a map without this. I've tried to put `std::hash` specialization at the end of the file first, but that doesn't work, because the compiler needs to first see the specialization, and only then its uses. 

If it first sees the use, e.g., in the map in `SlangSession`, it will already create a specialization for it, and the explicit specialization will be reported as duplicate. (I didn't know that one). The two alternatives are either use an explicit hasher, or close the `sgl` namespace after SlangModuleDesc, specialize the `std::hash`, and open the sgl namespace.

src/sgl/device/shader.h
skallweitNV: Is it worth adding something like `ModuleCacheKey` that is completely separate from `SlangModuleDesc`?

skallweitNV: something like
```c++
struct {
    std::string module_name;
    std::string source;
    std::string path; // normalized path
};
```

tdavidovicNV: I don't like the idea of creating deep copy of the module descriptor every time we need to do a lookup. What I could do is make the cache a separate class and just do all the dirty work in the cpp, at the cost of one extra indirection.

That way the hashing and comparison can all be done out of the view.

PR #486, opened 2025-09-04
Deleted copy from SlangModule and SlangProgram and added checks for hotreload registration

tdavidovicNV:
Disabled the ability of SlangModule and SlangProgram to be copied and thus break the registration in SlangSession.

Added asserts for SlangProgram and SlangModule registration and de-registration.

### DIFF
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -117,4 +117,4 @@ jobs:
       # Benchmark (Python)
       - name: Benchmark (Python)
         if: contains(matrix.flags, 'benchmark')
-        run: python tools/ci.py benchmark-python --run-id "${{ github.run_id }}" --mongodb-connection-string "${{ secrets.BENCHMARK_MONGODB_CONNECTION_STRING }}" --mongodb-database-name "nvr-ci"
+        run: python tools/ci.py benchmark-python --run-id "${{ github.run_id }}" --mongodb-connection-string "${{ secrets.BENCHMARK_MONGODB_CONNECTION_STRING }}" --mongodb-database-name "nvr-ci" --lock-gpu-clocks
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -183,24 +183,39 @@ def benchmark_python(args: Any):
         # Run for all device types plus nodevice tests
         device_types = device_types + ["nodevice"]
 
-    # Run benchmarks for each device type
-    for device_type in device_types:
-        print(f"Running benchmarks for device type: {device_type}")
+    try:
+        # Lock GPU clocks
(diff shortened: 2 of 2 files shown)


src/sgl/device/shader.cpp
```diff
@@ -602,26 +602,29 @@ std::string SlangSession::load_source(std::string_view module_name)
...
-        m_registered_modules.push_back(module);
+    SGL_ASSERT(
+        std::find(m_registered_modules.begin(), m_registered_modules.end(), module) == m_registered_modules.end()
+    );
+    m_registered_modules.push_back(module);
 }
 
 void SlangSession::_unregister_module(SlangModule* module)
 {
     auto existing = std::find(m_registered_modules.begin(), m_registered_modules.end(), module);
```
ccummingsNV, line 625: What with it being a python library, I've generally favoured not-crashing where reasonable. @skallweitNV you have a take on that? I'm not hugely fussy - I just find in practice if the library does something wrong, it's not much help to the user if it just dies.

ccummingsNV (commented):
I'm kind of uneasy on changing the error handling to things that just crash the python interpreter. If we want to make the checks detect errors, then shouldn't they be SGL_CHECKS that raise an exception that the python interpreter sees?

src/sgl/device/shader.cpp
tdavidovicNV: Well this is not in any way an error that user can cause. If this happens, world is on fire.. same as we're not actually checking in bounds access on `my_vector[7]`. It won't even be there in a python library most of the time.

src/sgl/device/shader.h
```diff
@@ -364,6 +364,12 @@ class SGL_API SlangModule : public Object {
     SlangModule(ref<SlangSession> session, const SlangModuleDesc& desc);
     ~SlangModule();
 
+    // Deleted because the hot reload mechanism in SlangSession assumes only objects created by it exist.
+    SlangModule(SlangModule&) = delete;
```
skallweitNV: in `macros.h` there is 
- `SGL_NON_COPYABLE(cls)`
- `SGL_NON_MOVABLE(cls)`
- `SGL_NON_COPYABLE_AND_MOVABLE(cls)`

so just putting `SGL_NON_COPYABLE_AND_MOVABLE(SlangModule)` should be enough here

src/sgl/device/shader.cpp
skallweitNV: well, I think the ASSERT is the right thing here because users shouldn't even be able to use the library in a way where this triggers

src/sgl/device/shader.h
tdavidovicNV: Updated for this

PR #488, opened 2025-09-05
Various fixes for mem leaks

ccummingsNV:
- Fixed use of 'type.name' in type lookups
- Fixed ability for buffer to cause internal lookups to be associated with the wrong cache
- Properly labelled a load of native objects
- Code to log heap status to text
- Add python GC visitors to some slangpy types

### DIFF
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -33,9 +33,9 @@ struct GcHelper<slangpy::NativeCallRuntimeOptions> {
     void traverse(slangpy::NativeCallRuntimeOptions*, GcVisitor& visitor)
     {
         visitor("uniforms");
-        visitor("this_ref");
+        visitor("_native_this");
     }
-    void clear(slangpy::NativeCallRuntimeOptions* opts) { opts->clear(); }
+    void clear(slangpy::NativeCallRuntimeOptions* opts) { opts->garbage_collect(); }
 };
 
 } // namespace sgl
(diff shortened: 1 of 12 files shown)


src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -1271,6 +1281,12 @@ SGL_PY_EXPORT(utils_slangpy)
             &NativeCallRuntimeOptions::set_uniforms,
             D_NA(NativeCallRuntimeOptions, uniforms)
         )
+        .def_prop_rw(
+            "this_ref",
+            &NativeCallRuntimeOptions::get_this,
+            &NativeCallRuntimeOptions::set_this,
+            D_NA(NativeCallRuntimeOptions, this_ref)
```
skallweitNV: minor inconsistency between the c++ and python names, is it worth changing the name to something that works for both?

src/slangpy_ext/utils/slangpy.h
```diff
@@ -598,6 +601,13 @@ class NativeCallRuntimeOptions : Object {
     /// Set the CUDA stream.
     void set_cuda_stream(NativeHandle cuda_stream) { m_cuda_stream = cuda_stream; }
 
+    /// Clear internal data for garbage collection
+    void clear()
```
skallweitNV: here it's called `clear` in the other case further down `garbage_collect`. should we choose a consistent name?

skallweitNV (dismissed):
minor comments

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: I'll take a look in the next PR - elsewhere I've named stuff '_native_bla' to indicate something is 'low level'. Could do _native_this.

src/slangpy_ext/utils/slangpy.h
ccummingsNV: Yeah - I was debating whether it should just be a general purpose clear, but I've commented it as specifically for garbage collection so...

PR #490, opened 2025-09-05
Leak testing

ccummingsNV:
- If live object tracking is enabled, uses it to detect tests that leak objects
- Adds a marker that says a test is known to leak, and what it leaks

### DIFF
--- a/.github/workflows/ci-benchmark.yml
+++ b/.github/workflows/ci-benchmark.yml
@@ -117,4 +117,4 @@ jobs:
       # Benchmark (Python)
       - name: Benchmark (Python)
         if: contains(matrix.flags, 'benchmark')
-        run: python tools/ci.py benchmark-python --run-id "${{ github.run_id }}" --mongodb-connection-string "${{ secrets.BENCHMARK_MONGODB_CONNECTION_STRING }}" --mongodb-database-name "nvr-ci"
+        run: python tools/ci.py benchmark-python --run-id "${{ github.run_id }}" --mongodb-connection-string "${{ secrets.BENCHMARK_MONGODB_CONNECTION_STRING }}" --mongodb-database-name "nvr-ci" --lock-gpu-clocks
--- a/slangpy/core/dispatchdata.py
+++ b/slangpy/core/dispatchdata.py
@@ -209,7 +209,7 @@ void {reflection.name}_entrypoint({params}) {{
         opts: "NativeCallRuntimeOptions",
         thread_count: uint3,
         vars: dict[str, Any] = {},
-        command_buffer: Optional[CommandEncoder] = None,
+        command_encoder: Optional[CommandEncoder] = None,
         **kwargs: dict[str, Any],
     ) -> None:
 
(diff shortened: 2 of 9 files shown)


src/sgl/core/object.h
```diff
@@ -143,7 +156,7 @@ class SGL_API Object {
 
 #if SGL_ENABLE_OBJECT_TRACKING
     /// Reports current set of live objects.
-    static void report_live_objects();
+    static std::vector<LiveObjectInfo> report_live_objects(bool log_to_tty = true);
```
skallweitNV, line 159: I don't like this API too much, mostly because we have similar calls in slang-rhi and D3D12. Let's leave `report_live_objects` as is (printing to stdout).
Can we add something like `get_live_object_infos()` to return the list?

pyproject.toml
```diff
@@ -72,3 +72,6 @@ reportInvalidTypeForm = "warning"
 
 [tool.pytest.ini_options]
 pythonpath = ["."]
+markers = [
```
skallweitNV, line 75: Can we move this definition to the slangpy pytest plugin somehow?

src/sgl/core/object.cpp
```diff
@@ -117,26 +117,37 @@ PyObject* Object::self_py() const noexcept
 
 #if SGL_ENABLE_OBJECT_TRACKING
 
-void Object::report_live_objects()
+std::string LiveObjectInfo::to_string()
+{
+    return fmt::format(
+        "address={}, self_py={}, ref_count={}, class_name={}",
```
skallweitNV, line 123: these `to_string` functions usually return something including the class name `LiveObjectInfo(...)`

slangpy/testing/helpers.py
```diff
@@ -82,6 +86,59 @@ def set_device_types(device_types_str: Optional[str]) -> None:
 USED_TORCH_DEVICES: bool = False
 METAL_PARAMETER_BLOCK_SUPPORT: Optional[bool] = None
 
+TRACKED_LIVE_OBJECTS: Optional[list[Any]] = None
+
+# Types to ignore when checking for leaked objects
+# - The reflection types are created and cached per device when buffers are loaded, so are hard
+#   to identify as actual leaks.
+# - CoopVec is created on demand within the device when the coopvec api is used, and so will appear
```
skallweitNV, line 94:
```suggestion
# - The reflection types are created and cached per device when buffers are loaded, so are hard
#   to identify as actual leaks.
# - CoopVec is created on demand within the device when the coopvec API is used, and so will appear
#   as a leak for cached devices.
IGNORE_LIVE_OBJECT_TYPES = ["NativeSlangType", "TypeLayoutReflection", "TypeReflection", "CoopVec"]

```

slangpy/testing/helpers.py
```diff
@@ -44,6 +45,9 @@
 else:
     raise RuntimeError("Unsupported platform")
 
+# If live object tracking is supported, enable leak tracking
+LEAK_TRACKING_ENABLED = hasattr(spy.Object, "report_live_objects")
```
skallweitNV, line 49: Should we move all the leak checking specifics into a new `leak_tracking.py` module?

PR #498, opened 2025-09-08
Use compute pipeline

skallweitNV:
Replaces use of `ComputeKernel` with `ComputePipeline`. This is in preparation of supporting raytracing pipelines in slangpy.


ccummingsNV (approved):
LGTM

PR #499, opened 2025-09-08
ci-latest-slang branch option

skallweitNV:
- Add a `slang_branch` option to `ci-latest-slang` workflow to allow testing against arbitrary version of slang.
- Remove pip cache which harms self-hosted runners (they have a persistent pip cache) as well as macos hosted runners, where the install from scratch is slightly faster
- Split "Build Slang" step into separate "Checkout Slang" and "Build Slang" steps

### DIFF
--- a/.github/workflows/ci-latest-slang.yml
+++ b/.github/workflows/ci-latest-slang.yml
@@ -65,7 +65,8 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
+          # Cache not needed on self-hosted runners
+          # cache: 'pip'
 
       # Setup Python environment.
       - name: Setup Python environment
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-latest-slang.yml
```diff
@@ -60,7 +65,8 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
+          # Cache not needed on self-hosted runners
```
ccummingsNV: If we run on MacOS too, do we need need 2 versions of this (with/without cache)? That's what the main ci does.

skallweitNV: Total python setup including env is around 30sec on macos runners. With the cache on previous runs it's slightly slower, so I would say that's fine.

ccummingsNV (approved):

PR #513, opened 2025-09-17
Fix for load from mem stream + add load from file / numpy functions

ccummingsNV:
- Added a fix to mem stream so it correctly handles read-beyond-end-of-file
- Added python functionality to load from a memory buffer
- Added clear load_from_file and load_from_numpy functions, to distringuish from Bitmap constructors that take a numpy array of actual pixels.

### DIFF
--- a/src/slangpy_ext/core/bitmap.cpp
+++ b/src/slangpy_ext/core/bitmap.cpp
@@ -148,6 +148,8 @@ SGL_PY_EXPORT(core_bitmap)
             "load_from_numpy",
             [](nb::ndarray<nb::device::cpu> data)
             {
+                if(!is_ndarray_contiguous(data))
+                    SGL_THROW("To load from a numpy array, ensure it is contiguous.");
                 MemoryStream stream(data.data(), data.nbytes());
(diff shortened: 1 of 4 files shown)


src/slangpy_ext/core/bitmap.cpp
```diff
@@ -138,6 +138,22 @@ SGL_PY_EXPORT(core_bitmap)
...
+            "load_from_file",
+            [](const std::filesystem::path& path) { return make_ref<Bitmap>(path); },
+            "path"_a,
+            D_NA(Bitmap, load_from_file)
+        )
+        .def_static(
+            "load_from_numpy",
+            [](nb::ndarray<nb::device::cpu> data)
+            {
+                MemoryStream stream(data.data(), data.nbytes());
```
tdavidovicNV, line 153: Wouldn't this happily accept strided numpy array that you'd read without the stride? I think you want a more strict conditions on the `nb::ndarray<here>` to make sure nanobind converts it to tightly strided c-continuous.

ccummingsNV:
[self-approve: minor fixes for PR]

PR #525, opened 2025-09-19
Add 2d matrix ops

ccummingsNV:
- Added some new matrix functions for 2d transforms
- Also made 3d scale/translate construction a bit more efficient

### DIFF
--- a/src/slangpy_ext/math/vector.cpp
+++ b/src/slangpy_ext/math/vector.cpp
@@ -59,7 +59,7 @@ void bind_vector_type(nb::module_& m, const char* name)
         vec.def_prop_ro("xy", &T::xy);
     }
     if constexpr (dimension == 4) {
-        vec.def_prop_ro("xyz", & T::xyz);
+        vec.def_prop_ro("xyz", &T::xyz);
     }
 
     vec.def("__len__", [](const T&) { return dimension; });


src/sgl/math/matrix_math.h
```diff
@@ -140,6 +149,21 @@ matrix<T, 4, 4> rotate(const matrix<T, 4, 4>& m, T angle, const vector<T, 3>& ax
...
+template<typename T>
+matrix<T, 3, 3> rotate_2d(const matrix<T, 3, 3>& m, T angle)
+{
+    T c = cos(angle);
+    T s = sin(angle);
+
+    matrix<T, 3, 3> result;
+    result.set_col(0, m.get_col(0) * c + m.get_col(1) * -s);
+    result.set_col(1, m.get_col(0) * s + m.get_col(1) * c);
+    result.set_col(2, m.get_col(2));
```
tdavidovicNV, line 162: I generally prefer the pattern where you create the matrix via `matrix_from_rotate_2d` and then you just multiply with it. That way, you have exactly one place where the matrix has to be checked, even thought it might be slightly less efficient (depends on how good the compiler is at eliminating multiply by 0)

tdavidovicNV (dismissed):

src/sgl/math/matrix_math.h
```diff
@@ -140,6 +149,21 @@ matrix<T, 4, 4> rotate(const matrix<T, 4, 4>& m, T angle, const vector<T, 3>& ax
...
+/// Apply a 2d rotation to a 3x3 matrix.
+template<typename T>
+matrix<T, 3, 3> rotate_2d(const matrix<T, 3, 3>& m, T angle)
+{
+    T c = cos(angle);
+    T s = sin(angle);
+
+    matrix<T, 3, 3> result;
+    result.set_col(0, m.get_col(0) * c + m.get_col(1) * -s);
+    result.set_col(1, m.get_col(0) * s + m.get_col(1) * c);
```
tdavidovicNV:
```
a b c    cs -sn  0     (a*cs + b*sn)  (a*-sn + b*cs)
d e f *  sn  cs  0 ==  (d*cs + e*sn)  (d*-sn + e*cs)
g h i    0   0   1     (g*cs + h*sn)  (g*-sn + h*cs)
```

I believe the multiplication code should be:
```
    result.set_col(0, m.get_col(0) * c + m.get_col(1) * s);
    result.set_col(1, m.get_col(0) * -s + m.get_col(1) * c);
```

And I'd definitely add a test that `rotate_2d` gives (about) the same result as matrix multiplication with a `make_from_rotate_2d` matrix

ccummingsNV:
[self-approve: CI is passing but there's corrupt git repo that needs sorting]

PR #531, opened 2025-09-27
Sort exr channels when writing via tinyexr

cek:
When writing images using tinyexr, sort channels in ABGR (alphabetical) order to maximize compatibility with external tools (e.g., tev, which displays images with swapped channels without this change).

The OpenEXR spec specifies that, "Channels are stored in alphabetical order, according to channel names," and thus many apps/utilities assume this is the case. However, TinyEXR does not enforce this by reordering channels.

### DIFF
--- a/src/sgl/core/bitmap.cpp
+++ b/src/sgl/core/bitmap.cpp
@@ -106,6 +106,7 @@ SGL_DIAGNOSTIC_POP
 #endif
 
 #include <algorithm>
+#include <numeric>
 #include <map>
 #include <mutex>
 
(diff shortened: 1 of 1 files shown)


cek-nv:
I have the (trivial) fix for the clang-format issue. I'm not sure what the protocol is here; please let me know if I should push the fix.

jhelferty-nv:
(Looks like `/format` doesn't work in slangpy repo, unlike in slang repo. So a manual fix is probably required.)

jhelferty-nv:
Thanks. As an aside, we usually use a git merge (via 'Update branch') instead of a rebase, it seems to keep github happier about tracking the changes since last review.

I'll re-approve, based on @tdavidovicNV's previous approval.

PR #545, opened 2025-10-03
Fixed texture_loader for CUDA

tdavidovicNV:
Fixed two issues that popped up with CUDA backend:
- TextureLoader assumed that if a format exists in the FORMAT_TABLE, it is automatically supported on the backend. This was not the case for `rgb32_float` which is not supported on CUDA. We now check if the device actually supports the format, on top of the format existing.
- Blit in `compute` mode requires the Texture to be `unordered_access` rather than `render_target`, so when mipmaps are being generated, the correct flag is now set.

### DIFF
--- a/src/sgl/utils/texture_loader.cpp
+++ b/src/sgl/utils/texture_loader.cpp
@@ -43,7 +43,8 @@ struct SourceImage {
  * \param options Texture loading options.
  * \return A pair containing the determined format and flag if the bitmap needs to be converted to RGBA to match the format.
  */
-inline std::pair<Format, bool> determine_texture_format(const Bitmap* bitmap, const TextureLoader::Options& options)
+inline std::pair<Format, bool>
+determine_texture_format(Device* device, const Bitmap* bitmap, const TextureLoader::Options& options)
 {
     SGL_ASSERT(bitmap != nullptr);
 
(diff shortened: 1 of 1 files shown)


src/sgl/utils/texture_loader.cpp
```diff
@@ -127,8 +128,12 @@ inline std::pair<Format, bool> determine_texture_format(const Bitmap* bitmap, co
     // Check if bitmap is RGB and we can convert to RGBA.
     bool convert_to_rgba = false;
     if (options.extend_alpha && pixel_format == PixelFormat::rgb) {
-        bool rgb_format_supported
-            = FORMAT_TABLE.find(make_key(PixelFormat::rgb, component_type, format_flags)) != FORMAT_TABLE.end();
+        // Find if the RGB format exists, if it does check if the device supports it
+        auto rgb_it = FORMAT_TABLE.find(make_key(PixelFormat::rgb, component_type, format_flags));
+        bool rgb_format_supported = rgb_it != FORMAT_TABLE.end();
+        if (rgb_format_supported && device->get_format_support(rgb_it->second) != FormatSupport::texture)
```
skallweitNV: you probably want `is_set(format_support, FormatSupport::texture)` here (mulitple flags could be set)

src/sgl/utils/texture_loader.cpp
```diff
@@ -127,8 +128,12 @@ inline std::pair<Format, bool> determine_texture_format(const Bitmap* bitmap, co
     // Check if bitmap is RGB and we can convert to RGBA.
     bool convert_to_rgba = false;
     if (options.extend_alpha && pixel_format == PixelFormat::rgb) {
-        bool rgb_format_supported
-            = FORMAT_TABLE.find(make_key(PixelFormat::rgb, component_type, format_flags)) != FORMAT_TABLE.end();
+        // Find if the RGB format exists, if it does check if the device supports it
+        auto rgb_it = FORMAT_TABLE.find(make_key(PixelFormat::rgb, component_type, format_flags));
```
skallweitNV: rewrite it like this?
```
bool rgb_format_supported = false;
if (auto it = FORMAT_TABLE.find(make_key(PixelFormat::rgb, component_type, format_flags);
    it != FORMAT_TABLE.end() && is_set(device->get_format_support(it->second), FormatSupport::texture)
    rbg_format_supporte = true;
```

skallweitNV (changes requested):
Needs to use `is_set` I think

src/sgl/utils/texture_loader.cpp
tdavidovicNV: I thought you didn't like these long ifs :-D

tdavidovicNV:
The requested change has been done

PR #546, opened 2025-10-06
Fix missing switch handling

tdavidovicNV:


src/sgl/core/data_type.h
```diff
@@ -63,7 +63,9 @@ inline size_t data_type_size(sgl::DataType type)
     case DataType::uint64:
     case DataType::float64:
         return 8;
-    }
+    default:
```
skallweitNV: in general it would be better to cover all cases and not use default? is it just the `void` type that should throw?

PR #553, opened 2025-10-07
We now pass raw Device* rather than ref to the do_async call to avoid…

tdavidovicNV:
… capture.


src/sgl/utils/texture_loader.cpp
```diff
@@ -394,7 +394,7 @@ ref<Texture> TextureLoader::load_texture(const Bitmap* bitmap, std::optional<Opt
 ref<Texture> TextureLoader::load_texture(const std::filesystem::path& path, std::optional<Options> options_)
 {
     Options options = options_.value_or(Options{});
-    SourceImage source_image = load_and_convert_source_image(m_device, path, options);
+    SourceImage source_image = load_and_convert_source_image(m_device.get(), path, options);
```
skallweitNV, line 397: this shouldn't be needed

tdavidovicNV: Yeah, but I figured I'd rather do the pattern same way even outside the `do_async`.

PR #555, opened 2025-10-07
LMDB cache

skallweitNV:
This adds a persistent cache based on LMDB. This will be used for the shader cache.
See comments in code for overview on how it works.


src/sgl/core/lmdb_cache.cpp
```diff
@@ -0,0 +1,324 @@
...
+    MDB_val key, val;
+    while (mdb_cursor_get(cursor, &key, &val, MDB_NEXT) == MDB_SUCCESS) {
+        entries.push_back({
+            .last_access = static_cast<const MetaData*>(val.mv_data)->last_access,
+            .key = key,
+        });
+    }
+
+    // Create heap based on last access time (oldest first).
+    auto cmp = [](const Entry& a, const Entry& b) { return a.last_access > b.last_access; };
```
ccummingsNV, line 297: Depending on cache size vs shader use this can be a very poor model, as it means if the cache is too small, the first shaders loaded are always evicted. If we want a model this simple it might be worth tracking whether the cache is too small to accomodate all entries within a single session.

ccummingsNV (approved):
Potential issue with eviction model if cache is too small. Otherwise LGTM.

src/sgl/core/lmdb_cache.cpp
skallweitNV: Not sure I follow, we will evict the oldest entries first. Sure, if the cache is way too small we will evict constantly (and we might also evict less than needed for the new entry). But with a reasonably sized cache, this should be fine, or am I missing something here?

PR #559, opened 2025-10-08
Make print test expect correct results on Metal and disable it for now

### LINKED ISSUE #497: `tests/devices/test_print.py` enforces incorrect behaviour for Metal
For whatever reason, https://github.com/shader-slang/slang/pull/8208 seems to also make Metal print half-precision floats correctly (I'm not exactly sure why, none of the changes should affect Metal). Anyway, the CI also runs slangpy tests, where `test_print.py` now fails due to it expecting Metal to print incorrect values:

https://github.com/shader-slang/slangpy/blob/0336e7a2e67a4283853d6f962f17dd7d54c91d3f/slangpy/tests/device/test_print.py#L118

All other platforms are expected to print this instead:

https://github.com/shader-slang/slangpy/blob/0336e7a2e67a4283853d6f962f17dd7d54c91d3f/slangpy/tests/device/test_print.py#L152

and the code doing the printing looks like this:

https://github.com/shader-slang/slangpy/blob/0336e7a2e67a4283853d6f962f17dd7d54c91d3f/slangpy/tests/device/test_print.slang#L138-L143

Not sure what the best course of action here is.

juliusikkala:
Closes #497. The problem with this test is explained in more detail in that issue, but in short, the Metal test has incorrect expectations. It was expecting that printf doesn't work with `float16_t` on Metal, and this causes the test to fail in a Slang compiler PR (https://github.com/shader-slang/slang/pull/8208) which appears to fix this bug as well.


slangpy/tests/device/test_print.py
```diff
@@ -6,8 +6,16 @@
 import slangpy as spy
 from slangpy.testing import helpers
 
+# Metal test disabled until printing of float16 values in Metal has been fixed
+# in Slang: https://github.com/shader-slang/slangpy/issues/497
+PRINT_DEVICE_TYPES = [
+    x
+    for x in helpers.DEFAULT_DEVICE_TYPES
+    if x in [spy.DeviceType.d3d12, spy.DeviceType.vulkan, spy.DeviceType.cuda]
+]
```
skallweitNV: Keep the comment but instead of defining `PRINT_DEVICE_TYPES` just skip the test, in `test_print` do:
```python
if device_type == spy.DeviceType.metal:
    pytest.skip("Skipped due to issue in Slang: https://github.com/shader-slang/slangpy/issues/497")
```

PR #561, opened 2025-10-09
Add persistent cache implementation based on LMDB cache

skallweitNV:
- Fix `LMDBCache` to reuse DB environments per process (LMDB only allows each process to open a specific DB once)
- Hook up RHI shader cache (shaders + pipelines) to an LMDB cache using `PersistentCache` wrapper
- Add test

### DIFF
--- a/src/sgl/core/lmdb_cache.cpp
+++ b/src/sgl/core/lmdb_cache.cpp
@@ -51,9 +51,11 @@ public:
     void commit()
     {
         SGL_CHECK(m_txn != nullptr, "Transaction is already committed or aborted");
-        if (int result = mdb_txn_commit(m_txn); result != MDB_SUCCESS)
-            LMDB_THROW("Failed to commit transaction", result);
+        // It's an error to call `mdb_txn_abort` after `mdb_txn_commit`, even if it fails.
+        MDB_txn* txn = m_txn;
         m_txn = nullptr;
+        if (int result = mdb_txn_commit(txn); result != MDB_SUCCESS)
+            LMDB_THROW("Failed to commit transaction", result);
     }
 
     operator MDB_txn*() { return m_txn; }


src/sgl/device/persistent_cache.cpp
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "persistent_cache.h"
+
+#include "sgl/core/error.h"
+#include "sgl/core/logger.h"
+
+namespace sgl {
+
+PersistentCache::PersistentCache(const std::filesystem::path& path, size_t max_size)
```
ccummingsNV, line 10: With most of our code we've passed in descriptor structures. Any reason not to stick with that here?

src/sgl/device/persistent_cache.cpp
```diff
@@ -0,0 +1,103 @@
...
+PersistentCacheStats PersistentCache::stats() const
+{
+    return {
+        .entry_count = m_cache->stats().entries,
+        .hit_count = m_hit_count.load(),
+        .miss_count = m_miss_count.load(),
+    };
+}
+
+SlangResult PersistentCache::queryInterface(const SlangUUID& uuid, void** outObject)
```
ccummingsNV, line 52: Is the none-standard naming convention because these are virtuals being overriden?

src/sgl/device/persistent_cache.cpp
skallweitNV: good point

skallweitNV: actually I leave this for now, we're not exposing this class to python.

src/sgl/core/lmdb_cache.cpp
```diff
@@ -321,4 +300,87 @@ void LMDBCache::evict()
     m_evictions.fetch_add(evictions);
 }
 
+// LMDB doesn't support opening the same DB environment multiple times in the same process.
+// To work around this, we keep a global list of open environments to reuse them if opened multiple times.
+
+struct DBCacheItem {
+    uint64_t ref_count;
+    ProcessID pid;
```
ccummingsNV, line 308: If this list of DBCacheItems is per process, why do they need to store a process id - won't they all end up with the same one?

skallweitNV: SGL is a shared library. It might be shared by more than one process.

PR #562, opened 2025-10-09
Allow calling interface parameters with implementing types

tdavidovicNV:


slangpy/bindings/boundvariable.py
```diff
@@ -264,6 +264,11 @@ def bind(
             self.slang_modifiers = modifiers.union(slang.modifiers)
         self.variable_name = self.name
 
+        # HACK! Part of the hack in callsignature.py specialize,
+        # where structs written to interface inputs need to be explicitly specialized.
```
ccummingsNV: Can you add 'explicitly specialized BEFORE binding'

slangpy/core/callsignature.py
```diff
@@ -154,6 +155,19 @@ def specialize(
...
                 else:
                     inputs.append(python_arg.python)
+
+            elif (
+                slang_param.type.type_reflection.kind == TypeReflection.Kind.interface
+                and isinstance(python_arg.python, StructMarshall)
+                and python_arg.python.slang_type.name != "Unknown"
+                and not python_arg.explicitly_vectorized
+            ):
+                # HACK! If we're calling a function with an interface parameter,
```
ccummingsNV, line 165: Can you note the other half of the hack is inside bind()

src/sgl/core/platform.h
```diff
@@ -13,11 +13,11 @@
 #include <vector>
 
 #if SGL_WINDOWS
-#define DECLARE_HANDLE(name)                                                                                           \
+#define SGL_DECLARE_HANDLE(name)                                                                                       \
```
ccummingsNV, line 16: Is this deliberately part of same checkin? no prob if it is - just checking :)

tdavidovicNV: It is not, that should have gone in before

tdavidovicNV: (It did go in before, I just didn't repoint slangpy)

slangpy/bindings/boundvariable.py
tdavidovicNV: Done

slangpy/core/callsignature.py
tdavidovicNV: Done

PR #564, opened 2025-10-10
Replace BS::thread_pool with nanothread

skallweitNV:


src/sgl/core/thread.h
```diff
@@ -14,15 +14,244 @@ namespace sgl::thread {
 SGL_API void static_init();
 SGL_API void static_shutdown();
 
-/// Block until all scheduled tasks are completed.
+/// Block until all register tasks are completed.
```
tdavidovicNV: registered

src/sgl/core/thread.cpp
```diff
@@ -4,30 +4,33 @@
...
+    }
 }
 
-BS::thread_pool& global_thread_pool()
+void register_task(Task* task)
 {
-    SGL_CHECK(s_global_thread_pool, "Global thread pool not initialized!");
-    return *s_global_thread_pool;
+    std::lock_guard lock(s_tasks_mutex);
+    s_tasks.push_back(task);
```
tdavidovicNV: This concept makes me rather nervous, because it can lead to deadlocks if people call the `wait_for_tasks` from a thread. Do we need that functionality? Can we pretend it doesn't exist until we need it?

(I think this was part of the original TBB interface and quickly replaced with task_group and waiting on those, because otherwise it was just dangerous)

skallweitNV: It shouldn't, `task_wait` will pick up pending tasks while it waits.

tdavidovicNV: A typical example is something like: Task A spawns Task B and waits for it. Task B inside spawns task C and waits for all. Deadlock, because Task A waits for B which waits for (among others) A.

tdavidovicNV (approved):
LGTM

PR #566, opened 2025-10-11
Implemented printing of strings

tdavidovicNV:
This should allow printing of `Strings`, but fails on Slang compilation.

A work-around would be adding a custom sgl::string that explicitly only stores the hash int, and can be used in place of String.

### DIFF
--- a/tests/sgl/device/test_shader.cpp
+++ b/tests/sgl/device/test_shader.cpp
@@ -85,7 +85,7 @@ TEST_CASE_GPU("shader")
         CHECK_EQ(path0.filename(), "_testshader_simple.slang");
     }
 
-    // Load a module with a 2-stage dependency chain and verify all 4 dependencies.
+    // Load a module with a 2-stage dependency chain and verify all 3 dependencies.
     SUBCASE("multi_module_dependency")
     {
         ref<SlangModule> module = ctx.device->load_module((dir / "_testshader_dependent.slang").string());


slangpy/tests/device/test_print.slang
```diff
@@ -1,8 +1,38 @@
...
+    StringHash string_from_interface_method();
+}
+
+struct Struct3 : Interface
+{
+    __init() { }
+
+    static const StringHash string_from_interface_static_const = "string_from_interface_static_const";
+    StringHash string_from_interface_method() { return "string_from_interface_method"; }
+}
```
skallweitNV: This is not really testing the string printing but various potential issues in the slang compiler related to structs/interfaces etc. I'd prefer the print test to just test printing ...

src/sgl/device/print.cpp
```diff
@@ -28,6 +28,7 @@ namespace print_buffer {
     /// This needs to be in sync with the enum in print.slang.
     enum class Type {
         boolean,
+        string,
```
skallweitNV: why not append that at the end of the existing list, just seems very arbitrary to put between boolean and int8 :)

src/sgl/device/print.cpp
```diff
@@ -149,6 +172,9 @@ namespace print_buffer {
         case Type::boolean:
             arg_store.push_back(decode_value<bool>(data.subspan(4), layout));
             break;
+        case Type::string:
+            arg_store.push_back(decode_value<std::string>(data.subspan(4), layout, &hashed_strings));
```
skallweitNV: shouldn't we pass the hash instead of string containing the hash bytes?

src/sgl/device/string_hash.slang
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Helper class to wrap Slang's String class, allowing it to be passed and returned,
+// as the `String` itself is supposed to be only really a literal passed directly into
+// a function call.
+struct StringHash
```
skallweitNV: can we just put this into `print.slang`? to me this looks like a bit of a workaround only necessary for printing?

src/sgl/device/print.cpp
tdavidovicNV: This is the place where you're already preparing values for the fmt::format. At this stage, I think we want the native C++ types put in the `Value<T>` which then gets passed to the appropriate `fmt::formatter<T>`.

slangpy/tests/device/test_print.py
```diff
@@ -37,6 +37,7 @@ def test_print(device_type: spy.DeviceType):
 0123456
 01234567
 bool: false=false, true=true
+pritable_string: foo=foo
```
skallweitNV: typo: `pritable_string` -> `printable_string`

src/sgl/device/print.cpp
skallweitNV: absolutely, i didn't have enough context when doing the review

tests/sgl/device/test_shader.cpp
```diff
@@ -85,11 +85,11 @@ TEST_CASE_GPU("shader")
         CHECK_EQ(path0.filename(), "_testshader_simple.slang");
     }
 
-    // Load a module with a 2-stage dependency chain and verify all 3 dependencies.
+    // Load a module with a 2-stage dependency chain and verify all 4 dependencies.
```
skallweitNV: this is wrong now

src/sgl/device/print.cpp
```diff
@@ -123,17 +128,35 @@ namespace print_buffer {
 
         SGL_ASSERT(value.element_count <= Value<T>::MAX_ELEMENT_COUNT);
 
-        // Elements are aligned to 4 bytes.
-        uint32_t element_size = ((uint32_t(sizeof(T)) + 3) / 4) * 4;
-        SGL_ASSERT(data.size() >= element_size * value.element_count);
+        if constexpr (std::is_same_v<T, std::string>) {
```
skallweitNV: you hinted at it, but `string_view` would probably be a lot cheaper here. also the hash -> string map should be constant, so this should just work.

PR #568, opened 2025-10-12
Feature/macos file dialogs

klukaszek:
I was working on a little SlangPy tool for an assignment, and noticed that the file dialog API was marked as unimplemented for macOS, but I needed it for loading different files via UI. So I figured I should just add it!

Here's a list of what has been changed.

  - Add macOS implementations for open_file_dialog,
    save_file_dialog, and choose_folder_dialog using
    NSOpenPanel/NSSavePanel with UTType filters.
  - Compile macOS platform code as Objective‑C++ (.mm) and
    enable ARC.
  - Link required macOS frameworks: AppKit, Foundation,
    UniformTypeIdentifiers.

---
    
I originally included a change to the CMakelist file that changed the minimum macOS version, but I reverted it and figured I should just leave a note.

- `libslang.dylib` seems to target macOS15.0, but SlangPy targets macOS13.0.
```
Linking CXX shared library Debug/libsgl.dylib
ld: warning: building for macOS-13.0, but linking with dylib '@rpath/libslang.dylib' which was built for newer version 15.0 
```

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -195,18 +195,6 @@ else()
     set(SGL_HAS_AGILITY_SDK OFF PARENT_SCOPE)
 endif()
 
-if(TARGET slang-rhi-optix)
-    get_target_property(OPTIX_DIR slang-rhi-optix INTERFACE_INCLUDE_DIRECTORIES)
-    sgl_copy_binary(${OPTIX_DIR}/optix.h optix)
-    sgl_copy_binary(${OPTIX_DIR}/optix_device.h optix)
-    sgl_copy_binary(${OPTIX_DIR}/optix_types.h optix)
-    sgl_copy_binary(${OPTIX_DIR}/internal/optix_device_impl.h optix/internal)
-    sgl_copy_binary(${OPTIX_DIR}/internal/optix_device_impl_transformations.h optix/internal)
-    set(SGL_HAS_OPTIX ON PARENT_SCOPE)
-else()
-    set(SGL_HAS_OPTIX OFF PARENT_SCOPE)
-endif()
-
 # imgui
 
 set(IMGUI_VERSION "1.90.1")
--- a/slangpy/types/buffer.py
+++ b/slangpy/types/buffer.py
@@ -357,7 +357,7 @@ class NDBuffer(NativeNDBuffer):
             device,
             dtype=dtype,
             shape=ndarray.shape,
-            usage=BufferUsage.shader_resource | BufferUsage.unordered_access,
+            usage=usage,
             memory_type=memory_type,
         )
         res.copy_from_numpy(ndarray)
--- a/src/sgl/device/device.cpp
+++ b/src/sgl/device/device.cpp
@@ -232,6 +232,7 @@ Device::Device(const DeviceDesc& desc)
     m_info.adapter_name = rhi_device_info.adapterName;
     m_info.adapter_luid = from_rhi(rhi_device_info.adapterLUID);
     m_info.timestamp_frequency = rhi_device_info.timestampFrequency;
+    m_info.optix_version = rhi_device_info.optixVersion;
     m_info.limits.max_texture_dimension_1d = rhi_device_info.limits.maxTextureDimension1D;
     m_info.limits.max_texture_dimension_2d = rhi_device_info.limits.maxTextureDimension2D;
     m_info.limits.max_texture_dimension_3d = rhi_device_info.limits.maxTextureDimension3D;
--- a/src/sgl/device/device.h
+++ b/src/sgl/device/device.h
@@ -178,6 +178,9 @@ struct DeviceInfo {
     /// The frequency of the timestamp counter.
     /// To resolve a timestamp to seconds, divide by this value.
     uint64_t timestamp_frequency;
+    /// The version of OptiX used by the device (0 if OptiX is not supported).
+    /// The format matches the OPTIX_VERSION macro, e.g. 90000 for version 9.0.0.
+    uint32_t optix_version;
     /// Limits of the device.
     DeviceLimits limits;
 };
--- a/src/sgl/device/shader.cpp
+++ b/src/sgl/device/shader.cpp
@@ -312,8 +312,12 @@ void SlangSession::create_session(SlangSessionBuild& build)
 
     // Set downstream argument for optix include path.
     if (device_type == DeviceType::cuda) {
-        auto optix_path = platform::runtime_directory() / "optix";
-        session_options.add(slang::CompilerOptionName::DownstreamArgs, "nvrtc", "-I" + optix_path.string());
+        uint32_t optix_version = m_device->info().optix_version;
+        if (optix_version > 0) {
+            std::string version_tag = fmt::format("{}_{}", optix_version / 10000, (optix_version % 10000) / 100);
+            auto optix_path = platform::runtime_directory() / "optix" / version_tag;
+            session_options.add(slang::CompilerOptionName::DownstreamArgs, "nvrtc", "-I" + optix_path.string());
+        }
     }
 
     // Set intermediate dump options.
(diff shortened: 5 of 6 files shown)


slangpy/__init__.py
```diff
@@ -1,21 +1,33 @@
...
-        # The DLLs are in the build directory.
         build_dir = open(os.path.join(package_dir, ".build_dir")).readline().strip()
         os.add_dll_directory(build_dir)
     else:
         print("Cannot locate sgl.dll.")
         sys.exit(1)
-
-del os, sys
+elif sys.platform == "darwin":
+    # On macOS during development, pre-load dependent dylibs from the build dir
```
skallweitNV: Is this actually needed? AFAICT development builds (editable pip install) did work for me.

src/sgl/core/platform_macos.mm
```diff
@@ -23,6 +23,14 @@
 #include <iostream>
 #include <fstream>
 
+#import <AppKit/AppKit.h>
+#import <Foundation/Foundation.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+#import <dispatch/dispatch.h>
+#import <Availability.h>
+
+// No hard minimum enforced here; rely on CMake's CMAKE_OSX_DEPLOYMENT_TARGET
```
skallweitNV: I'd remove this comment

src/sgl/core/platform_macos.mm
```diff
@@ -86,21 +94,157 @@ void set_keyboard_interrupt_handler(std::function<void()> handler)
 // File dialogs
 // -------------------------------------------------------------------------------------------------
 
+static std::vector<std::string> parse_extensions_from_filters(std::span<const FileDialogFilter> filters)
```
skallweitNV, line 106: A comment describing how the pattern needs to be transformed would be welcome.

src/sgl/core/platform_macos.mm
```diff
@@ -86,21 +94,157 @@ void set_keyboard_interrupt_handler(std::function<void()> handler)
...
+{
+    std::vector<std::string> exts;
+    for (const auto& f : filters) {
+        std::string pattern = f.pattern;
+        for (char& c : pattern)
+            if (c == ';')
+                c = ',';
```
skallweitNV: Is this replacement done for any reason? Looks like the `,` never ends up in the extensions, so we could also just search for `;` below instead?!

src/sgl/CMakeLists.txt
```diff
@@ -374,6 +380,11 @@ target_link_libraries(sgl
         # $<$<PLATFORM_ID:Linux>:gtk3>
 )
 
+# Link macOS frameworks when building on macOS
+if(SGL_MACOS)
+    target_link_libraries(sgl PRIVATE "-framework AppKit" "-framework Foundation" "-framework UniformTypeIdentifiers")
```
skallweitNV: Can we add these with a generate expression similar to it's done for windows above?

slangpy/__init__.py
klukaszek: Yeah entirely just a skill issue on my part. I didn't realize that the `resources/setpath.sh.in` didn't set DYLD_LIBRARY_PATH and I wasn't activating my venv for my other project (assumed that setpath worked), so nothing would load on my end.

Resolved in upcoming commit.

src/sgl/CMakeLists.txt
klukaszek: I was having trouble getting the generator to work with `-framework *`, as it produced `-framework -l*` which is incorrect.

I've updated the solution to instead find the libraries if the user is on macOS, stash them as CMake symbols, and then have them linked like the other dependencies.

src/sgl/core/platform_macos.mm
klukaszek: The filter logic was mostly just grafted over from an old project that explicitly requested comma separation and I didn't think too hard about it. Resolved!

src/sgl/core/platform_macos.mm
klukaszek: Removed

slangpy/__init__.py
```diff
@@ -1,20 +1,16 @@
 import os, sys
 
+package_dir = os.path.normpath(os.path.dirname(__file__))
```
skallweitNV: All these changes in this file should be reverted.

src/sgl/core/platform_macos.mm
```diff
@@ -86,21 +92,166 @@ void set_keyboard_interrupt_handler(std::function<void()> handler)
...
+
+   - Split on ';' (Windows-style multi pattern separator)
+   - Trim whitespace
+   - Strip leading '*' and '.'
+   - Lowercase result
+   - Collect unique extensions (no dot)
+*/
+static std::vector<std::string> parse_extensions_from_filters(std::span<const FileDialogFilter> filters)
+{
+    std::vector<std::string> exts;
```
skallweitNV, line 108: I feel like this could be written a bit more readable using the string utils in `sgl/core/string.h`. It might be slightly less efficient, but readability matters :)
Maybe something like this:
```c++
std::vector<std::string> exts;
for (const auto& filter : filters) {
    for (const auto& pattern : string::split(filter.pattern, ";")) {
        std::string ext = string::to_lower(string::remove_leading_whitespace(pattern, " \n\r\t.*"));
        if (exts.find(ext) == exts.end())
            exts.push_back(ext);
    }
}
return exts;
```

src/sgl/CMakeLists.txt
```diff
@@ -184,6 +183,22 @@ target_sources(sgl PRIVATE
...
+    set_source_files_properties(core/platform_macos.mm PROPERTIES COMPILE_FLAGS "-fobjc-arc")
+    find_library(SGL_APPKIT_FRAMEWORK AppKit REQUIRED)
+    find_library(SGL_FOUNDATION_FRAMEWORK Foundation REQUIRED)
+    find_library(SGL_UTI_FRAMEWORK UniformTypeIdentifiers REQUIRED)
+    add_library(sgl_appkit INTERFACE)
+    target_link_libraries(sgl_appkit INTERFACE "${SGL_APPKIT_FRAMEWORK}")
+    add_library(sgl_foundation INTERFACE)
+    target_link_libraries(sgl_foundation INTERFACE "${SGL_FOUNDATION_FRAMEWORK}")
+    add_library(sgl_uti INTERFACE)
+    target_link_libraries(sgl_uti INTERFACE "${SGL_UTI_FRAMEWORK}")
```
skallweitNV: Maybe we could simplify this into:
```cmake
add_library(sgl_macos_deps INTERFACE)
target_link_libraries(sgl_macos_deps INTERFACE ${SGL_APPKIT_FRAMEWORK} ${SGL_FOUNDATION_FRAMEWORK} ${SGL_UTI_FRAMEWORK})
```

src/sgl/CMakeLists.txt
```diff
@@ -372,8 +387,13 @@ target_link_libraries(sgl
         # $<$<PLATFORM_ID:Windows>:setupapi.lib>  # Used in MonitorInfo
         # Linux system libraries.
         # $<$<PLATFORM_ID:Linux>:gtk3>
+        # macOS system libraries.
+        $<$<PLATFORM_ID:Darwin>:sgl_appkit>
+        $<$<PLATFORM_ID:Darwin>:sgl_foundation>
+        $<$<PLATFORM_ID:Darwin>:sgl_uti>
```
skallweitNV: Then this becomes
```cmake
$<$<PLATFORM_ID:Darwin>:sgl_macos_deps>
```

src/sgl/CMakeLists.txt
```diff
@@ -372,8 +387,13 @@ target_link_libraries(sgl
...
 )
 
+
```
skallweitNV: stray newline

skallweitNV (changes requested):

slangpy/__init__.py
klukaszek: Reverted entirely.

src/sgl/core/platform_macos.mm
klukaszek: Applied your solution

PR #576, opened 2025-10-15
Fix build issue with VC ToolSets version 14.4X

jkwak-work:
This is another attempt to get the PR #315 working.

Here is what is different from PR #315.
The MSVC Toolset version for VS2022 is v14.3.
However MS used up the minor version number from 14.31.XXX to 14.39.XXX and started using 14.40.XXX.

Although the first three digits of the version numbers are 144, 14.4X is still considered v14.3 toolset.
The current CMake setup needs to recognize this correctly.

You can find more official details and explanations from the following URL:
https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/
> The C++ Project System (MS Build) is being updated to support '14.4x' MSVC toolsets under the v143 Platform Toolset.

### DIFF
--- a/docs/src/developer_guide/compiling.rst
+++ b/docs/src/developer_guide/compiling.rst
@@ -252,7 +252,7 @@ The following table lists the available configuration options:
       - Enable header validation
     * - ``SGL_MSVC_TOOLSET_VERSION``
       - ``Unset``
-      - Specify an exact MSVC toolset version to use (e.g., `14.38.33130`)
+      - Specify an exact MSVC toolset version to use (e.g., `14.44.35207`)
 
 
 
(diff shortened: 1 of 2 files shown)


jkwak-work:
Thanks for testing it out.

Let me do a little more research on that.

The part I am not clear now is how vcpkg works. It might be a matter of upgrading vcpkg on my local machine or slangpy is not setup to use the latest?

jkwak-work:
Changing to a draft because I don't think I got to the root of the cause.

If I remember correctly, vcpkg was using 14.44 version and cmake was using 14.38 version.
At that point, I thought that the problem was on vcpkg, because it looked like it was using v144, which turned out to be just v143.

I think the problem is on cmake not recognizing 14.44.XXX as v143.
It could be a bug in cmake itself or it could be a big in CMake scripts we wrote.

I need to investigate more.

jkwak-work:
Closing because I can workaround with vs2026 and cmake 4.2.

PR #587, opened 2025-10-17
Adds ability to control per-thread printing.

tdavidovicNV:
This adds a per-thread control of whether print is enabled or not. By default is is on. Can be turned off for any particular thread, allowing easier debugging prints across many modules.


skallweitNV (approved):
LGTM

ccummingsNV (approved):

PR #588, opened 2025-10-18
Added handling of ya bitmaps (found in vMaterials) by extending to rgba

tdavidovicNV:


slangpy/tests/utils/test_texture_loader.py
```diff
@@ -82,6 +82,8 @@ class FormatEntry:
     FormatEntry(PixelFormat.rgb, ComponentType.uint8, Format.rgba8_uint, Flags.extend_alpha),
     FormatEntry(PixelFormat.rgb, ComponentType.uint8, Format.rgba8_unorm, Flags.load_as_normalized | Flags.extend_alpha),
     FormatEntry(PixelFormat.rgb, ComponentType.uint8, Format.rgba8_unorm_srgb, Flags.load_as_srgb | Flags.extend_alpha),
+    # ya handling
+    FormatEntry(PixelFormat.ya, ComponentType.int8, Format.rgba8_sint, Flags.none),
```
skallweitNV, line 86: It's not obvious to me why we would want to convert these to an "rgba" format when there are "rg" formats. Sure, the alpha channel will end up in the right place ("a" vs "b" channel), but on the other hand we're wasting 2 channels with redundant information.

tdavidovicNV: Because if you're loading a Bitmap, you expect color to end up in rgb, and alpha to end up in alpha. If you load `ya` into `rg`, then your code needs to know that that specific texture actually has `alpha` in `g`. I see this as two different modes. If I point image loader at an image, I expect to get an image with channels where you'd expect them. Kinda same as if you just go and open the `png` in paint.net or something.

I just don't think getting a red-green image instead of greyscale+alpha just because someone saves png as greyscale instead of rgb is a good default behavior.

But I agree that this might need finer control (to make the same code reusable for both cases) and we could probably make this part of the TextureManager rather than of the loader.

PR #589, opened 2025-10-20
clang-format disallow lambdas on single line

skallweitNV:
This PR sets clang-format's `AllowShortLambdasOnASingleLine` to false.

I think the single line lambdas are more harmful than useful. First, they make setting breakpoints inside lambdas hard. Also, generally the formatting can get pretty bad. Disabling `AllowShortLambdasOnASingleLine` makes things more verbose and there are still cases where I'd prefer the short form, but overall it looks like the better trade-off to disable them.

### DIFF
--- a/examples/pathtracer/pathtracer.cpp
+++ b/examples/pathtracer/pathtracer.cpp
@@ -777,9 +777,24 @@ struct App {
             .vsync = false,
         });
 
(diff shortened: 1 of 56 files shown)


ccummingsNV (dismissed):
I'm fine with this personally - I find lambdas one of the harder bits of c++ to parse, so putting them on multiple lines is probably helpful.

PR #603, opened 2025-10-29
Adjusted CHECK_OP macros to print both args and their values

tdavidovicNV:
Previously, the `CHECK_OP` macros made a distinction before the left-hand-side and right-hand-side argument. Upon failure, it would print the string for the left hand side (but not value), and value for the right hand side (but not the string).

For example:
```
SGL_CHECK_EQ(subresource_data.size, subresource_layout.size_in_bytes);
```
would fail like:
```
"subresource_data.size" must be equal to 16356.
```

But the message would not report what the actual (incorrect value is).

This PR makes the reported string closely match doctest and reports:
```
SGL_CHECK_EQ(subresource_data.size, subresource_layout.size_in_bytes) is NOT correct! Values: SGL_CHECK_EQ(16355, 16356)
```
printing both the arguments, and their values.

### DIFF
--- a/src/sgl/core/error.h
+++ b/src/sgl/core/error.h
@@ -103,14 +103,29 @@ namespace detail {
 /// Helper for throwing an exception if a pointer is null.
 #define SGL_CHECK_NOT_NULL(arg) SGL_CHECK(arg != nullptr, "\"{}\" must not be null", #arg)
 
-#define SGL_CHECK_EQ(arg, value) SGL_CHECK(arg == value, "\"{}\" must be equal to {}", #arg, value)
-#define SGL_CHECK_NE(arg, value) SGL_CHECK(arg != value, "\"{}\" must not be equal to {}", #arg, value)
-#define SGL_CHECK_LT(arg, value) SGL_CHECK(arg < value, "\"{}\" must be less than {}", #arg, value)
-#define SGL_CHECK_LE(arg, value) SGL_CHECK(arg <= value, "\"{}\" must be less than or equal {}", #arg, value)
-#define SGL_CHECK_GT(arg, value) SGL_CHECK(arg > value, "\"{}\" must be greater than {}", #arg, value)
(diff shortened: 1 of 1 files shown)


src/sgl/core/error.h
```diff
@@ -103,14 +103,29 @@ namespace detail {
...
 #define SGL_CHECK_NOT_NULL(arg) SGL_CHECK(arg != nullptr, "\"{}\" must not be null", #arg)
 
-#define SGL_CHECK_EQ(arg, value) SGL_CHECK(arg == value, "\"{}\" must be equal to {}", #arg, value)
-#define SGL_CHECK_NE(arg, value) SGL_CHECK(arg != value, "\"{}\" must not be equal to {}", #arg, value)
-#define SGL_CHECK_LT(arg, value) SGL_CHECK(arg < value, "\"{}\" must be less than {}", #arg, value)
-#define SGL_CHECK_LE(arg, value) SGL_CHECK(arg <= value, "\"{}\" must be less than or equal {}", #arg, value)
-#define SGL_CHECK_GT(arg, value) SGL_CHECK(arg > value, "\"{}\" must be greater than {}", #arg, value)
-#define SGL_CHECK_GE(arg, value) SGL_CHECK(arg >= value, "\"{}\" must be greater than or equal {}", #arg, value)
+#define SGL_CHECK_EQ(lhs, rhs)                                                                                         \
+    SGL_CHECK(lhs == rhs, "SGL_CHECK_EQ({}, {}) is NOT correct! Values: SGL_CHECK_EQ({}, {})", #lhs, #rhs, lhs, rhs)
```
skallweitNV, line 107: I would prefer not to have `SGL_CHECK_EQ` in the output. This is mostly for user facing errors, so a more readable message without implementation details would be better IMO.

tdavidovicNV:
Closing this as the original macro is intended for very user facing argument checking, rather than reporting on the more subtle code errors. (If you need to print the rhs expression, you probably need to explain to the user more than "this failed")

PR #606, opened 2025-10-31
Update slangpy for library rename and versioning

jhelferty-nv:
For shader-slang/slang#8828

Handle new filename patterns introduced by shader-slang/slang#4016

Python wheels don't handle symlinks well (see `PEP 778`), so on macOS and Linux we need to resolve the `libslang-compiler` symlink to its innermost filename before linking. We also need this filename so it can be added to the list of files included in the wheel.

In order to keep the wheel size to a minimum, we only copy in files that are actually used; this means explicitly finding the runtime modules (`slang-glsl-module` and `slang-glslang`), which after the `slang-compiler` rename will have a version number in their filename. Since these are runtime modules, they don't have symlinks pointing to them. Instead, we discover that version number from the `slang-tag-version.h` header (in the same way `libslang-compiler` does internally).

Removes slang-rt dependency, which was not used by slangpy.

### DIFF
--- a/src/sgl/device/raytracing.cpp
+++ b/src/sgl/device/raytracing.cpp
@@ -66,6 +66,47 @@ AccelerationStructureBuildDescConverter::AccelerationStructureBuildDescConverter
                 rhi_build_input.proceduralPrimitives.aabbBuffers[i]
                     = detail::to_rhi(proceduralPrimitives->aabb_buffers[i]);
             rhi_build_inputs.push_back(rhi_build_input);
+        } else if (auto* spheres = std::get_if<AccelerationStructureBuildInputSpheres>(&input)) {
+            rhi::AccelerationStructureBuildInput rhi_build_input{
+                .type = rhi::AccelerationStructureBuildInputType::Spheres,
+                .spheres{
+                    .vertexBufferCount = narrow_cast<uint32_t>(spheres->vertex_position_buffers.size()),
+                    .vertexCount = spheres->vertex_count,
+                    .vertexPositionFormat = static_cast<rhi::Format>(spheres->vertex_position_format),
+                    .vertexPositionStride = spheres->vertex_position_stride,
+                    .vertexRadiusFormat = static_cast<rhi::Format>(spheres->vertex_radius_format),
+                    .vertexRadiusStride = spheres->vertex_radius_stride,
+                    .indexBuffer = detail::to_rhi(spheres->index_buffer),
+                    .indexFormat = static_cast<rhi::IndexFormat>(spheres->index_format),
+                    .indexCount = spheres->index_count,
+                    .flags = static_cast<rhi::AccelerationStructureGeometryFlags>(spheres->flags),
+                },
+            };
+            for (size_t i = 0; i < spheres->vertex_position_buffers.size(); ++i)
+                rhi_build_input.spheres.vertexPositionBuffers[i] = detail::to_rhi(spheres->vertex_position_buffers[i]);
+            for (size_t i = 0; i < spheres->vertex_radius_buffers.size(); ++i)
+                rhi_build_input.spheres.vertexRadiusBuffers[i] = detail::to_rhi(spheres->vertex_radius_buffers[i]);
+            rhi_build_inputs.push_back(rhi_build_input);
+        } else if (auto* linearSweptSpheres = std::get_if<AccelerationStructureBuildInputLinearSweptSpheres>(&input)) {
+            rhi::AccelerationStructureBuildInput rhi_build_input{
+                .type = rhi::AccelerationStructureBuildInputType::LinearSweptSpheres,
+                .linearSweptSpheres{
+                    .vertexBufferCount = narrow_cast<uint32_t>(linearSweptSpheres->vertex_position_buffers.size()),
+                    .vertexCount = linearSweptSpheres->vertex_count,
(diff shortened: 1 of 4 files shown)


cmake/utils.cmake
```diff
@@ -5,3 +5,18 @@ macro(ternary var boolean value1 value2)
         set(${var} ${value2})
     endif()
 endmacro()
+
+# Function to extract #define values from C/C++ headers
+function(extract_define_from_header header_file define_name output_var)
+    file(STRINGS "${header_file}" define_line REGEX "#define ${define_name}")
```
skallweitNV: you might want to handle whitespace between `#define` and the name like this as well:
```suggestion
    file(STRINGS "${header_file}" define_line REGEX "#define[ \t]+${define_name}")
```

cmake/utils.cmake
```diff
@@ -5,3 +5,18 @@ macro(ternary var boolean value1 value2)
...
+
+    # Handle string defines: #define FOO "value"
+    if(define_line MATCHES "\"([^\"]*)\"")
```
skallweitNV, line 19: this might incorrectly match something like `#define FOO 123 // this is some "bar" stuff` no?

external/CMakeLists.txt
```diff
@@ -108,58 +108,86 @@ set(SLANG_INCLUDE_DIR ${slang_SOURCE_DIR}/include)
...
+    if(EXISTS "${SLANG_TAG_VERSION_HEADER}")
+        extract_define_from_header("${SLANG_TAG_VERSION_HEADER}" "SLANG_VERSION_NUMERIC" SLANG_VERSION_FROM_HEADER)
+        if(SLANG_VERSION_FROM_HEADER)
+            set(SLANG_VERSION "${SLANG_VERSION_FROM_HEADER}")
+            message(STATUS "Using Slang version from local build: ${SLANG_VERSION}")
+        else()
+            message(WARNING "Could not extract SLANG_VERSION_NUMERIC from ${SLANG_TAG_VERSION_HEADER}, using default version: ${SLANG_VERSION}")
+        endif()
+    else()
+        message(WARNING "Slang tag version header not found: ${SLANG_TAG_VERSION_HEADER}, using default version: ${SLANG_VERSION}")
```
skallweitNV: I think we can omit this warning as it's not relevant. If the file is not available we are on an old unversioned version of slang, so there is no problem, i.e. nothing to warn about IMO.

external/CMakeLists.txt
```diff
@@ -108,58 +108,86 @@ set(SLANG_INCLUDE_DIR ${slang_SOURCE_DIR}/include)
...
-    sgl_copy_binary(${SLANG_DIR}/lib/libslang-glslang.dylib .)
-    sgl_copy_binary(${SLANG_DIR}/lib/libslang-glsl-module.dylib .)
-    # sgl_copy_binary(${SLANG_DIR}/lib/libslang-llvm.dylib .)
-    sgl_copy_binary(${SLANG_DIR}/lib/libslang-rt.dylib .)
+    # Extract slang version from local build's tag-version header
+    # Only applicable for versioned slang builds (since slang-compiler rename)
+    set(SLANG_TAG_VERSION_HEADER "${SLANG_DIR}/include/slang-tag-version.h")
+    if(EXISTS "${SLANG_TAG_VERSION_HEADER}")
+        extract_define_from_header("${SLANG_TAG_VERSION_HEADER}" "SLANG_VERSION_NUMERIC" SLANG_VERSION_FROM_HEADER)
+        if(SLANG_VERSION_FROM_HEADER)
```
skallweitNV: I would probably omit these warnings as well. Instead we should error out below when we detect `slang-compiler` library being used and `SLANG_VERSION_FROM_HEADER` is not available. Actually, maybe we should move parsing the `slang-tag-version.h` down there as well, we really just need to do this when dealing with new slang releases after all.

external/CMakeLists.txt
```diff
@@ -108,58 +108,86 @@ set(SLANG_INCLUDE_DIR ${slang_SOURCE_DIR}/include)
...
+        else()
+            set(slang_lib_suffix "so")
+        endif()
     endif()
-endif()
+
+    # Determine whether to use versioned slang-compiler or old name
+    if(EXISTS ${slang_lib_dir}/${slang_lib_prefix}slang-compiler.${slang_lib_suffix})
+        set(slang_lib_base slang-compiler)
+        set(slang_module_version "-${SLANG_VERSION}")
```
skallweitNV, line 148: Here is where we need to be able to get the slang version from the header file. If that fails, we are screwed and should error out.

cmake/utils.cmake
jhelferty-nv: Done. Added partial support for `/**/` (except for multiline).

cmake/utils.cmake
jhelferty-nv: Done.

external/CMakeLists.txt
jhelferty-nv: Yeah, I've turned them into errors after moving them down.

external/CMakeLists.txt
jhelferty-nv: Made the other cases an error.

Windows is a bit of a special case; we don't version the windows modules, so it isn't really needed. I left it in the logic flow, though, just because it's more confusing if I special case it.

skallweitNV (approved):

PR #621, opened 2025-11-05
Sign versioned .so files

jhelferty-nv:
Address slangpy versioning issue from shader-slang/slang#8901


jhelferty-nv:
@gtong-nv Thoughts on whether I should try testing this before checking it in with a wheel action, or just commit it and see if it fixes the signing problem? (I'm not sure what the side effects are of the wheel action, I assume it would push?)

jhelferty-nv:
Built successfully in workflow `wheels #80`. I pulled down the macos wheel for 3.10 and tested locally, worked with one of the samples. Going ahead and merging.

Issue #622, opened 2025-11-05
Slangpy release wheels failed to upload to PyPi


gtong-nv:
Since v0.38.0, the wheels are built but not able to upload to PyPi
https://github.com/shader-slang/slangpy/actions/runs/19090807896

```
Checking dist/slangpy-0.38.0-cp310-cp310-macosx_15_0_arm64.whl: PASSED
Traceback (most recent call last):
  File "/root/.local/bin/twine", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/root/.local/lib/python3.13/site-packages/twine/__main__.py", line 33, in main
    error = cli.dispatch(sys.argv[1:])
  File "/root/.local/lib/python3.13/site-packages/twine/cli.py", line 139, in dispatch
    return main(args.args)
  File "/root/.local/lib/python3.13/site-packages/twine/commands/check.py", line 195, in main
    return check(parsed_args.dists, strict=parsed_args.strict)
  File "/root/.local/lib/python3.13/site-packages/twine/commands/check.py", line 141, in check
    warnings, is_ok = _check_file(filename, render_warning_stream)
                      ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/lib/python3.13/site-packages/twine/commands/check.py", line 84, in _check_file
    package = package_file.PackageFile.from_filename(filename, comment=None)
  File "/root/.local/lib/python3.13/site-packages/twine/package.py", line 202, in from_filename
    data = dist.read()
  File "/root/.local/lib/python3.13/site-packages/twine/wheel.py", line 56, in read
    archive = zipfile.ZipFile(fqn)
  File "/usr/local/lib/python3.13/zipfile/__init__.py", line 1385, in __init__
    self._RealGetContents()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/zipfile/__init__.py", line 1482, in _RealGetContents
    raise BadZipFile("Bad magic number for central directory")
zipfile.BadZipFile: Bad magic number for central directory
Checking dist/slangpy-0.38.0-cp310-cp310-manylinux_2_34_x86_64.whl: 
```

PR #623, opened 2025-11-05
Verify wheel builds before upload to PyPi

gtong-nv:
1. Wheel Corruption During CI/CD
    - Issue: Wheels were valid after building but corrupted after artifact download
    - Cause: GitHub Actions merge-multiple: true was corrupting binary files
    - Fix: Manual artifact combination with separate download paths
2. License Metadata Errors
    - Issue: malformed field 'license-file' errors during PyPI upload
    - Cause: Metadata version mismatch between setuptools auto-detection and PEP 639
    - Fix: PEP 639 compliant license configuration in pyproject.toml
3. Verify the wheels with zip and twine.
    - `twine` has a known issue with old `packaging`; install newer `packaging` with pip

Fixes: https://github.com/shader-slang/slangpy/issues/622

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -71,7 +71,7 @@ set(SGL_LOCAL_SLANG OFF CACHE BOOL "Use a local build of slang instead of downlo
 set(SGL_LOCAL_SLANG_DIR "${CMAKE_SOURCE_DIR}/../slang" CACHE PATH "Path to a local slang build")
 set(SGL_LOCAL_SLANG_BUILD_DIR "build/Debug" CACHE STRING "Build directory of the local slang build")
 
-set(SLANG_VERSION "2025.21")
+set(SLANG_VERSION "2025.21.1")
 set(SLANG_URL_BASE "https://github.com/shader-slang/slang/releases/download/v${SLANG_VERSION}/slang-${SLANG_VERSION}")
 
 if(SGL_WINDOWS)
--- a/src/sgl/core/data_struct.cpp
+++ b/src/sgl/core/data_struct.cpp
@@ -699,6 +699,12 @@ struct X86Program : public Program {
 
     static std::unique_ptr<Program> compile(const DataStruct& src_struct, const DataStruct& dst_struct)
     {
+        // Require SSE4.1 as minimum for JIT compilation
+        // Pre-2008 CPUs without SSE4.1 will fall back to VM interpreter
+        if (!runtime().cpuFeatures().x86().hasSSE4_1()) {
+            return nullptr;
+        }
+
         asmjit::CodeHolder code;
         code.init(runtime().environment(), runtime().cpuFeatures());
 
@@ -736,6 +742,7 @@ private:
         asmjit::x86::Compiler& c;
         bool has_avx;
         bool has_f16c;
+        bool has_fma;
 
         // Each register can hold either a 64-bit integer or a 64-bit floating point value.
         struct Register {
@@ -749,8 +756,10 @@ private:
         Builder(asmjit::x86::Compiler& c)
             : c(c)
         {
+            // SSE4.1 is guaranteed at this point (checked in compile())
             has_avx = runtime().cpuFeatures().x86().hasAVX();
             has_f16c = runtime().cpuFeatures().x86().hasF16C();
+            has_fma = runtime().cpuFeatures().x86().hasFMA();
         }
 
         Register& get_register(uint32_t reg)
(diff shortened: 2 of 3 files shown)


.github/workflows/wheels.yml
```diff
@@ -78,6 +79,37 @@ jobs:
           name: wheels-${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.python }}
           path: ./wheelhouse/*.whl
 
+      - name: Verify wheels
```
skallweitNV: CIBW already has a step to verify the wheels, should we really add another one?
The corruption happened when fetching the artifacts AFAICT so this seems just extra churn to me.

.github/workflows/wheels.yml
```diff
@@ -78,6 +79,37 @@ jobs:
...
+          python -c "import twine; print(f'twine: {twine.__version__}')"
+          echo "=== Checking individual wheels ==="
+          for wheel in ./wheelhouse/*.whl; do
+            if [ -f "$wheel" ]; then
+              size=$(stat -c%s "$wheel" 2>/dev/null || stat -f%z "$wheel" 2>/dev/null || echo "unknown")
+              echo "Checking: $wheel (${size} bytes)"
+
+              # Check if it's a valid ZIP
+              if python -c "import zipfile; zipfile.ZipFile('$wheel').testzip()" 2>/dev/null; then
+                echo "  ✓ Valid ZIP format"
```
skallweitNV: I'm personally not a fan of using extended UTF characters or emojis in logs. The problem becomes apparent when looking at the logs in basic text editor (or just access raw logs on github in the browser).

.github/workflows/wheels.yml
```diff
@@ -87,13 +119,65 @@ jobs:
     environment: pypi
 
     steps:
+      # Alternative: Download artifacts separately to avoid merge corruption
```
skallweitNV: This comment really doesn't tell me much

.github/workflows/wheels.yml
```diff
@@ -87,13 +119,65 @@ jobs:
...
       - uses: actions/download-artifact@v4
         with:
           pattern: wheels-*
-          merge-multiple: true
-          path: dist
+          path: artifacts
```
skallweitNV, line 93: So did `merge-multiple` recently changed behavior? I'm glad you found a workaround, but I still don't have any idea what the real issue was. Why did this work in the past?

.github/workflows/wheels.yml
```diff
@@ -87,13 +119,65 @@ jobs:
...
+        run: |
+          echo "=== Files in dist/ ==="
+          find dist/ -type f -exec ls -la {} \;
+          echo "=== Checking for duplicate filenames ==="
+          find dist/ -name "*.whl" -exec basename {} \; | sort | uniq -d
+          echo "=== File types ==="
+          find dist/ -name "*.whl" -exec file {} \;
+
+      # Verify downloaded wheels before upload.
+      - name: Verify downloaded wheels
```
skallweitNV, line 104: Again, I'm questioning this verification a bit. We're basically testing that GitHub didn't screw something up during upload/download. I guess it makes sense to be a little defensive here, although again, I don't really understand what the root problem was. BUT, I think we should either check the ZIP or run the twine check, but probably not both.

.github/workflows/wheels.yml
gtong-nv: I will remove this, and only do a verification after downloading the artifacts.

.github/workflows/wheels.yml
gtong-nv: removed

.github/workflows/wheels.yml
gtong-nv: I'm not sure about the root cause. AI suggests this could be a problem when there are multiple files with identical/similar names.
Changed it to "Manually combine artifacts" resolved the ` BadZipFile("Bad magic number for central directory")` issue.

.github/workflows/wheels.yml
gtong-nv: I will remove the ZIP check. The `gh-action-pypi-publish` below is not helpful when giving out error message, especially unclear which of the *.whl is problematic, hence verifying here one by one.
Besides, it still upload corrupcted files despite of its check. See v0.38.0 uploaded but yanked  https://pypi.org/project/slangpy/#history

>  don't really understand what the root problem was
I'm not 100% sure either. I tried dozens of times with different tricks trying to get it work, including 
1. not using  ` merge-multiple`
2. upgrade `packaging` version
3. update `license-file` field - malformed meta data otherwise

PR #624, opened 2025-11-05
Fix asmjit usage on old x86_64

### LINKED ISSUE #364: vfmadd213sd use incorrectly gated by AVX feature flag
`src/sgl/core/data_struct.cpp` incorrectly uses the AVX feature flag to gate the use of `vfmadd213sd`, instead of the FMA feature flag, which results in an illegal instruction error on my (old) i7-3820.

jhelferty-nv:
Fixes #364

Fixed detection and fallback paths for old x86_64 processors when deciding what instructions to generate using asmjit

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -71,7 +71,7 @@ set(SGL_LOCAL_SLANG OFF CACHE BOOL "Use a local build of slang instead of downlo
 set(SGL_LOCAL_SLANG_DIR "${CMAKE_SOURCE_DIR}/../slang" CACHE PATH "Path to a local slang build")
 set(SGL_LOCAL_SLANG_BUILD_DIR "build/Debug" CACHE STRING "Build directory of the local slang build")
 
-set(SLANG_VERSION "2025.21")
+set(SLANG_VERSION "2025.21.1")
 set(SLANG_URL_BASE "https://github.com/shader-slang/slang/releases/download/v${SLANG_VERSION}/slang-${SLANG_VERSION}")
 
 if(SGL_WINDOWS)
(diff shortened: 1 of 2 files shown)


src/sgl/core/data_struct.cpp
```diff
@@ -736,6 +736,8 @@ struct X86Program : public Program {
         asmjit::x86::Compiler& c;
         bool has_avx;
         bool has_f16c;
+        bool has_sse4_1;
```
skallweitNV: do we really need to support the case where SSE 4.1 is not supported? seems like that would only be the case on really really old CPUs that have no business running slangpy?!

skallweitNV (commented):
The AVX checks make sense, I'd get rid of the sse4.1 check though

src/sgl/core/data_struct.cpp
skallweitNV: To elaborate a bit, my worry is that there might be more places where we rely on SSE 4.1?! Without properly testing this on a platform that doesn't have SSE 4.1, we're probably better of falling back to the non-jitted version if basic SSE is not available.

jhelferty-nv: I'm not sure, tbh. I only added it because it was easy to get an LLM to generate it, and the test suite passed when I forced all of the 'has' features to false.

jhelferty-nv: Either way, the branch is only evaluated at generation time. I don't think it should impact performance of the fast path.

jhelferty-nv: Per discussion, instead of providing an asmjit path for processors without SSE 4.1, the code now falls back to non-asmjit path.

PR #625, opened 2025-11-06
TensorRef Marshalling Optimization

szihs:
Add fast path for PyTorch tensors in NativeTensorMarshall to bypass buffer allocation and directly write tensor metadata (shape, strides, data pointers) to shader cursors. This optimization significantly reduces kernel launch latency when using PyTorch tensors with slangpy.

   ## Key Changes

   - Add `write_pytorch_tensor_fields()` to handle PyTorch tensors directly
   - Extract shape and strides using pre-allocated vectors
   - Support differential tensors (primal, grad_in, grad_out)
   - Apply broadcast stride zeroing for transformed tensors
   - Fall back to base class for non-PyTorch tensor types

### DIFF
--- a/src/slangpy_ext/utils/slangpytensor.cpp
+++ b/src/slangpy_ext/utils/slangpytensor.cpp
@@ -332,20 +332,23 @@ void NativeTensorMarshall::write_shader_cursor_pre_dispatch(
                     "inout parameter gradients need separate buffers for inputs and outputs (see Tensor.with_grads)"
                 );
         }
-    } else {
-        TensorRef* tensorref;
-        if (nb::try_cast(value, tensorref)) {
-            auto pytorch_tensor_opt = tensorref->tensor();
-            // Only use fast path for CUDA tensors - other backends need interop buffer
-            if (pytorch_tensor_opt.has_value() && context->device()->type() == DeviceType::cuda) {
-                ShaderCursor field = cursor[binding->variable_name()];
-                write_pytorch_tensor_fields(context, binding, field, tensorref, read_back);
-                return;
-            }
-        }
-        // Fall back to base class
-        NativeMarshall::write_shader_cursor_pre_dispatch(context, binding, cursor, value, read_back);
+        return;
     }
+
+    // Check if we have a TensorRef with PyTorch tensor for fast path
+    TensorRef* tensorref;
+    if (nb::try_cast(value, tensorref)) {
+        auto pytorch_tensor_opt = tensorref->tensor();
+        // Only use fast path for CUDA tensors - other backends need interop buffer
+        if (pytorch_tensor_opt.has_value() && context->device()->type() == DeviceType::cuda) {
+            ShaderCursor field = cursor[binding->variable_name()];
+            write_pytorch_tensor_fields(context, binding, field, tensorref, read_back);
+            return;
+        }
+    }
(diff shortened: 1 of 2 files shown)


src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -162,6 +333,17 @@ void NativeTensorMarshall::write_shader_cursor_pre_dispatch(
                 );
         }
```
ccummingsNV, line 334: Correct logic, but I would structure the code as:

`
if(bla) {
    //do stuff
   return;   
}
if(bla2) {
    //do stuff
    return;
}

//fallback to base class
foo()
}
`

i.e. succesive if/return statements. Bit easier to read, and avoids deep nesting when adding more conditions.

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -123,6 +151,149 @@ Shape NativeTensorMarshall::get_shape(nb::object data) const
...
+
+    // Check trailing dimensions match (like Python: shape[-len(expected):])
+    if (expected_shape_vec.size() > 0) {
+        size_t start_idx = tensor_shape.size() - expected_shape_vec.size();
+        for (size_t i = 0; i < expected_shape_vec.size(); i++) {
+            int expected_dim = expected_shape_vec[i];
+            int tensor_dim = tensor_shape[start_idx + i];
+            // -1 means any size is allowed
+            if (expected_dim != -1 && tensor_dim != expected_dim) {
+                throw nb::value_error(
```
ccummingsNV, line 191: Unless there's a specific reason to use nb::value_error, use SGL_THROW

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -17,6 +17,34 @@ extern void write_shader_cursor(ShaderCursor& cursor, nb::object value);
 
 namespace sgl::slangpy {
 
+namespace {
+    /// Helper function to extract shape from PyTorch tensor
+    /// Pre-allocates vector to avoid repeated allocations
+    std::vector<int> extract_shape(const nb::ndarray<nb::pytorch, nb::device::cuda>& tensor)
+    {
+        std::vector<int> shape;
```
ccummingsNV, line 25: A future potential optimization both here and inside the 'shape' class will be to use SGL's short vector. I suspect we'll save a lot that way, but I recommend doing a 'use short vectors where possible' pass as a single PR rather than mixing it in with this one.

ccummingsNV (changes requested):
A couple of minor changes to suggest but otherwise looks great

src/slangpy_ext/utils/slangpytensor.cpp
szihs: It's required else the pytest fails will 

FAILED slangpy/tests/slangpy_tests/test_torchintegration.py::test_add_values_fail[func_and_shape2-1-DeviceType.cuda] - RuntimeError: Tensor shape (5, 5) does not match expected shape (3)

src/slangpy_ext/utils/slangpytensor.cpp
szihs: Sure. We can take it for the future PR

PR #627, opened 2025-11-07
Added Texture::descriptor_handle getters based on the default views.

tdavidovicNV:
This adds convenience methods to use with bindless textures.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -2,8 +2,12 @@ name: wheels
 on:
   workflow_dispatch:
     inputs:
-      upload:
+      upload_to_pypi:
         description: 'Upload wheels to PyPI? (0: no, 1: yes)'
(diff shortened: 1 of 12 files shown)


src/sgl/device/resource.h
```diff
@@ -593,6 +593,9 @@ class SGL_API Texture : public Resource {
 
     ref<TextureView> create_view(TextureViewDesc desc);
 
+    DescriptorHandle descriptor_handle_ro() const;
+    DescriptorHandle descriptor_handle_rw() const;
```
skallweitNV, line 599:
```suggestion
    /// Get bindless descriptor handle for read access.
    DescriptorHandle descriptor_handle_ro() const;
    /// Get bindless descriptor handle for read-write access.
    DescriptorHandle descriptor_handle_rw() const;
```

PR #628, opened 2025-11-08
docs: fix incorrect Tensor constructor API in autodiff documentation

### LINKED ISSUE #537: Documentation error: Tensor constructor with element_type parameter doesn't exist
## Issue Description

The autodiff documentation references a `Tensor` constructor API that does not exist in the current SlangPy implementation.

## Documentation Reference

Source: https://github.com/shader-slang/slangpy/blob/147a6f15c141948effca5219006e3d4d587512f7/docs/src/autodiff/autodiff.rst#L29

The documentation shows:
```python
result = spy.Tensor(device, element_type=module.float, shape=(4,))
module.polynomial(a=2, b=8, c=-1, x=x, _result=result)
```

## Actual API

The actual `Tensor.__init__` signature is:
```python
(self, storage: 'Buffer', dtype: 'SlangType', shape: 'TShapeOrTuple', strides: 'Optional[TShapeOrTuple]' = None, offset: 'int' = 0, grad_in: 'Optional[Tensor]' = None, grad_out: 'Optional[Tensor]' = None)
```

Issues:
1. **No `element_type` parameter exists** - it should be `dtype`
2. **Constructor requires `storage` as first parameter**, not `device`
3. The constructor is low-level and not intended for typical usage

## Correct Approach

Users should use factory methods instead:
```python
# Correct way to pre-allocate a tensor
result = spy.Tensor.empty(device, shape=(4,), dtype=module.float)
module.polynomial(a=2, b=8, c=-1, x=x, _result=result)
```

Or use the string-based approach:
```python
result: spy.Tensor = module.polynomial(a=2, b=8, c=-1, x=x, _result="tensor")
```

## Related Discussion

Discord thread: https://discord.com/channels/1303735196696445038/1305995870046650368/1422744660593344643

## Suggested Fix

Update the documentation to show one of these correct patterns:
- `spy.Tensor.empty(device, shape=..., dtype=...)`
- `spy.Tensor.zeros(device, shape=..., dtype=...)`
- Or keep using the `_result="tensor"` string approach which works fine

## Environment
- Checked against current SlangPy installation
- All factory methods (`empty`, `zeros`, `from_numpy`) consistently use `dtype`, not `element_type`

aidanfnv:
Replace incorrect `spy.Tensor(device, element_type=...)` call with correct `spy.Tensor.empty(device, dtype=...)` factory method.

The Tensor constructor requires a Buffer storage parameter as the first argument and uses 'dtype' not 'element_type'. Users should use factory methods like empty(), zeros(), or from_numpy() for typical usage.

Fixes #537

Generated with [Claude Code](https://claude.ai/code)


jkwak-work (approved):
Looks good to me

.github/workflows/ci.yml
```diff
@@ -15,7 +15,6 @@ on:
   pull_request:
     branches: [main]
     paths-ignore:
-      - "docs/**"
```
jkwak-work, line 18: Where can I find the branch rule that requires build to pass?
If that is the case, the setting of "paths-ignore" here makes no sense.

We had the same problem on slang repo and I am not sure how we resolved it.

jkwak-work (approved):
Looks good to me.

But I think we should properly address the "paths-ignore" setting problem.

.github/workflows/ci.yml
aidanfnv: You can find the rules in the repo's settings, under `Settings -> Branches -> Branch protection rules -> main`.
Click on Edit, and you find the checks under `Protect matching branches -> Require status checks to pass before merging -> Status checks that are required`.

In the slang repo, we do not require that the individual CI checks pass, instead we have a job named `check-ci` that is required, and it fails if any of the CI jobs failed or was cancelled, and passes otherwise. The first job in the CI workflow is `filter`, which sets a flag `should-run` to `false` for docs-only changes and `true` otherwise, and the CI jobs only run if that flag is `true` and are otherwise skipped. That way we still get a pass/fail check without actually running the CI.

PR #631, opened 2025-11-10
Add test for extension cache issue

jhelferty-nv:
Test for shader-slang/slang#8954

Add a test to verify that the extension cache gets updated after adding modules interactively.

### DIFF
--- a/src/sgl/device/resource.cpp
+++ b/src/sgl/device/resource.cpp
@@ -496,6 +496,20 @@ ref<TextureView> Texture::create_view(TextureViewDesc desc)
     return m_device->create_texture_view(this, std::move(desc));
 }
 
+DescriptorHandle Texture::descriptor_handle_ro() const
+{
+    rhi::DescriptorHandle rhi_handle = {};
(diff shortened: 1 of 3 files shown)


slangpy/tests/slangpy_tests/test_reflection2.py
```diff
@@ -371,5 +371,44 @@ def test_refl_duck_typing(device_type: DeviceType):
         compare_struct_values(refl_val, spy_val)
 
 
+# Limit the device types tested here to just one, since these tests are not device specific.
```
ccummingsNV: You can run on all devices - I've lost count of the number of times we've had a 'not device specific' test break on a specific device :)

ccummingsNV (changes requested):
Run on all devices, otherwise LGTM

slangpy/tests/slangpy_tests/test_reflection2.py
jhelferty-nv: Fixed. I've also run the commit hook.

PR #637, opened 2025-11-12
Add precompiled module test

mkeshavaNV:


skallweitNV:
Can you check if setting `slang::CompilerOptionName::UseUpToDateBinaryModule` breaks precompiled module loading? If so, we should figure out how Slang needs to change to support this use-case, because we need that flag for the module cache, which should work side-by-side with precompiled modules (without source).

slangpy/tests/device/test_precompiled_module.slang
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import sgl.device.print;
+
+[shader("compute")]
+[numthreads(1, 1, 1)]
+void compute_main(uint3 tid: SV_DispatchThreadID)
+{
+    print("Hello from precompiled module!");
+}
```
skallweitNV: you never call this .. might as well just get rid of it

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+        / "_deps"
+        / "slang-src"
+        / "bin",
+        Path(__file__).parent.parent.parent.parent
+        / "build"
+        / "macos-arm64-clang"
+        / "_deps"
+        / "slang-src"
+        / "bin",
+    ]
```
skallweitNV, line 33: you could write something like
```python
root_dir = Path(__file__).parent.parent.parent.parent
build_dirs = [
    root_dir / "build/pip/_deps/slang-src/bin",
    root_dir / "build/linux-gcc/_deps/slang-src/bin",
    root_dir / "build/windows-msvc/_deps/slang-src/bin",
    root_dir / "build/macos-arm64-clang/_deps/slang-src/bin",
]
```

note to self: we need to get rid of the different build folders ..

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+    for build_dir in build_dirs:
+        slangc_path = build_dir / "slangc"
+        if slangc_path.exists():
+            return slangc_path
+        slangc_path = build_dir / "slangc.exe"
+        if slangc_path.exists():
+            return slangc_path
+
+    # Try to find slangc in PATH
+    slangc_path = shutil.which("slangc")
```
skallweitNV: this feels dangerous to me, we might pick up some other `slangc`

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+    if slangc_path:
+        return Path(slangc_path)
+
+    # If not found, fail in CI but return None locally
+    if is_running_in_ci():
```
skallweitNV: i would move check to the main test function

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+        raise RuntimeError("slangc compiler not found in CI environment")
+
+    return None
+
+
+def compile_precompiled_module(
```
skallweitNV, line 46: maybe pass in the slangc path here, so you don't need to call `find_slangc()`

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+def test_precompiled_module_without_source(device_type: spy.DeviceType, tmpdir: str):
+    """Test that precompiled modules are loaded when source .slang file is not present.
+
+    This definitively verifies that .slang-module files are used by deleting the source
+    after compilation and confirming the module still loads and executes correctly.
+    """
+    # Check if slangc is available
+    slangc = find_slangc()
+    if slangc is None:
+        pytest.skip("slangc compiler not found (skipping outside CI)")
```
skallweitNV, line 83: this logic becomes cleaner when the CI check is here as well

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+        test_dir.parent.parent.parent / "src",
+    ]
+    for shader_dir in potential_shader_dirs:
+        if shader_dir.exists() and (shader_dir / "sgl" / "device" / "print.slang").exists():
+            sgl_shaders_dir = shader_dir
+            break
+
+    include_paths = []
+    if sgl_shaders_dir:
+        include_paths.append(sgl_shaders_dir)
```
skallweitNV: is the plan here to test importing other modules (`sgl.device.print`)

slangpy/tests/device/test_precompiled_modules.py
```diff
@@ -0,0 +1,180 @@
...
+    # Check if slangc is available
+    slangc = find_slangc()
+    if slangc is None:
+        pytest.skip("slangc compiler not found (skipping outside CI)")
+
+    test_dir = Path(__file__).parent
+    cache_dir = Path(tmpdir)
+
+    source_file = test_dir / "test_precompiled_module.slang"
+    source_copy = cache_dir / "test_no_source.slang"
```
skallweitNV: why do we create a copy of the source file? couldn't we just precompile the `test_precompiled_module.slang` to `test_no_source.slang-module` directly? or does slangc determine the module name from the input filename?

skallweitNV (commented):
Some comments on how this could be cleaned up

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: Yeah I was originally trying that, but removed all that  from the slang file.  Removed all this code.

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: done, yeah that works.

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: done

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: done

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: done

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: ok, trying to find slangc from the build directories alone.

slangpy/tests/device/test_precompiled_modules.py
mkeshavaNV: done

slangpy/tests/device/test_precompiled_module.slang
mkeshavaNV: done

PR #641, opened 2025-11-14
scalar/vector "select" intrinsic

skallweitNV:


src/sgl/math/vector_math.h
```diff
@@ -31,6 +31,33 @@ namespace sgl::math {
...
 
+// ----------------------------------------------------------------------------
+// Selection
+// ----------------------------------------------------------------------------
+
+template<typename T, int N>
+[[nodiscard]] constexpr vector<T, N>
+select(vector<bool, N> condition, vector<T, N> true_value, vector<T, N> false_value) noexcept
+{
+    if constexpr (N == 1)
```
ccummingsNV, line 42: I reckon the compiler would work this out with a for loop :) Your call though

PR #651, opened 2025-11-24
Added support for assigning objects with to_cursor to cursor objects

tdavidovicNV:
We now allow assigning objects to cursors (buffer and shader), if those objects support this interface:
```
template<typename TCursor>
    void to_cursor(TCursor& cursor) const
```
When such objects are assigned to a cursor (using `set` or `operator=`) we do not call the generic element-wise `set`, but instead call the `to_cursor`, allowing assigning such objects without unpacking their members in the calling code.

An example would be, before:
```
struct TextureHandle
{
    uint handle_data;
};
```

had to be assigned like so:
```
cursor["my_handle"]["handle_data"] = my_handle.handle_data;
```

with this change, we can do this:
```
struct TextureHandle
{
    uint handle_data;

    template<typename TCursor>
        void to_cursor(TCursor& cursor) const
    {
        cursor["handle_data"] = handle_data;
    }
};

cursor["my_handle"] = my_handle; // equal to `my_handle.to_cursor(cursor["my_handle"])`
```

This allows, among other things, much easier generic programming when setting Slang structs with mix of PODs and structs.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -210,6 +210,9 @@ set(SLANG_RHI_SLANG_INCLUDE_DIR ${SLANG_INCLUDE_DIR} CACHE STRING "" FORCE)
 set(SLANG_RHI_SLANG_BINARY_DIR ${SLANG_DIR} CACHE STRING "" FORCE)
 add_subdirectory(slang-rhi)
 
+# Copy shared slang-rhi-device.h header to shaders directory (wrapped in sgl/device/rhi.slang)
+sgl_copy_shader_file(${CMAKE_CURRENT_SOURCE_DIR}/slang-rhi/include/slang-rhi-device.h slang-rhi)
+
 if(TARGET slang-rhi-nvapi)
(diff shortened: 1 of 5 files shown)


src/sgl/device/shader_cursor.h
```diff
@@ -84,6 +84,14 @@ class SGL_API ShaderCursor : public CursorWriteWrappers<ShaderCursor, ShaderOffs
     }
 
     template<typename T>
+        requires(HasToCursor<T, BufferElementCursor>)
+    void set(const T& value)
+    {
+        value.to_cursor(*this);
```
skallweitNV: `to_cursor` feels strange to me .. something like `bind_to_cursor` or just `bind` would be more appropriate maybe?!

skallweitNV: I still find `to_cursor` confusing, it implies to me that we convert the object it's called on to a cursor. Can we do `write_to_cursor` or `bind_to_cursor` instead?

skallweitNV (dismissed):

PR #657, opened 2025-12-01
Fixed the way link is made unique, to preserve the module order.

tdavidovicNV:


slangpy/core/module.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         self.slangpy_device_module = device_module.session.load_module("slangpy")
 
         # Extract linked modules
-        self.link = list(set([x.module if isinstance(x, Module) else x for x in link]))
+        self.link = list(dict.fromkeys([x.module if isinstance(x, Module) else x for x in link]))
```
skallweitNV, line 65: Maybe add a comment that `dict.fromkeys` is used to preserve the module order but removes duplicates.
Side question, what happens when a module is added more than once in a linkage?

tdavidovicNV: Will do the comment.

What actually happens? I don't know. What should happen, the first one is taken and the rest is ignored.

Issue #664, opened 2025-12-16
Support generic entrypoint


kaizhangNV:
Slangpy currently doesn't support writing generic entrypoint, e.g.

```
[shader("compute")]
void computeMain<T: I, int a, float b>()
{

}
```

Currently slangpy has no method to specialize this entry point, even though we can still find this function in reflection API. However specializing an entry point is different from specializing a normal function. To specialize an entry point, a systematical way is that we need to get the `IEntryPoint` component type, and call `IEntryPoint::specialize()` to specialize the generic entry point.

PR #670, opened 2025-12-31
Support generic entrypoint

wlewNV:
Issue #664

Add the ability to specialize generic shader entrypoints with concrete types or expressions, enabling use of Slang's generic/template features.

Changes:
- Add SpecializationArg struct with Kind enum (mirrors slang)
- Extend SlangEntryPointDesc with specialization_args vector
- Add SlangEntryPoint::specialize() method that creates a new entrypoint with the given specialization arguments

New Tests:
- slangpy/tests/device/test_generic_entrypoint.slang
- slangpy/tests/device/test_generic_entrypoint.py


src/sgl/device/shader.cpp
```diff
@@ -40,6 +40,27 @@ std::string TypeConformance::to_string() const
...
+        kind_str = "unknown";
+        break;
+    case Kind::type:
+        kind_str = "type";
+        break;
+    case Kind::expr:
+        kind_str = "expr";
+        break;
+    }
+    return fmt::format("SpecializationArg(kind={}, value=\"{}\")", kind_str, value);
```
skallweitNV: You can just pass `kind` directly and it will be converted to a string because you defined names using `SGL_ENUM_INFO`.

```suggestion
    return fmt::format("SpecializationArg(kind={}, value=\"{}\")", kind, value);
```

src/sgl/device/shader.h
```diff
@@ -431,9 +431,51 @@ class SGL_API SlangModule : public Object {
     mutable std::set<SlangEntryPoint*> m_registered_entry_points;
 };
 
+/// \brief Specialization argument for generic entrypoints.
+/// Mirrors slang::SpecializationArg but uses string values for Python compatibility.
+struct SGL_API SpecializationArg {
+    /// Kind of specialization argument (mirrors slang::SpecializationArg::Kind).
+    enum class Kind {
```
skallweitNV: Can we move this outside and rename to `SpecializationArgKind`. This way the name is the same for C++ and Python.

src/sgl/device/shader.cpp
```diff
@@ -1039,6 +1060,53 @@ void SlangEntryPoint::init(SlangSessionBuild& build_data) const
...
+    // If we have specialization arguments, specialize the entry point.
+    if (!desc.specialization_args.empty()) {
+        slang::ProgramLayout* layout = slang_module->getLayout();
+
+        // Build the slang specialization args from our descriptors.
+        std::vector<slang::SpecializationArg> slang_spec_args;
+        slang_spec_args.reserve(desc.specialization_args.size());
+
+        // We need to keep string storage alive for expr args.
+        std::vector<std::string> expr_storage;
```
skallweitNV: Maybe I don't see the obvious but how is this different from using `arg.value.c_str()` directly? `desc.specialization_args` should be stable for the duration of calling slang's `specialize` function.

Code would be simpler/faster if we don't do this extra string copy.

Issue #636, opened 2025-11-12
Numpy arrays  created with np.dtype can't be passed to slangpy


ccummingsNV:
When constructing an array such as:

```
np.dtype([
    ('valid', np.float32, (1,)),
    ('material_id', np.int32, (1,)),
    ('uv', np.float32, (2,)),
    ('wi', np.float32, (3,)),
    ('wo', np.float32, (3,)),
    ('mip_level', np.int32, (1,)),
    ('target', np.float32, (3,)),
])
```

And then passing it to NDBuffer

```
gpu_batch = spy.NDBuffer(
    module.device,
    dtype=TrainingSample,
    shape=(num_samples,),
)
gpu_batch.copy_from_numpy(data[start:stop])
```

Nanobind fails unless you create a view onto the data of a basic datatype (eg uint8). This appears to be because dtype is not an ndarray:

```
data.dtype
dtype([('valid', '<f4', (1,)), ('material_id', '<i4', (1,)), ('uv', '<f4', (2,)), ('wi', '<f4', (3,)), ('wo', '<f4', (3,)), ('mip_level', '<i4', (1,)), ('target', '<f4', (3,))])
```

We should handle this case internally, and also extend `spy.NDBuffer.from_numpy` to support taking a data type, so the user doesn't have to construct it first.

ccummingsNV:
@bmillsNV not specifically, though I'll probably get to it eventually if nobody else does. Just recording as a feature we should add.

PR #671, opened 2026-01-05
Support structured numpy dtypes in NDBuffer.copy_from_numpy()

wlewNV:
Issue #636

Previously, passing ANY structured numpy dtype to NDBuffer.copy_from_numpy() would fail because nanobind's numpy binding doesn't accept compound dtypes.

This change does the following:
- copy_from_numpy() now uses BufferCursor for structured dtypes, which handles platform-specific alignment
- Simple dtypes (e.g., float32, int32) use a direct memory copy
- from_numpy() raises a helpful error for structured dtypes without explicit Slang type


ccummingsNV:
Closing this as it's been addressed for tensors in separate PR

PR #672, opened 2026-01-05
Add Python bindings for glfwSetWindowPos

tzeltnerNV:
This is useful for creating apps that have multiple open windows simultaneously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If you use SlangPy in a research project leading to a publication, please cite t
     title = {SlangPy},
     author = {Simon Kallweit and Chris Cummings and Benedikt Bitterli and Sai Bangaru and Yong He},
     note = {https://github.com/shader-slang/slangpy},
-    version = {0.39.0},
+    version = {0.40.1},
     year = 2025
 }
 ```
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,7 +9,46 @@ SlangPy uses a `semantic versioning <http://semver.org>`__ policy for its API.
 
 Next
 -------
-- Significant rewrite of type inference system for better handling of generics and complex types.
+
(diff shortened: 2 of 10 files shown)


tzeltnerNV:
I had some trouble re-generating `src/slangpy_ext/py_doc.h`. Eventually I think I succeeded but it resulted in >10k changed lines of code which seems suspicious...

tzeltnerNV:
I rebased my changes onto `main`. There are now only minimal changes in `src/slangpy_ext/py_doc.h`.

skallweitNV (approved):
LGTM, thanks!

src/slangpy_ext/core/window.cpp
```diff
@@ -54,6 +54,7 @@ SGL_PY_EXPORT(core_window)
     window.def_prop_ro("width", &Window::width, D(Window, width));
     window.def_prop_ro("height", &Window::height, D(Window, height));
     window.def("resize", &Window::resize, "width"_a, "height"_a, D(Window, resize));
+    window.def("set_position", &Window::set_position, "x"_a, "y"_a, D(Window, set_position));
```
skallweitNV, line 57: Actually, on second thought, maybe we should make this a property and add the getter as well.

tzeltnerNV: Good idea. Shall I do it via `std::pair<uint32_t, uint32_t>` or split it up into two separate getters/setters for x & y on the C++ side? I think nanobind should support properties with tuples.

skallweitNV: I extended `Window` class in this PR: https://github.com/shader-slang/slangpy/pull/698

Sorry, but this was quicker than explaining the changes I had in mind.

Issue #551, opened 2025-10-07
Slangpy Math library matrix multiplication incomplete


sleeplessai:
The slangpy.math library does not support all dimensions of matrix multiplication.
To reproduce using the following snippet,

```python
import slangpy as sgl

spalt2world = sgl.float4x3([-0.0402, 0.0673, 0.0824,        0.0000, -0.0626, -0.0861,       0.0398, 0.0000, -4.8637,        -1.1342, -6.3513, 1.0000,])
world2ndc = sgl.float4x4([-0.0724, 0.4652, 1.6573, 0.7338,        -0.5483, 2.4286, -0.7057, 1.5564,       -0.9765, -0.2153, 0.0178, 4.2501,       -0.9764, -0.2153, 0.0178, 4.2597,])
ndc2pix = sgl.float4x3([628.0000, 0.0000, 0.0000,       627.5000, 0.0000, 414.0000,     0.0000, 413.5000, 0.0000,       0.0000, 0.0000, 1.0000,])

full_transformation1 = sgl.math.transpose(spalt2world) @ world2ndc @ ndc2pix
full_transformation2 = sgl.math.mul(sgl.math.transpose(spalt2world), world2ndc) @ ndc2pix
```
I got the warning and error

>     full_transformation1 = sgl.math.transpose(spalt2world) @ world2ndc @ ndc2pix
>                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
> TypeError: unsupported operand type(s) for @: 'slangpy.math.float3x4' and 'slangpy.math.float4x4'

    full_transformation2 = sgl.math.mul(sgl.math.mul(sgl.math.transpose(spalt2world), world2ndc), ndc2pix)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

> TypeError: mul(): incompatible function arguments. The following argument types are supported:
>     1. mul(x: slangpy.math.float2x2, y: slangpy.math.float2x2) -> slangpy.math.float2x2
>     2. mul(x: slangpy.math.float2x2, y: slangpy.math.float2) -> slangpy.math.float2
>     3. mul(x: slangpy.math.float2, y: slangpy.math.float2x2) -> slangpy.math.float2
>     4. mul(x: slangpy.math.float2x3, y: slangpy.math.float3x2) -> slangpy.math.float2x2
[...]

PR #674, opened 2026-01-05
Fix incomplete matrix multiplication bindings

wlewNV:
Issue #551

Previously, the Python bindings for matrix multiplication only supported square results.

This fix adds all 27 valid matrix-matrix combinations: MxN @ NxP -> MxP where M,N,P ∈ {2,3,4}.


skallweitNV (approved):
LGTM

PR #676, opened 2026-01-06
Fix array-type returns

### LINKED ISSUE #380: Very poor errors when attempting to coerce Tensors with array types
From Tizian

Here is a short reproducer:
shape = (4096 * 4096 * 8,)
tensor = spy.Tensor.empty(module.device, shape, 'half')
tensor.storage.copy_from_numpy(np.zeros(shape, dtype=np.float16))
tensor = tensor.with_grads(zero=True)

module['tensor_test_channels<8>'](tensor)
for
half[NumChannels] tensor_test_channels<int NumChannels>(half[NumChannels] data)
{
    [ForceUnroll]
    for (int i = 0; i < NumChannels; ++i)
    {
        data[i] = 2.h * data[i];
    }
    return data;
}
[1:55](https://nvidia.slack.com/archives/C07JB8BB7Q8/p1753707347775299?thread_ts=1753701420.569139&cid=C07JB8BB7Q8)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "D:\src\Falcor-nm-slangpy\external\slangpy\slangpy\core\function.py", line 348, in __call__
    return self.call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\src\Falcor-nm-slangpy\external\slangpy\slangpy\core\function.py", line 231, in call
    return self._native_call(self.module.call_data_cache, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\src\Falcor-nm-slangpy\external\slangpy\slangpy\builtin\valueref.py", line 221, in create_output
    pt = slang_type_to_return_type(self.value_type)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\src\Falcor-nm-slangpy\external\slangpy\slangpy\builtin\value.py", line 66, in slang_type_to_return_type
    raise ValueError(f"Slang type {slang_type} has no associated python value type")
ValueError: Slang type Object(0x1aa307d24f8) has no associated python value type
[1:58](https://nvidia.slack.com/archives/C07JB8BB7Q8/p1753707533301159?thread_ts=1753701420.569139&cid=C07JB8BB7Q8)
The other issue seems to be that I cannot pass a shape (N, N, 8) of half into that function. Only (N, N) of half[8].
new_buffer_channels_wrong = spy.NDBuffer(module.device, 'half',    shape=(128, 128, 8))
new_buffer_channels_right = spy.NDBuffer(module.device, 'half[8]', shape=(128, 128))
I can bypass this with some slightly more awkward
test = spy.NDBuffer.empty(module.device, shape=(4096, 4096), dtype='half[8]')
test.copy_from_numpy(array)
instead of
spy.NDBuffer.from_numpy(module.device, array)
(array.shape == (4096, 4096, 8))

NV-xiaoyongs:
Fixes #380 

Fix array-type returns (e.g. half[8]) and improve NDBuffer.from_numpy for array dtypes:
- Fix poor error when returning Slang array types like half[8]; map ArrayType to Python lists.
- Update ValueRef marshalling to allocate/read/write ArrayType outputs correctly.
- Improve NDBuffer.from_numpy(..., dtype="half[8]") to infer buffer shape by stripping trailing element-shape dims (e.g. (H,W,8) -> (H,W)).
- Add tests covering half[8] return values and array-dtype buffer construction.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -104,7 +104,9 @@ elseif(SGL_MACOS)
     endif()
 endif()
 
-sgl_download_package(slang ${SLANG_URL})
+if(NOT SGL_LOCAL_SLANG)
+    sgl_download_package(slang ${SLANG_URL})
+endif()
 
 set(SLANG_DIR ${slang_SOURCE_DIR})
 set(SLANG_INCLUDE_DIR ${slang_SOURCE_DIR}/include)
--- a/slangpy/tests/core/test_keycode_world_keys.py
+++ b/slangpy/tests/core/test_keycode_world_keys.py
@@ -1,0 +1,14 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import slangpy as spy
+
+
+def test_keycode_world_1_and_2_exist():
+    # GLFW defines these in the "printable" range:
+    #   GLFW_KEY_WORLD_1 = 161 (non-US #1)
+    #   GLFW_KEY_WORLD_2 = 162 (non-US #2)
+    #
+    # When reported via KeyboardEvent.key, the Python wrapper must be able to
+    # materialize these as valid KeyCode enum values (no ValueError).
+    assert spy.KeyCode(161) == spy.KeyCode.world_1
+    assert spy.KeyCode(162) == spy.KeyCode.world_2
--- a/slangpy/tests/device/test_coopvec.py
+++ b/slangpy/tests/device/test_coopvec.py
@@ -18,12 +18,12 @@ SIZE_CHECKS = [
     (4, 4, spy.DataType.float32, 64),
     (4, 8, spy.DataType.float32, 128),
     (8, 4, spy.DataType.float32, 128),
-    (4, 4, spy.DataType.float16, 64),  # rounded up to 64 from 32
+    (4, 4, spy.DataType.float16, 32),
     (4, 8, spy.DataType.float16, 64),
     (8, 4, spy.DataType.float16, 64),
-    (4, 4, spy.DataType.float8_e4m3, 64),  # rounded up to 64 from 16
+    (4, 4, spy.DataType.float8_e4m3, 16),
     (8, 16, spy.DataType.float8_e4m3, 128),
-    (4, 4, spy.DataType.float8_e5m2, 64),  # rounded up to 64 from 16
+    (4, 4, spy.DataType.float8_e5m2, 16),
     (16, 8, spy.DataType.float8_e5m2, 128),
 ]
 
--- a/slangpy/tests/device/test_generic_entrypoint.py
+++ b/slangpy/tests/device/test_generic_entrypoint.py
@@ -1,0 +1,117 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+import numpy as np
+
+import slangpy as spy
+from slangpy.testing import helpers
+
+
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_generic_entrypoint_specialize(device_type: spy.DeviceType):
+    """Test specializing a generic entrypoint with a concrete type."""
+    device = helpers.get_device(type=device_type)
+
+    module = device.load_module("test_generic_entrypoint.slang")
+
+    # Get the generic entrypoint
+    generic_entry = module.entry_point("compute_main")
+
+    # Specialize it with DoubleIt
+    specialized_entry = generic_entry.specialize([spy.SpecializationArg.from_type("DoubleIt")])
+
+    # Link and create kernel
+    program = device.link_program(modules=[module], entry_points=[specialized_entry])
+    kernel = device.create_compute_kernel(program)
+
+    # Create input data
+    input_data = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)
+    buffer = device.create_buffer(
+        data=input_data,
+        usage=spy.BufferUsage.shader_resource | spy.BufferUsage.unordered_access,
+    )
+
+    # Dispatch
+    kernel.dispatch(thread_count=[4, 1, 1], data=buffer, count=4)
+
+    # Verify results (DoubleIt multiplies by 2)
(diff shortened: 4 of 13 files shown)


src/sgl/core/lmdb_cache.cpp
```diff
@@ -129,23 +129,40 @@ void LMDBCache::set(const void* key_data, size_t key_size, const void* value_dat
...
-        LMDB_THROW("Failed to write data", result);
-
-    MetaData meta_data{.last_access = get_current_time_ns()};
-    MDB_val mdb_val_meta = {sizeof(MetaData), &meta_data};
-    if (int result = mdb_put(txn, m_db.dbi_meta, &mdb_key, &mdb_val_meta, 0); result != MDB_SUCCESS)
-        LMDB_THROW("Failed to write metadata", result);
-
-    txn.commit();
+    // LMDB map size limits are based on committed pages, not just active payload size.
+    // Eviction should be triggered based on committed size to avoid MDB_MAP_FULL before
```
ccummingsNV: This should be in a separate PR, and reviewed by @skallweitNV

slangpy/builtin/valueref.py
```diff
@@ -223,6 +226,8 @@ def read_calldata(
     def create_output(self, context: CallContext, binding: BoundVariableRuntime) -> Any:
         pt = slang_type_to_return_type(self.value_type)
         if pt is not None:
+            # For arrays this will create an empty list; the actual value will be filled during
```
ccummingsNV: Suggest this is redundant comment (unless you want to specify what happens for other types)

slangpy/builtin/valueref.py
```diff
@@ -170,15 +170,18 @@ def create_calldata(
...
                     element_count=1,
                     struct_size=binding.vector_type.buffer_layout.stride,
                     usage=BufferUsage.shader_resource | BufferUsage.unordered_access,
                 )
                 cursor = BufferCursor(binding.vector_type.buffer_layout.reflection, buffer, False)
-                cursor[0].write(data.value)
-                cursor.apply()
+                # For write-only outputs, the initial Python value may be just a placeholder
+                # (e.g. [] for half[8]). Don't pre-initialize GPU memory in that case.
+                if access[0] == AccessType.readwrite:
```
ccummingsNV: Suggest access[0] != AccessType.write 
That way you have 2 advantages:
- if enum changes in future this line is less likely to become wrong,
- the functionality of this 'if' is not dependent on that on line 170
Both less likely to create future bugs

slangpy/builtin/value.py
```diff
@@ -60,6 +60,9 @@ def slang_type_to_return_type(slang_type: kfr.SlangType) -> Any:
             return getattr(slangpy, f"uint{slang_type.rows}x{slang_type.cols}")
         elif slang_type.slang_scalar_type in BOOL_TYPES:
             return getattr(slangpy, f"bool{slang_type.rows}x{slang_type.cols}")
+    elif isinstance(slang_type, kfr.ArrayType):
+        # Arrays are represented as nested Python lists when returned as a value.
```
ccummingsNV: Suggest this is redundant comment

slangpy/tests/slangpy_tests/test_array_return_arrays.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+
+
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_return_array_value_from_python_list(device_type: DeviceType):
+    device = helpers.get_device(device_type)
+    module = helpers.create_module(device, MODULE)
+
+    # Scalar call (no vectorization): return type is half[8] as a Python value.
+    func = module.find_function("tensor_test_channels<8>")
```
ccummingsNV, line 33: Does this work if you rely on slangpy to do the generic resolution (i.e. just find 'tensor_test_channels' and pass it an array of length 8). No problem if not, but if it does, would be good to add a test to verify that stays working.

ccummingsNV (changes requested):
Few suggested changes with code maintenance / other engineers in mind, but looks good. 

The lmdb work should be in a separate PR though as far as I can tell.

src/sgl/core/lmdb_cache.cpp
skallweitNV: Agreed this should be a separate PR. I'll have to review this in more detail, but overall looks like the right direction.

NV-xiaoyongs: removed

slangpy/builtin/valueref.py
NV-xiaoyongs: removed

slangpy/builtin/valueref.py
NV-xiaoyongs: updated

slangpy/builtin/value.py
NV-xiaoyongs: removed

slangpy/tests/slangpy_tests/test_array_return_arrays.py
NV-xiaoyongs: Confirmed: generic resolution works. Calling tensor_test_channels (no <8>) with a Python list of length 8 resolves NumChannels = 8. I added a regression test (test_return_array_value_generic_resolution) to ensure this stays working.

src/sgl/core/lmdb_cache.cpp
skallweitNV: @NV-xiaoyongs will you open a new PR for the lmdb cache bug? Would be greatly appreciated.

slangpy/tests/slangpy_tests/test_array_return_arrays.py
ccummingsNV: Thanks!

src/sgl/core/lmdb_cache.cpp
skallweitNV: no need anymore .. opened #739

Issue #526, opened 2025-09-20
Missing `KeyCode` enum entries for `GLFW_KEY_WORLD_1` / `GLFW_KEY_WORLD_2`


Gegell:
On QWERTZ keyboards the first maps to the less/greater/pipe key.
Their values are 161 and 162 respectively, and as far as I can see these are missing from:
https://github.com/shader-slang/slangpy/blob/b53036dfd40e7572d3fdce53946d53f1c0bfefe0/src/sgl/core/input.h#L90-L91
Similarly its then missing from the python wrapper enum `KeyCode(enum.Enum)`.
Subsequently trying to access the key in the `KeyboardEvent` as in:
```py
window = spy.Window()
window.on_keyboard_event = lambda evt: print(evt.key)
```
raises a `ValueError: 162 is not a valid KeyCode.`

PR #677, opened 2026-01-06
Fix KeyCode WORLD_1/WORLD_2 in Python bindings

NV-xiaoyongs:
Issue: #526 

Add KeyCode entries for GLFW_KEY_WORLD_1/2 (161/162) and register them in SGL_ENUM_INFO so Python can materialize these keys without ValueError.

Add a regression test covering KeyCode(161/162).


tests/sgl/core/test_lmdb_cache.cpp
```diff
@@ -415,6 +415,9 @@ TEST_CASE("stress-single-threaded")
 {
     StressTest::Options options;
     options.path = testing::get_case_temp_directory() / "cache";
+    // The stress test can write a lot of data (values up to 64KB across thousands of candidates).
```
skallweitNV: I've seen the issue on macos runners in CI, but this change sort of defeats the purpose of the test. We want to make sure the cache keeps working when being close to capacity.

For now, I would prefer skipping the test on macos instead and figure out why it fails (because I've not seen it fail on other machine types).

Can we remove this change from this PR?

skallweitNV: Please just remove these changes. There is already a first stab at a proper fix in https://github.com/shader-slang/slangpy/pull/676

NV-xiaoyongs: Thank you for reviewing. The test_lmdb_cache changes have been removed.

skallweitNV: Thanks!

Issue #535, opened 2025-09-30
Support torch.nn in slangpy


mkeshavaNV:
slangtorch seems to support using `torch.nn` directly. This is a placeholder task to add this support in slangpy.

jkwak-work:
Re-assigning to @NV-xiaoyongs

PR #679, opened 2026-01-07
Support torch.nn in slangpy

wlewNV:
Issue: #535


ccummingsNV:
Thanks @wlewNV 

I think let's just hold off on this for a week or so, as our upcoming optimizations will probably intersect it a lot. We'll keep the PR here as all the tests will be valid, but I need to rewrite how the unpacking/signature checking works so it doesn't go via nanobind (this is apparently extremely expensive, especially on linux). 

Once I start optimizing the tensor marshalling path, I'll merge your tests to my local branch and make sure they're accounted for as part of the work.

PR #697, opened 2026-01-09
Tensor refactor

ccummingsNV:


docs/src/basics/tensors.rst
```diff
@@ -1,19 +1,22 @@
-Buffers
+Tensors
```
skallweitNV, line 1: Should this doc be renamed to `tensors.rst`?

docs/src/basics/buffers.rst
```diff
@@ -1,19 +1,22 @@
...
+- **Stride**: A tuple of integers describing the layout of each dimension. By default, ``Tensor`` uses a row-major layout, where the right-most dimension has the smallest stride.
 
 .. warning::
-   **Buffer Indexing Conventions**
+   **Indexing Conventions**
 
-   Multi-dimensional buffers store data using the convention where the right-most dimension has the smallest stride.
-   However, buffers can be indexed using either array coordinates, which follow the same convention, or vector coordinates, which follow a different convention where the x component has the smallest stride.
-   This means the same buffer position requires different coordinate values: e.g. for a 2D buffer, array indexing uses `[row, col]` while vector indexing uses `(col, row)` for the same location.
+   Multi-dimensional tensors store data using the convention where the right-most dimension has the smallest stride.
```
skallweitNV: pretty much a repetition of the comment above on `**Stride**`

docs/src/tensors/differentiable.rst
```diff
@@ -0,0 +1,343 @@
...
+    device = spy.create_device()
+    module = spy.Module.load_from_file(device, "example.slang")
+
+    # Create input tensor with gradients (automatically zeroed)
+    x = spy.Tensor.from_numpy(device, np.array([1, 2, 3, 4], dtype=np.float32))
+    x = x.with_grads(zero=True)
+
+    # Forward pass: evaluate polynomial
+    # Result: 2*x^2 + 8*x - 1
+    module.polynomial(idx=spy.grid(x.shape), a=2, b=8, c=-1, x=x, result=result)
```
skallweitNV, line 71: `result` is undefined here

docs/src/tensors/differentiable.rst
```diff
@@ -0,0 +1,343 @@
...
+1. Create input tensors with gradient storage
+2. Run the forward pass
+3. Attach gradients to outputs and initialize them
+4. Run the backward pass with ``.bwds()``
+5. Read accumulated gradients from input tensors
+
+Why Use Interface Types?
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the none-differentiable case, the interface types (``ITensor``, ``IWTensor``, ``IRWTensor``) are recommended for maximum flexibility (and in future, performance), but they are not critical. However for
```
skallweitNV: non-differentiable

docs/tensorupdate.rst
```diff
@@ -0,0 +1,81 @@
+.. _tensorupdate:
+
+Migration guide for Tensor update
+=================================
+
+In SlangPy version X we undertook a major rewrite of the Tensor API to bring it inline with existing standards and simplify its use going forwards. The primary changes made are:
```
skallweitNV: replace `X`

docs/tensorupdate.rst
```diff
@@ -0,0 +1,81 @@
...
+
+2. Replace differentiable Tensors parameters in Slang
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have exclusively used the Tensor type as function parameters, simply search and replacing as below should be sufficient.
+
+- Replace all instances of ``Tensor`` with ``IDiffTensor``
+- Replace all instances of ``RWTensor`` with ``IRWDiffTensor``
+- Replace all instances of ``GradInTensor`` with ``IWDiffTensor``
+- Replace all instances of ``GradOutTensor`` with ``IDiffTensor``
```
skallweitNV, line 44: is this swapped?

slangpy/slang/core.slang
```diff
@@ -99,17 +99,17 @@ public struct NoneType
 #ifdef __TARGET_CUDA__
     int dummy;
 #endif
-    public void load<T, let N : int>(ContextND<N> context, out T value) { value = T();}
-    public void store<T, let N : int>(ContextND<N> context, in T value) {};
+    public void __slangpy_load<T, let N : int>(ContextND<N> context, out T value) { value = T();}
```
skallweitNV, line 102: Why `__` instead of `_` ?

slangpy/slang/tensor_indices_each.slang
```diff
@@ -0,0 +1,205 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
skallweitNV, line 1: Is this kept intentionally?

slangpy/tests/slangpy_tests/fusetest.slang
```diff
@@ -0,0 +1 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
skallweitNV: There is a bunch of these empty files. Should these just be removed?

slangpy/slang/tensor.slang
```diff
@@ -1,473 +1,479 @@
...
+    public void store<I : __BuiltinIntegerType>(I idx[D], T value)                                                     \
+    {                                                                                                                  \
+        write_buffer(_idx(idx, _strides, _offset), value);                                                             \
+    }                                                                                                                  \
+    public void store<I : __BuiltinIntegerType>(vector<I, D> idx, T value)                                             \
+    {                                                                                                                  \
+        write_buffer(_idx(idx, _strides, _offset), value);                                                             \
+    }                                                                                                                  \
+    void _write_each<each I : __BuiltinIntegerType>(T value, expand each I idx)                                 \
```
skallweitNV: Looks like these files could use formatting.

skallweitNV: I wonder if most of these should be `[ForceInline]` to help Slang output better code?

slangpy/slang/tensor_indices_generated.slang
```diff
@@ -0,0 +1,1101 @@
...
+
+public extension<T: IDifferentiable, TensorType : IWDiffTensor<T, 4>> TensorType where T.Differential : IAtomicAddable
+{
+    [Differentiable]
+    [BackwardDerivative(_store_bwd_indices)]
+    public void store(int i0, int i1, int i2, int i3, T value)
+    {
+        this._write_primal_each(value, i0, i1, i2, i3);
+    }
+    void _store_bwd_indices(int i0, int i1, int i2, int i3, inout DifferentialPair<T> grad)
```
skallweitNV, line 680: if we have `store` shouldn't the bwd version be called `store_bwd` or `_store_bwd`?

slangpy/slang/tensor_indices_generated.slang
```diff
@@ -0,0 +1,1101 @@
...
+// Tensor index extensions generated by tools/generate_tensors.py
+implementing slangpy;
+
+public extension<T, TensorType : ITensor<T, 0>> TensorType
+{
+    public T load()
+    {
+        return this._read_each();
+    }
+    public __subscript()->T
```
skallweitNV, line 12: is this valid code?

slangpy/slang/tensor_indices_generated.slang
```diff
@@ -0,0 +1,1101 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Tensor index extensions generated by tools/generate_tensors.py
```
skallweitNV, line 2: I guess this exist because there is no way to write these in a generic way using variadics?

skallweitNV:
It's probably worth going over the unit test logs and fix deprecation warnings etc.

slangpy/slang/tensor.slang
```diff
@@ -1,473 +1,479 @@
...
-                value[i][j] = get(idx);
-            }
-        }
+        get;
+        set;
     }
-
-    [Differentiable]
-    public void load<let M : int, let N : int>(ContextND<D - 2> context, out matrix <T, M, N> value)
+    public __subscript<I : __BuiltinIntegerType>(vector<I, D> indices)->T
```
csyonghe, line 71: the problem of having both arrays and vector typed overloads is that if the user do `tensor[{x,y,z}]`, we may get an overload resolution error. Being able to `tensor[{x,y,z}]` is the closest thing you can get without relying on variadics, and I'd like us to preserve this syntax. 

My preference would be for the vector case to be just a normal `load()` method.

slangpy/slang/tensor.slang
```diff
@@ -1,473 +1,479 @@
...
-
-    [TreatAsDifferentiable]
-    public T getv(vector<int,D> idx) { return buffer[layout.at_vec(idx)]; }
-
-    public void load<let SliceD : int>(ContextND<D - SliceD> ctx, out Tensor<T, SliceD> value)
+    public void __slangpy_load(Context0D context, out This value)
+    {
+        value = this;
+    }
+    public void __slangpy_load<let SliceD : int>(ContextND<D - SliceD> ctx, out Tensor<T, SliceD> value)
```
csyonghe, line 267: it these __ methods are never supposed to be called by the user, they should be made to have `internal` visiblity.

slangpy/slang/tensor.slang
ccummingsNV: That makes it more awkward for people using vectors though doesn't it? I'm really keen not to make graphics programmers 2nd class citizens with this work.

Why would you want to use tensor[{x,y,z}] when you can do tensor[x,y,z], which (at least currently) is a simple function call with no variadics at all?

slangpy/slang/tensor_indices_generated.slang
ccummingsNV: yup

slangpy/slang/tensor_indices_generated.slang
ccummingsNV: I found the bwd functions sometimes failed / caused silent errors / crashed things when the names weren't unique. As 'store' is overloaded, we would end up having multiple _store_bwd functions. So I adopted the standard across the tensors of naming the bwd versions explicilty to avoid any confusion.

slangpy/slang/tensor.slang
ccummingsNV: Didn't realise it existed in slang :) We can force inline a load of these, sure.

slangpy/tests/slangpy_tests/fusetest.slang
ccummingsNV: Yup - accidental check in

slangpy/slang/tensor_indices_each.slang
ccummingsNV: Yes - @csyonghe was considering how to do this properly. Ideally the generated version is just a stop gap until slang supports the functionality required to use generics for these functions. I kept this one around so it's easy for the slang team to switch to it and make it work.

slangpy/slang/core.slang
ccummingsNV: Seemed the standard for "this is a 'system' function"

docs/tensorupdate.rst
ccummingsNV: Nope! 'GradInTensor' would typically have mapped to a writable tensor, because you'd be reading gradients if you were writing primals.

slangpy/slang/tensor.slang
ccummingsNV: the problem is they are supposed to be called by the generated kernel, which isn't inside the slangpy module.

slangpy/slang/tensor_indices_generated.slang
ccummingsNV: I believe so. @csyonghe ?

PR #702, opened 2026-01-13
Reduce CUDA event overhead by creating events every 32 submissions instead of every submission

szihs:


szihs:
https://github.com/shader-slang/slangpy/pull/747 should fix it properly. So closing this one.

PR #707, opened 2026-01-13
Block allocator for slangpy

ccummingsNV:
Adds block allocator similar to slang-rhi and uses for shader object.

NOTE: We need to decide what to do RE nanobind to get this in.


szihs:
Nit : consider adding memory usage tracking for debugging purposes.

PR #708, opened 2026-01-13
Use short vector for shader object refs

ccummingsNV:
ShaderObject was using std::set to store lists of other objects / interop buffers, however no deduplication was ever performed, as every entry added to the set is new. Switched to short_vector so most use is zero alloc.


ccummingsNV:
@szihs 
short vector issue here: https://github.com/shader-slang/slangpy/issues/727

PR #712, opened 2026-01-13
Additions to shader object + shader cursor to support tensor optimizations

ccummingsNV:


src/sgl/device/shader_cursor.h
```diff
@@ -72,6 +78,8 @@ class SGL_API ShaderCursor : public CursorWriteWrappers<ShaderCursor, ShaderOffs
 
     void set_data(const void* data, size_t size) const;
 
+    void* reserve_data(size_t size) const;
```
skallweitNV, line 82: We should have the same warning comment here as we added in slang-rhi.

src/sgl/device/shader_cursor.cpp
```diff
@@ -224,6 +224,63 @@ ShaderCursor ShaderCursor::find_field(std::string_view name) const
...
 }
 
+int32_t ShaderCursor::find_field_index(std::string_view name) const
+{
+    if (!is_valid())
+        return -1;
+
+    slang::TypeLayoutReflection* type_layout = m_type_layout;
+
+    // For constant buffers and parameter blocks, get the element type layout directly
```
szihs, line 234: Is it safe to skip dereference() for constant buffers and parameter block to create a new shaderCursor object?

src/sgl/device/shader_cursor.cpp
```diff
@@ -459,6 +516,15 @@ void ShaderCursor::set_data(const void* data, size_t size) const
     m_shader_object->set_data(m_offset, data, size);
 }
 
+void* ShaderCursor::reserve_data(size_t size) const
+{
+    if ((TypeReflection::ParameterCategory)m_type_layout->getParameterCategory()
+        != TypeReflection::ParameterCategory::uniform)
+        SGL_THROW("\"{}\" cannot reserve data", m_type_layout->getName());
+    return m_shader_object->reserve_data(m_offset, size);
```
szihs, line 524: we should add some size validation to disallow reserving data greater than type size.

szihs (commented):
Allows Tensor optimizations by direct memory writes.

src/sgl/device/shader_cursor.cpp
ccummingsNV: It wouldn't be safe to skip it for creation of a new shader cursor, as it requires a new shader object. However for simply doing a field lookup it is fine.

src/sgl/device/shader_cursor.cpp
ccummingsNV: This is handled at the rhi layer

slangpy/tests/device/test_shader_cursor.py
```diff
@@ -365,5 +365,75 @@ def write_vars(
         assert named_typed_result == named_typed_reference
 
 
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_shader_cursor_field_by_index(device_type: spy.DeviceType):
+    """Test find_field_index and get_field_by_index methods."""
+    if device_type == spy.DeviceType.vulkan and sys.platform == "darwin":
+        pytest.skip("Test shader doesn't currently compile on MoltenVK")
```
skallweitNV: You can probably remove this, we don't run on MoltenVK anymore

src/sgl/device/shader_cursor.cpp
skallweitNV: LGTM

skallweitNV (approved):
LGTM, comment on `reserve_data` is still missing though.

PR #713, opened 2026-01-14
Fix float3 alignment bug on Metal for gradient accumulation

### LINKED ISSUE #118: The memory layout for a tensor is different from it's gradient tensor
In test slangpy/tests/test_differential_function_call.py, we notice that the memory layout of the input tensor is different from it's gradient tensor.
This task is to track this issue for further investigation.

Marked as " # TODO: Somehow the grads are not layout the same as input tensors." in the test script.

szihs:
The issue was that AtomicTensor used sizeof(T) to compute byte offsets for RWByteAddressBuffer operations, but sizeof(float3) returns 12 while Metal buffers use 16-byte stride for float3.

Changes:
- Add stride-aware atomic methods (atomicAddWithStride, etc.) in atomics.slang
- Add _element_byte_stride field to AtomicTensor in tensor.slang
- Pass element_stride from C++ when binding tensors in slangpytensor.cpp
- Update tests to use proper tensor reads now that alignment is fixed

Fixes #118

### DIFF
--- a/slangpy/tests/utils/test_texture_loader.py
+++ b/slangpy/tests/utils/test_texture_loader.py
@@ -302,5 +302,102 @@ def test_load_texture_array(device_type: spy.DeviceType):
     assert texture.array_length == 2
 
 
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_ya_handling_default_expands_to_rgba(device_type: spy.DeviceType):
+    """Test that default YA handling expands to RGBA (Y to RGB, A to A)."""
+    device = helpers.get_device(type=device_type)
+
+    bitmap = Bitmap(
+        pixel_format=PixelFormat.ya,
+        component_type=ComponentType.uint8,
+        width=32,
+        height=16,
+    )
+    a = np.array(bitmap, copy=False)
+    a[:, :, 0] = 100  # Y (luminance)
+    a[:, :, 1] = 200  # A (alpha)
+
+    loader = TextureLoader(device)
+    options = TextureLoader.Options()
+    options.load_as_normalized = True
+    options.load_as_srgb = False
+    texture = loader.load_texture(bitmap=bitmap, options=options)
+
+    assert texture.format == Format.rgba8_unorm
+
+    data = texture.to_numpy()
+    assert data.shape == (16, 32, 4)
+    assert np.allclose(data[:, :, 0], 100, atol=1)  # R = Y
+    assert np.allclose(data[:, :, 1], 100, atol=1)  # G = Y
+    assert np.allclose(data[:, :, 2], 100, atol=1)  # B = Y
+    assert np.allclose(data[:, :, 3], 200, atol=1)  # A = A
+
+
+@pytest.mark.parametrize("device_type", helpers.DEFAULT_DEVICE_TYPES)
+def test_ya_handling_preserve_as_rg(device_type: spy.DeviceType):
+    """Test that preserve_as_rg keeps YA as 2-channel RG (Y to R, A to G)."""
+    device = helpers.get_device(type=device_type)
+
+    format_support = device.get_format_support(Format.rg8_unorm)
+    if not FormatSupport.shader_load in format_support:
+        pytest.skip("RG format not supported")
+
+    bitmap = Bitmap(
+        pixel_format=PixelFormat.ya,
+        component_type=ComponentType.uint8,
+        width=48,
+        height=24,
+    )
+    a = np.array(bitmap, copy=False)
+    a[:, :, 0] = 100  # Y (luminance)
+    a[:, :, 1] = 200  # A (alpha)
+
+    loader = TextureLoader(device)
+    options = TextureLoader.Options()
+    options.load_as_normalized = True
+    options.load_as_srgb = False
+    options.ya_handling = spy.YAHandling.preserve_as_rg
+    texture = loader.load_texture(bitmap=bitmap, options=options)
+
+    assert texture.format == Format.rg8_unorm
+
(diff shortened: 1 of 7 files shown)


ccummingsNV:
This is a neat and cleanly implemented work around, but it comes with a performance cost both to CPU and GPU, and doesn't help users who want to write atomic accumulators for structs.

@csyonghe @bmillsNV the correct fix here seems to me to be an 'alignedSizeOf' in slang (or simply to make metal return the correct size - depends on what you argue is correct / useful). Is it practical to get that in? I'd rather we do this the right way than with a work around.

If we can agree that is the correct fix, I'm happy to merge tensor work in with that test disabled on Mac, provided we fix it before launching the tensor refactor a few weeks from now.

slangpy/slang/atomics.slang
```diff
@@ -108,11 +108,20 @@ public extension<T: IAtomicAddable> StorageTraits<T>
     public static void atomicAdd(AtomicBufferType buffer, uint idx, T value) { T::atomicAdd(buffer, idx, value); }
     public static T atomicLoad(AtomicBufferType buffer, uint idx) { return buffer[idx]; }
     public static void atomicStore(AtomicBufferType buffer, uint idx, T value) { buffer[idx] = value; }
+    // Stride-aware versions (stride parameter ignored on CUDA since it uses element indexing)
```
ccummingsNV, line 108: The none-stride versions should call the strided versions, so we don't implement the same logic twice

slangpy/slang/tensor.slang
```diff
@@ -365,17 +365,22 @@ public struct AtomicTensor<T: IAtomicAddable, let D : int> : IRWTensor<T, D>
     // Underlying data storage
     public StorageTraits<T>::AtomicBufferType _data;
 
+    // Element byte stride - the actual stride in bytes between elements in the buffer.
+    // This is passed from C++ and may differ from sizeof(T) due to alignment requirements.
+    // For example, on Metal, sizeof(float3)=12 but buffer stride=16.
+    public uint _element_byte_stride;
```
ccummingsNV, line 371: On none-metal platforms, make this a static const

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -401,6 +401,15 @@ void NativeTensorMarshall::write_shader_cursor_fields(
...
     );
     field["_offset"] = buffer->offset();
+
+    // Write element byte stride if the field exists (for AtomicTensor).
+    // This is needed because sizeof(T) in shader code may differ from the actual
+    // buffer element stride due to alignment requirements (e.g., sizeof(float3)=12
+    // on Metal but buffer stride=16).
+    ShaderCursor element_byte_stride_field = field.find_field("_element_byte_stride");
+    if (element_byte_stride_field.is_valid()) {
+        element_byte_stride_field = static_cast<uint32_t>(buffer->element_stride());
```
ccummingsNV: I think 'element_stride' may be slow (a call to python) or unreliable (not always valid). Worth checking it is always correct for tensors and native only.

slangpy/slang/atomics.slang
szihs: updated.

slangpy/slang/tensor.slang
szihs: done.

src/slangpy_ext/utils/slangpytensor.cpp
szihs: I will check this and revert.

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -642,6 +651,15 @@ void NativeTensorMarshall::write_tensor_fields_from_pointer(
         offsets.offset.uniform_offset - m_cached_offsets.field_offset.uniform_offset,
         offset
     );
+
+    if (offsets.element_byte_stride.is_valid()) {
```
ccummingsNV: I merged in latest from my uniform optimizations and uncovered a bug here - you provide the element byte stride to the tensor (previously if on mac, now 'if the field exists') but only if it comes from a Buffer. If the user passes a torch tensor in, we'd need to extract the element byte size from it and follow the same pattern.

ccummingsNV: not entirely sure why mac tests are actually passing though! maybe we don't have torch tests running on it!

ccummingsNV: (confirmed - torch tests not running on mac)

szihs: Yes, updated accordingly.

src/slangpy_ext/utils/slangpytensor.cpp
szihs: Updated to use 
`buffer->desc().struct_size `which is a native C++ field

ccummingsNV (changes requested):
Happy with the work, though it probably needs a bit of merging with the latest torch integration, as a lot of torch tensor marshall code is now native

szihs:
@ccummingsNV - Thanks , I have updated to use the latest torch tensor integration.

ccummingsNV (approved):
Assuming tests pass, looks good to me

fknfilewalker:
I think the better solution would be to use `packed_` versions for all vectors when used within a StructuredBuffer so that functionality is aligned with other backends. This transformation could be done during lowering.

[https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=34](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=34)

PR #723, opened 2026-01-15
Support reinterpreting torch.Tensor as Tensor<StructType, N>

### LINKED ISSUE #522: Support `reinterpret` of a torchTensor as `Tensor<MyStructType, N>`
This is particularly useful for porting slang-torch code over, and for developers who wish to have more structure in the Slang code that interprets the last dimension of a tensor in a named, meaningful way.

We can warn that this may lead to platform-dependent behavior because data layout can change between different platforms.

The user is at their own risk when using this feature.

wlewNV:
Fixes #522 

**Note: This change is built on-top of the tensor refactor. Once that change has been merged, this change can be moved from draft.**

Add as_struct_tensor() helper to reinterpret a torch.Tensor as a tensor of struct elements, enabling more structured Slang code when porting from slang-torch.

Changes:
- Add StructTensorRefMarshall for struct tensor marshalling
- Support both CUDA (fast path) and non-CUDA backends (interop buffer)

Note: As stated in the parent issue, struct layout may vary between platforms due to padding/alignment.


ccummingsNV:
Hi - can we hold off on this until the tensor refactor perf work is done. It's a great feature, but marshalling of tensors is going to change dramatically

ccummingsNV:
@wlewNV this needs merging or closing :)

PR #737, opened 2026-01-15
Add format_callback option to TextureLoader for texture conversion

### LINKED ISSUE #599: Generalize the Bitmap->Texture conversion logic
Ref: https://github.com/shader-slang/slangpy/pull/588

When presented with `ya` Bitmap, there are two options. One is to threat the input as image that should be loaded as in photoshop, i.e., present the used with `rgb` == `y` and `a` in the `alpha` channel. This was, the behavior on the reading Slang side can be the same no matter the whether the input `png` is `rgba` or `ya`. However, this is wasteful.

The other option is to load this as `rg` texture, but then the reader needs to know from some metadata that the texture is a 2 channel one and value in `g` needs to be copied to `a` while value in `r` needs to be copied into channels `gb`. This does not seem doable with just the information that is on the standard `Texture2D` object.

As both options have significant merits, the control on which way should the loader behave should be configurable with a wider range of options than we have right now.

wlewNV:
Fixes #599

Adds a configurable callback mechanism for handling bitmap formats that don't have a direct mapping in the standard FORMAT_TABLE (e.g., YA/greyscale+alpha bitmaps).

When loading YA (greyscale+alpha) bitmaps, there are two options:
1. Expand to RGBA (wastes memory but shader code is consistent)
2. Keep as RG (efficient but shader must interpret R=luminance, G=alpha)

Previously, slangpy always expanded YA to RGBA with no way to customize.

This change adds an optional `format_callback` to TextureLoader.Options that is invoked when the standard format table doesn't have a mapping. Users can return a FormatOverride to specify the desired format, or None to use the default behavior.

Before
```
# YA bitmap always expands to RGBA (4 channels, wasteful)
loader = TextureLoader(device)
texture = loader.load_texture(ya_bitmap)
```

After:
```
def ya_to_rg(device, bitmap):
    if bitmap.pixel_format == PixelFormat.ya:
        return FormatOverride(format=Format.rg8_unorm, convert_to=None)
    return None  # Default for other formats

loader = TextureLoader(device)
options = TextureLoader.Options()
options.format_callback = ya_to_rg
texture = loader.load_texture(ya_bitmap, options)
```


src/sgl/utils/texture_loader.h
```diff
@@ -35,6 +47,8 @@ class SGL_API TextureLoader : public sgl::Object {
         /// Resource usage flags for the texture.
         /// \c TextureUsage::render_target will be added automatically if \c generate_mips is true.
         TextureUsage usage{TextureUsage::shader_resource};
+        /// Strategy for handling YA (greyscale + alpha) bitmaps.
```
skallweitNV: Can we move this to below `extend_alpha` so all the options for conversion are nearby.

ccummingsNV:
@wlewNV do you consider this ready to merge? I have updated it to latest + approved if so.

ccummingsNV (approved):

PR #749, opened 2026-01-22
add slangPy neural slang integarion tests

NV-xiaoyongs:
WIP

add neural slang integration tests

### DIFF
--- a/slangpy/tests/slangpy_tests/test_neural_bindless.py
+++ b/slangpy/tests/slangpy_tests/test_neural_bindless.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+Neural integration tests for bindless resource types.
+
+Reviewer-requested coverage:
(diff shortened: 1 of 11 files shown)


NV-xiaoyongs:
> > > Thought here - does neural.slang have CI? I think we should be shifting to a model where 'integration tests' run in the context of the project being integrated. They can still be triggered by slangpy CI or nightly runs if need be, but we should avoid them living in the slangpy repo.
> > 
> > 
> > Thanks! At the moment there’s no CI for neural.slang. I’m still becoming familiar with the directory structure—can you advise on the right location to place these neural.slang files?
> 
> My point, I think, is that slangpy is probably not the right place for neural.slang files. Where is neural.slang actually coming from? I don't see it checked in.

Thanks for the clarification. Yeah, we might need to move it to slang. The neural.slang PRs are under review as well.

https://github.com/shader-slang/slang/pull/9512
https://github.com/shader-slang/slang/pull/9713

NV-xiaoyongs:
move to: https://github.com/shader-slang/slang/pull/9825

PR #759, opened 2026-01-27
PyTorch optimizations

ccummingsNV:


src/slangpy_ext/utils/torch_bridge.h
```diff
@@ -0,0 +1,166 @@
...
+    /// Returns nullptr if not a tensor or bridge unavailable
+    /// Format: "[torch,Dn,Sm]" where n=ndim, m=scalar_type
+    int get_signature(PyObject* obj, char* buffer, size_t buffer_size) const
+    {
+        if (!m_api) {
+            snprintf(buffer, buffer_size, "slangpy_torch not available");
+            return -1;
+        }
+        if (m_api->get_signature(obj, buffer, buffer_size) != 0) {
+            snprintf(buffer, buffer_size, "Failed to get signature");
```
szihs: Currently it is returning success even if m_api->get_signature(obj, buffer, buffer_size) != 0 fails.

Proposed fix   
 return m_api->get_signature(obj, buffer, buffer_size);  // Propagate result

slangpy/torchintegration/autogradhook.py
```diff
@@ -1,177 +1,174 @@
...
+        # Return any output tensors plus the result if present
+        res = tuple(outputs)
+        if result is not None:
+            assert isinstance(result, torch.Tensor)
+            if not "_result" in kwargs:
+                pair = NativeTorchTensorDiffPair(None, None, len(pairs), False)
+                kwargs["_result"] = pair
+                ctx.pairs.append(pair)
+            res += (result,)
+        return res
```
szihs, line 113: When torch autograd is enabled, the autograd wrapper returns output tensors collected from pairs and also appends result returned by forwards_cd.call(...). If the user provided _result (or an out-parameter tensor), that same tensor is both:
`res = (my_output_tensor, my_output_tensor)`


# Proposed code:
```
res = tuple(outputs)
if result is not None:
    assert isinstance(result, torch.Tensor)
    if "_result" not in kwargs:
        pair = NativeTorchTensorDiffPair(None, None, len(pairs), False)
        kwargs["_result"] = pair
        ctx.pairs.append(pair)
        res += (result,)  # Only append if we created a new pair
    # else: _result was user-provided and already in outputs via pairs
return res
```

src/slangpy_torch/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+[build-system]
+# Note: torch is NOT listed here because this package MUST be installed with
+# --no-build-isolation to ensure ABI compatibility with the user's PyTorch.
+# The user must have a version of torch installed before use.
+requires = ["setuptools>=64"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "slangpy-torch"
+version = "0.4.0"
```
skallweitNV: why is this 0.4.0? is this current slangpy version? how do we version this in the future?

src/slangpy_torch/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```bash
+pip install slangpy-torch --no-build-isolation
+```
+
+### Verify Installation
+
+```python
+import torch # Before slangpy_torch
+import slangpy_torch
```
skallweitNV, line 36: is this really needed? any way around it?

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+#include <cstdint>
+#include <cstddef>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Default inline storage size for common cases (covers 99%+ of tensors)
+// Callers needing more dimensions can provide larger buffers
+#define TENSOR_BRIDGE_DEFAULT_DIMS 16
```
skallweitNV, line 28: leftover?

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+//   py_tensor_obj: PyObject* that must be a torch.Tensor
+//   out: Output structure to populate. shape/strides pointers will be set
+//        to shape_buffer/strides_buffer if buffer_capacity >= ndim
+//   shape_buffer: Caller-provided buffer for shape data (may be NULL)
+//   strides_buffer: Caller-provided buffer for strides data (may be NULL)
+//   buffer_capacity: Number of elements the buffers can hold
+// Returns 0 on success, non-zero on error
+// Note: out->ndim is always set. If ndim > buffer_capacity, shape/strides
+//       will be nullptr and caller should retry with larger buffers.
+typedef int (*TensorBridge_ExtractFn)(
```
skallweitNV, line 111: i think defining a result code would be helpful, instead of just an int
errors are not at all specified right now

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+
+// Check if a PyObject* is a torch.Tensor
+// Returns 1 if true, 0 if false
+typedef int (*TensorBridge_IsTensorFn)(void* py_tensor_obj);
+
+// Get a minimal signature for a PyObject* if it's a torch.Tensor
+// Returns: pointer to static buffer with signature string if tensor, NULL if not a tensor
+// Format: "[torch,Dn,Sm]" where n=ndim, m=scalar_type
+// This is faster than full extraction when only signature is needed (~15ns)
+typedef int (*TensorBridge_GetSignatureFn)(void* py_tensor_obj, char* buffer, size_t buffer_size);
```
skallweitNV, line 131: what is the `int` return code?

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+
+// Get the last error message (if any function returned non-zero)
+// Returns a pointer to a static thread-local buffer
+typedef const char* (*TensorBridge_GetErrorFn)(void);
```
skallweitNV: does this also appliy o `IsTensor`? that one returns returns non-zero when there is no error it seems like
just raising this because the comment mentions "if any function returned non-zero"

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+// Returns the cudaStream_t pointer, or nullptr if CUDA is not available
+// device_index: the CUDA device index (0, 1, 2, etc.)
+typedef void* (*TensorBridge_GetCurrentCudaStreamFn)(int device_index);
+
+// Copy tensor data to a contiguous CUDA buffer
+// This handles non-contiguous tensors by using PyTorch's copy mechanism.
+// dest_cuda_ptr: destination CUDA pointer (e.g., from interop buffer mapped memory)
+// dest_size: size in bytes of destination buffer
+// Returns 0 on success, non-zero on error
+typedef int (*TensorBridge_CopyToBufferFn)(void* py_tensor_obj, void* dest_cuda_ptr, size_t dest_size);
```
skallweitNV, line 143: are there any layout rules for the tensor in  `dest_cuda_ptr`?

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+
+// Copy data from a contiguous CUDA buffer back to a tensor
+// This handles non-contiguous tensors by using PyTorch's copy mechanism.
+// src_cuda_ptr: source CUDA pointer (e.g., from interop buffer mapped memory)
+// src_size: size in bytes of source buffer
+// Returns 0 on success, non-zero on error
+typedef int (*TensorBridge_CopyFromBufferFn)(void* py_tensor_obj, void* src_cuda_ptr, size_t src_size);
```
skallweitNV, line 150: same question

src/slangpy_torch/tensor_bridge_api.h
```diff
@@ -0,0 +1,164 @@
...
+    // Element size in bytes
+    int32_t element_size;
+
+    // Total number of elements
+    int64_t numel;
+
+    // Storage offset (for views)
+    int64_t storage_offset;
+
+    // CUDA stream (cudaStream_t, or nullptr for CPU tensors)
```
skallweitNV: maybe clarify that this is just the current cuda stream at the time the info is queried

tools/ci.py
```diff
@@ -227,6 +227,21 @@ def coverage_report(args: Any):
...
 
 
+def install_slangpy_torch(args: Any):
+    """Install the slangpy-torch extension from source."""
+    slangpy_torch_dir = PROJECT_DIR / "src" / "slangpy_torch"
+    if not slangpy_torch_dir.exists():
+        print(f"slangpy_torch directory not found: {slangpy_torch_dir}")
+        return
+
+    cmd = [sys.executable, "-m", "pip", "install", "wheel"]
```
skallweitNV, line 245: why do we have to manually install `wheel` here?

tools/ci.py
```diff
@@ -227,6 +227,21 @@ def coverage_report(args: Any):
...
+    run_command(cmd)
+
+    # Use --no-build-isolation to compile against the user's installed PyTorch
+    cmd = [sys.executable, "-m", "pip", "install", str(slangpy_torch_dir), "--no-build-isolation"]
```
skallweitNV, line 249: what happens after the CI run? is this just installed in the environment?

.gitignore
```diff
@@ -36,4 +36,9 @@ slangpy/tests/slangpy_tests/logs/
 
 sgl_tests.log
 
+src/slangpy_torch/*.egg-info/
```
skallweitNV: maybe better to put these into a `.gitignore` file in `src/slangpy_torch`
this will make it easier to separate that folder into a separate repo if need be

.github/workflows/ci.yml
```diff
@@ -155,6 +155,11 @@ jobs:
       - name: Build
         run: python tools/ci.py build
 
+      # Install slangpy-torch extension (requires MSVC on Windows, PyTorch installed)
+      - name: Install slangpy-torch
+        if: runner.os != 'macos' && contains(matrix.flags, 'unit-test')
+        run: python tools/ci.py install-slangpy-torch
```
skallweitNV, line 161: same question, we just keep this installed after the run? and rely on the fact that reinstalling it will update to latest version?

docs/proposals/torch_bridge_fallback.md
```diff
@@ -0,0 +1,530 @@
+# Proposal: Python Fallback for TorchBridge
```
skallweitNV: not sure we want to keep this?!

slangpy/tests/slangpy_tests/test_torchintegration.py
```diff
@@ -54,24 +54,19 @@ def compare_tensors(a: torch.Tensor, b: torch.Tensor):
 @pytest.mark.parametrize(
     "pair",
     [
-        (torch.empty((1,), dtype=torch.float32).cuda(), "D1,C2,B32,L1"),
-        (torch.empty((1,), dtype=torch.float32, requires_grad=True).cuda(), "D1,C2,B32,L1"),
-        (torch.empty((1,), dtype=torch.float16).cuda(), "D1,C2,B16,L1"),
-        (torch.empty((1,), dtype=torch.int32).cuda(), "D1,C0,B32,L1"),
-        (torch.empty((1,), dtype=torch.uint8).cuda(), "D1,C1,B8,L1"),
-        (torch.empty((1, 1, 1), dtype=torch.uint8).cuda(), "D3,C1,B8,L1"),
+        (torch.empty((1,), dtype=torch.float32).cuda(), "D1,S6"),
```
skallweitNV, line 57: just curious, what do the `S{x}` mean?

slangpy/tests/utils/test_torch_bridge.py
```diff
@@ -0,0 +1,229 @@
...
+except ImportError:
+    pytest.skip("PyTorch not installed", allow_module_level=True)
+
+if sys.platform == "darwin":
+    pytest.skip("PyTorch requires CUDA, that is not available on macOS", allow_module_level=True)
+
+import slangpy
+
+
+class TestTorchBridgeAvailability:
```
skallweitNV, line 25: do we use this pattern for testing in other places?

src/slangpy_ext/utils/torch_bridge.h
```diff
@@ -0,0 +1,418 @@
...
+        return get_signature(h.ptr(), buffer, buffer_size);
+    }
+
+    /// Get last error message from native API.
+    /// @return Error message string, or empty string if using Python fallback.
+    const char* get_error() const
+    {
+        if (!m_force_python_fallback && m_api)
+            return m_api->get_error();
+        return ""; // Python fallback uses exceptions
```
skallweitNV: would it make sense to also throw exceptions when using the bridge to make error handling identical?

slangpy/core/calldata.py
```diff
@@ -3,32 +3,39 @@
 import os
 import re
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any
```
skallweitNV, line 6: changes in this file I haven't reviewed in detail ..

slangpy/tests/slangpy_tests/test_torchintegration.py
skallweitNV: found it later, just the scalar type

src/slangpy_torch/pyproject.toml
ccummingsNV: It's 0.4.0 just because its the 4th version I uploaded to pypi for testing. I think we probably in future:
- don't match version between slangpy + slangpy-torch (or we'll end up with pointless releases when we want to virtually never touch this)
- in future, build versions like "0.5.0+pt27" for 0.5.0 with wheels built for pytorch 2.7

src/slangpy_torch/README.md
ccummingsNV: yes, though users should never import slangpy_torch themselves unless they want to manually test the api.

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: cool - I'll add error codes in the next PR

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: I was undecided on how best to handle these, because I didn't want the bridge C api to fire exceptions. I was mainly populating it for when the sgl side bridge wanted to raise an exception. I'll go over the error handling next though.

tools/ci.py
ccummingsNV: good point - we should probably pip uninstall it at the end (and as a safety check, at the start)

.gitignore
ccummingsNV: Sure - didn't know you could do that!

slangpy/tests/utils/test_torch_bridge.py
ccummingsNV: We don't, but I don't really have any issue with the AI tests generating it - it's quite convenient to have them separated into categories. It lets the vscode tree view subdivide them cleanly without needing more files.

image

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: Nope - it's a copy from a tensor to a contigous buffer. The reason for the function is that the torch tensor might not be flat - it could be spread across memory, but for D3D to work with it it needs to be.

tools/ci.py
ccummingsNV: It's not installed by default in a normal ci run, and is needed to build the plugin. Remember it's building from local packages, not the isolated build env.

slangpy/tests/slangpy_tests/test_torchintegration.py
ccummingsNV: Scalar type, to replace the dlpack type codes

ccummingsNV: Scalar type, to replace the dlpack type codes

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: yup we should remove this

tools/ci.py
ccummingsNV: It's not installed by default in a normal ci run, and is needed to build the plugin. Remember it's building from local packages, not the isolated build env.

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: keeping for now as its handy for some tests/benchmarks

src/slangpy_torch/tensor_bridge_api.h
ccummingsNV: Removed it

skallweitNV (approved):
LGTM, thanks for all the cleanup.

PR #760, opened 2026-01-27
Remove further use of shape->vector conversions

ccummingsNV:
Minor optimizations by replacing the use of std::vector with manipulation of contents of slangpy Shape

### DIFF
--- a/nsyspython.bat
+++ b/nsyspython.bat
@@ -1,1 +1,1 @@
-nsys profile --trace=cuda,nvtx python %*
+nsys profile --trace=cuda,nvtx --gpuctxsw=true --cuda-trace-all-apis=true python %*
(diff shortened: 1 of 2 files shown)


src/sgl/utils/slangpy.h
```diff
@@ -263,6 +287,12 @@ class SGL_API Shape {
     /// Get size (i.e. number of dimensions) of shape.
     size_t size() const { return m_size; }
 
+    /// Iterator support for range-based for loops and algorithms
+    const int* begin() const { return data(); }
+    const int* end() const { return data() + m_size; }
+    int* begin() { return data(); }
+    int* end() { return data() + m_size; }
```
tdavidovicNV, line 294: Should this be using all those `reference` and `pointer` typedefs by now, since it supports begin/end?

ccummingsNV: Maybe, but tbh I need to be able to reason about these things in terms of performance, and all those typedefs make my brain hurt :)

Issue #758, opened 2026-01-26
Small Performance Win: Replace get_this / update_this probing


wlewNV:
These functions dominate arg-packing cost despite rare usage. Replace runtime probing with:
- cached type-based policies, or
- explicit protocol registration.

PR #763, opened 2026-01-28
Replace get_this / update_this probing

wlewNV:
Fixes https://github.com/shader-slang/slangpy/issues/758

[Work-in-Progress]
PLACEHOLDER DESCRIPITON

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -72,35 +72,8 @@ def set_print_generated_shaders(value: bool):
     _PRINT_GENERATED_SHADERS = value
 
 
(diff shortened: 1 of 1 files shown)


ccummingsNV:
Hi - that's a good start, but the optimization we have in mind is a bit more in depth. We do need to spin over these a bit more before diving in I think!

The idea here is to eliminate the need for get_this and update_this altogether, by making a base type from which a user can inherit.  we could have the get_this/update_this checks right at the end as a fallback, but we want to avoid ever paying this cost really - any microseconds is too many microseconds

PR #767, opened 2026-02-02
Add `TextureLoader::load_texture` overloads for multiple options

skallweitNV:
- Add `TextureLoader::load_textures` overloads that take a list of `TextureLoader::Options`, one for each texture to load.
- Rename `pydoc` target to `slangpy_pydoc` and make available even when not built as a master project.

### DIFF
--- a/src/slangpy_ext/CMakeLists.txt
+++ b/src/slangpy_ext/CMakeLists.txt
@@ -146,81 +146,79 @@ foreach(submodule IN ITEMS platform thread math ui tev crashpad renderdoc slangp
 
 endforeach()
 
-if(SGL_MASTER_PROJECT)
-    set(MKDOC_ARGS "")
+set(MKDOC_ARGS "")
 
-    # Add system include directories.
-    foreach(value ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
-        list(APPEND MKDOC_ARGS -I${value})
+# Add system include directories.
+foreach(value ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
+    list(APPEND MKDOC_ARGS -I${value})
+endforeach()
+
+# Add target include directories.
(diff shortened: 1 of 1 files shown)


src/sgl/utils/texture_loader.cpp
```diff
@@ -290,21 +290,21 @@ inline ref<Texture> create_texture(
...
     Device* device,
     Blitter* blitter,
     std::span<SourceImage> source_images,
     std::span<thread::TaskHandle> source_image_tasks,
-    const TextureLoader::Options& options
+    std::span<const TextureLoader::Options> options
 )
 {
     SGL_ASSERT(source_images.size() == source_image_tasks.size());
+    SGL_ASSERT(source_images.size() == options.size());
```
tdavidovicNV, line 302: Can we easily pass same option for everyone? Create a span with just one data point, cheating on the size?

tdavidovicNV (approved):
I find it weird that for a single texture, we don't even need Options, but for multiple textures, each *must* have its own Option (no opt out, no "all use the same"). I think it will make the call site either inefficient, or cluttered.

skallweitNV:
No, `load_textures` now just has overloads that can take an option for each texture to load. You still have the `load_textures` overload where options is optional and the same for every texture if provided. Or am I missing the point?

src/sgl/utils/texture_loader.cpp
skallweitNV: I thought about that, for simplicity I opted for just creating a vector of the same options instead .. overhead is negligible when loading multiple images anyway.

tdavidovicNV:
> No, `load_textures` now just has overloads that can take an option for each texture to load. You still have the `load_textures` overload where options is optional and the same for every texture if provided. Or am I missing the point?

No, I am missing the point. I thought this was instead of, not in addition to. Sorry, low coffee :-D

PR #772, opened 2026-02-03
Switched to Slang 2026.1.2

tdavidovicNV:


ccummingsNV (dismissed):

tdavidovicNV:
This has now been switched from 2026.2 to 2026.1.2, due to changes in what Slang version is the latest released.
(Both 2026.1.2 and 2026.2 have previously failed, so they rolled back and retried 2026.1.2, which is the correct version and what we want to compile against)

PR #774, opened 2026-02-04
Set CUDA context at SlangPy entry points for reduced overhead

szihs:
- Add set_cuda_context_current() wrapper in Device class
  - Call setCudaContextCurrent() in NativeCallData::exec() before kernel execution
  - Add Python binding for manual context control
  - Update tests to set CUDA context before test execution

  Works with slang-rhi CUDA context optimization to eliminate per-call
  push/pop overhead by setting context once at entry points.

### DIFF
--- a/slangpy/testing/helpers.py
+++ b/slangpy/testing/helpers.py
@@ -213,7 +213,9 @@ def get_device(
     )
 
     if use_cache and cache_key in DEVICE_CACHE:
-        return DEVICE_CACHE[cache_key]
+        device = DEVICE_CACHE[cache_key]
+        device.set_cuda_context_current()  # Ensure CUDA context is current for cached devices
+        return device
 
     device = Device(
         type=type,


slangpy/testing/helpers.py
```diff
@@ -239,6 +239,10 @@ def get_device(
                 "Metal device does not support parameter blocks (requires argument buffer tier 2)"
             )
 
+    # Set CUDA context current on this thread (no-op for non-CUDA devices).
```
ccummingsNV: Is this necessary? If so we probably need to look into it. If cuda interop is not using the primary context then it should be.

src/sgl/device/device.h
```diff
@@ -656,6 +656,10 @@ class SGL_API Device : public Object {
     rhi::IDevice* rhi_device() const { return m_rhi_device; }
     rhi::ICommandQueue* rhi_graphics_queue() const { return m_rhi_graphics_queue; }
 
+    /// Set the device's CUDA context as current on this thread.
+    /// For non-CUDA devices, this is a no-op.
+    void set_cuda_context_current() { m_rhi_device->setCudaContextCurrent(); }
```
ccummingsNV: We still need the api at this level and the rhi to be push/pop I think - otherwise how does the user maintain a context stack? i.e. if they change it, how do they change it back?

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -462,6 +462,11 @@ nb::object NativeCallData::exec(
     nb::kwargs kwargs
 )
 {
+    // Set CUDA context current on this thread (no-op for non-CUDA devices).
```
ccummingsNV: Again, the goal here is to not need this at all. It should be the user's job to change context, and it should be very rarely needed as the correct context should already be active.

ccummingsNV (changes requested):
I think there's work to do here. We need to be in a place where we provide this mechanism for pushing/popping the cuda context, but it should be virtually never needed. If it is, we need to fix that.

The only time I'd expect a user to need it is if, for example, they were explicitly switching context to change gpu.

src/sgl/device/device.h
szihs: Ack

slangpy/testing/helpers.py
szihs: Setting explicit context setting in Device Constructor.

src/sgl/device/device.h
szihs: Added support

src/slangpy_ext/utils/slangpy.cpp
szihs: Not entirely true in case of slangpy (native) flow. There is no active context since unlike  PyTorch which calls torch.cuda.init() internally does the job of setting the primary context

Issue #740, opened 2026-01-16
Support TensorView/DiffTensorview in slangpy


mkeshavaNV:
These are types supported in slangtorch natively. The `DiffTensorView` comes from slang from [here](https://github.com/shader-slang/slang/blob/5894ce5fe3f59f9e71b0a39e713715efe0ae27e0/source/slang/diff.meta.slang#L959). The goal of this task is to support these types in slangpy as well so as to support porting over slangtorch kernels to slangpy with ease.

mkeshavaNV:
> [@mkeshavaNV](https://github.com/mkeshavaNV) Looks like there's overlap between this and [#610](https://github.com/shader-slang/slangpy/issues/610) ?

Kind of.  https://github.com/shader-slang/slangpy/issues/610 is more about missing *equivalence*. We have the equivalence now with the new Tensor types. This issue specifically is about supporting the actual `DiffTensorView` itself so that we support slang-torch kernels in slangpy. I will start a discussion on discord on this soon.

PR #775, opened 2026-02-04
Add DiffTensorView<T> and TensorView<T> support in slangpy

mkeshavaNV:
## Overview
slangpy has pytorch tensor support via it's own types implemented via the `Tensor*` set of structs. However, slangpy currently does not support `TensorView` or `DiffTensorView` that slangtorch supports. The goal here is to support backward compatibility for code that was originally written in slangtorch, that supports `TensorView`.

This implements full TensorView support in SlangPy, allowing torch.Tensor arguments to bind to TensorView<T> parameters in Slang functions. 

Issue: #740 

## Implementation

- Add TensorViewType class to reflection layer for proper type detection
- Generate wrapper struct with identical memory layout for CallData (TensorView is a magic type that Slang can't layout directly)
- Handle native offset calculation and cursor reads
- Adds a new `TensorView` test

## Generated Code

```
typealias _t_a = TensorView<float>;
static const int _m_a = 0;
typealias _t_b = TensorView<float>;
static const int _m_b = 0;
typealias _t_output = TensorView<float>;
static const int _m_output = 0;

struct CallData
{
    uint3 _thread_count;
    _t_a a;
    _t_b b;
    _t_output output;
}

void _trampoline(Context __slangpy_context__, CallData __calldata__)
{
    TensorView<float> a;
    TensorView<float> b;
    TensorView<float> output;
    a = __calldata__.a;
    b = __calldata__.b;
    output = __calldata__.output;
    add_tensorview(a, b, output);
}
```

## Release Notes

* **Documentation**
  * Updated PyTorch integration documentation with TensorView compatibility information and backward-compatibility notes.

* **New Features**
  * Added support for TensorView and DiffTensorView types in the type resolution system.
  * Extended parameter marshalling with trampoline hook points for custom code generation.

* **Tests**
  * Added comprehensive test suites for TensorView and DiffTensorView functionality with CUDA and PyTorch integration.

* **Chores**
  * Improved cross-platform consistency for file path handling and import organization.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -74,7 +74,7 @@ set(SGL_LOCAL_SLANG OFF CACHE BOOL "Use a local build of slang instead of downlo
 set(SGL_LOCAL_SLANG_DIR "${CMAKE_SOURCE_DIR}/../slang" CACHE PATH "Path to a local slang build")
 set(SGL_LOCAL_SLANG_BUILD_DIR "build/Debug" CACHE STRING "Build directory of the local slang build")
 
-set(SGL_SLANG_VERSION "2026.1.2" CACHE STRING "Slang version to download when not using a local build")
+set(SGL_SLANG_VERSION "2026.2.1" CACHE STRING "Slang version to download when not using a local build")
 
 set(SLANG_URL_BASE "https://github.com/shader-slang/slang/releases/download/v${SGL_SLANG_VERSION}/slang-${SGL_SLANG_VERSION}")
 


src/slangpy_ext/utils/slangpytensor.h
```diff
@@ -131,11 +144,13 @@ class NativeTensorMarshall : public NativeMarshall {
...
         int array_stride;
-        ShaderOffset data;     // Offset for _data field
-        ShaderOffset shape;    // Offset for _shape field
-        ShaderOffset strides;  // Offset for _strides field
-        ShaderOffset offset;   // Offset for _offset field
-        bool is_valid = false; // Whether offsets have been initialized
+        ShaderOffset data;
+        ShaderOffset shape;
+        ShaderOffset strides;
+        ShaderOffset offset;          // Tensor only
```
mkeshavaNV: I did think of using a union here, but maybe this is better readability wise? I was also thinking of maybe forking off a separate struct for offsets for TensorView, but that seems like an overkill. Suggestions welcome for this here.

slangpy/core/calldata.py
```diff
@@ -278,8 +278,8 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
             sanitized = ""
             if _DUMP_GENERATED_SHADERS or _DUMP_SLANG_INTERMEDIATES:
                 os.makedirs(".temp", exist_ok=True)
-                santized_module = re.sub(r"[<>, ./:]", "_", build_info.module.name)
-                sanitized = re.sub(r"[:<>, ./:]", "_", build_info.name)
+                santized_module = re.sub(r"[<>, ./:\\]", "_", build_info.module.name)
+                sanitized = re.sub(r"[:<>, ./:\\]", "_", build_info.name)
```
mkeshavaNV, line 282: needed for weird windows' test errors:
` '.temp/C_\\actions-runner\\_work\\slangpy\\slangpy\\slangpy\\te_copy_tensorview-bec7b171.slang'`

slangpy/core/callsignature.py
```diff
@@ -379,9 +379,17 @@ def generate_code(
                     if x.create_param_block
                     else f"call_data.{x.variable_name}"
                 )
```
ccummingsNV, line 386: if we've got a 'TensorViewData', why not just add an '__slangpy_load' function on it? Then you don't need bespoke generated code just for tensor views

slangpy/core/callsignature.py
```diff
@@ -428,9 +436,15 @@ def generate_code(
...
-            cg.trampoline.append_statement(
-                f"{data_name}.__slangpy_store(__slangpy_context__.map(_m_{x.variable_name}), {x.variable_name})"
-            )
+            # SlangPy tensors require an explicit __slangpy_store() call to write results
+            # back through the marshalling interfaces. With TensorView the .store() method
+            # writes directly to GPU memory.
+            if isinstance(x.vector_type, TensorViewType):
```
ccummingsNV: As with the load, the __slangpy_store could just be empty

slangpy/slang/tensorviewdata.slang
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+implementing slangpy;
+
+// TensorViewData - a struct with identical memory layout to CUDA's TensorView.
+// Used as a wrapper in CallData because TensorView is a magic type with size=0
+// in Slang's reflection, which breaks struct layout calculations.
+public struct TensorViewData {
```
ccummingsNV: Is this really needed slang side? I'm guessing TensorView can be passed as a uniform / a member of call data, and can be extended to have __slangpy_load/store functions. I'd assumed TensorViewData would only be needed c++ side, as a way of populating the structure before binding.

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -170,8 +171,25 @@ def is_writable(self) -> bool:
 
     def resolve_types(self, context: BindContext, bound_type: SlangType):
         """Resolve types during binding phase."""
+        if isinstance(bound_type, TensorViewType):
+            return self._resolve_tensorview(context, bound_type)
         return spytc.resolve_types(self, context, bound_type)
 
+    def _resolve_tensorview(self, context: BindContext, bound_type: TensorViewType):
```
ccummingsNV: should this logic be in spytc.resolve_types?

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -158,13 +158,35 @@ namespace {
...
     TensorFieldOffsets offsets;
-    offsets.data = tensor_cursor["_data"].offset();
-    offsets.shape = tensor_cursor["_shape"].offset();
-    offsets.strides = tensor_cursor["_strides"].offset();
-    offsets.offset = tensor_cursor["_offset"].offset();
-    offsets.is_valid = true;
-    offsets.array_stride
-        = (int)tensor_cursor["_shape"].slang_type_layout()->getElementStride(SLANG_PARAMETER_CATEGORY_UNIFORM);
+
+    ShaderCursor dim_count_cursor = tensor_cursor.find_field("dimensionCount");
```
ccummingsNV: I had envisaged us simply storing the start of the structure for a TensorView, and then pushing the whole thing into the shader object in one go using set_data.

i.e. given the memory layotu is always known, and always identical, we can just populate a c++ side structure at push it into the shader object

src/slangpy_ext/utils/slangpytensor.h
```diff
@@ -131,11 +144,13 @@ class NativeTensorMarshall : public NativeMarshall {
     /// Public so NativeTorchTensorMarshall can reuse them
     struct TensorFieldOffsets {
         int array_stride;
-        ShaderOffset data;     // Offset for _data field
-        ShaderOffset shape;    // Offset for _shape field
-        ShaderOffset strides;  // Offset for _strides field
-        ShaderOffset offset;   // Offset for _offset field
-        bool is_valid = false; // Whether offsets have been initialized
+        ShaderOffset data;
```
ccummingsNV: As above, my instinct is that if you know its a tensor view, you just need to know where the start of the tensor field is and you can bind 1 structure.

slangpy/slang/tensorviewdata.slang
mkeshavaNV: This was the problem I was referring yesterday as well. `TensorView` has size 0, and is marked as magic type in slang. So reflection does not work. Here are some debug prints to show this better:

From `src/slangpy_ext/utils/slangpy.cpp`:
```
[DEBUG C++]   field 'input': type='TensorViewData', offset=24, size=56
[DEBUG C++]   field 'output': type='TensorViewData', offset=80, size=56
```
This would be 0 for `TensorView`.

slangpy/core/callsignature.py
mkeshavaNV: Done

slangpy/core/callsignature.py
mkeshavaNV: Done

slangpy/torchintegration/torchtensormarshall.py
mkeshavaNV: Done.

src/slangpy_ext/utils/slangpytensor.cpp
mkeshavaNV: Done. Yeah this works. I was just looking at the flow of Tensor implementation and doing it, but this is much cleaner. Thanks!

src/slangpy_ext/utils/slangpytensor.h
mkeshavaNV: I believe this comment was referring to the fact that we can just copy entire struct at once, correct? That is now done in `slangpytorchtensor.cpp`.

slangpy/builtin/tensorcommon.py
```diff
@@ -266,6 +286,8 @@ def resolve_dimensionality(
     Once a target type has been selected for vectorization, this function is called
     to determine the dimensionality of the call from the perspective of a bound variable.
     """
+    if isinstance(vector_target_type, TensorViewType):
+        return self.dims
```
ccummingsNV: I think this should probably be 0. i.e if I pass a pytorch tensor of shape [10,10,10] to a TensorView, it should end up with a slang side tensor of dimensionality 3. As a result, the dispatch itself is 1-to-1.

This dimensionality is how many dimensions the call needs to correctly vectorize. So for a 3D tensor:
- passing to a float, would need 3 dimensions
- passing to a vector would need 2 dimensions
- passing to a  matrix would need 1 dimension
- passing to a **2D** tensor would also need 1 dimension
- passing to a 3D tensor would 0 dimensions (i.e. the python argument maps 1-to-1 with the slang argument.

With tensor views, we're assuming the slang side tensor has the same number of dimensions as the python tensor, so it should still be 1-to-1.

slangpy/slang/tensorviewdata.slang
ccummingsNV: Sure, but why does it matter is my question? Does it break the actual shader object logic altogether? i.e. can we not just push a c++ side 'TensorViewData' to the shader object by calling set_data, and assume it is 56 bytes? Maybe it'd need the shader object logic to handle this as well?

This just feels like a hack that we should try to minimize, or ideally, fix properly. The correct fix is that Slang shouldn't be returning a size of 0 for a type that isn't of size 0. As you said 'reflection does not work' - this is a bug!

slangpy/tests/slangpy_tests/test_tensorview.py
```diff
@@ -0,0 +1,76 @@
...
+    module = Module.load_from_file(
+        device,
+        str(Path(__file__).parent / "test_tensorview.slang"),
+    )
+
+    a = torch.tensor([1.0, 2.0, 3.0, 4.0, 5.0], device="cuda", dtype=torch.float32)
+    b = torch.tensor([10.0, 20.0, 30.0, 40.0, 50.0], device="cuda", dtype=torch.float32)
+    output_tensor = torch.zeros(5, device="cuda", dtype=torch.float32)
+
+    module.add_tensorview(a, b, output_tensor)
```
ccummingsNV, line 72: You should turn on the high debug level for your module. I suspect that with your current code, this dispatch is generating a 1D dispatch of 5 threads, where it should be a 0D dispatch with 0 threads.

Once that's done, it would be good to create tests that take an extra 'coordinate', and pass in a spy.grid, to verify your logic is solid for multi-thread dispatches.

slangpy/tests/slangpy_tests/test_type_resolution.py
```diff
@@ -1121,6 +1170,21 @@ def __call__(self, module: spy.Module) -> spy.Texture:
...
+    ("func_tensorview_float", _TorchTensor("float", 2), "TensorView<float>", 2),
+    ("func_tensorview_int", _TorchTensor("int", 1), "TensorView<int>", 1),
+    ("func_tensorview_half", _TorchTensor("half", 2), "TensorView<half>", 2),
+
+    # Type mismatch should fail
+    ("func_tensorview_float", _TorchTensor("int", 1), None, None),
+    ("func_tensorview_int", _TorchTensor("float", 1), None, None),
+
+    # Generic TensorView<T> resolution
+    ("func_tensorview_generic", _TorchTensor("float", 1), "TensorView<float>", 1),
```
ccummingsNV, line 1218: Nice - thanks for adding this!

slangpy/tests/slangpy_tests/test_tensorview.py
```diff
@@ -0,0 +1,76 @@
...
+def test_tensorview_copy(device_type: DeviceType):
+    """Test copy_tensorview function from test_tensorview.slang."""
+    device = helpers.get_torch_device(device_type)
+
+    module = Module.load_from_file(
+        device,
+        str(Path(__file__).parent / "test_tensorview.slang"),
+    )
+
+    input_tensor = torch.tensor([1.0, 2.0, 3.0, 4.0, 5.0], device="cuda", dtype=torch.float32)
```
ccummingsNV, line 49: Good tests, but can we add some for normal slangpy tensors as well, given you support them?

slangpy/tests/slangpy_tests/test_type_resolution.py
ccummingsNV: Maybe add a couple that verify a normal slangpy tensor also resolves correctly?

slangpy/slang/tensorviewdata.slang
mkeshavaNV: 1. Regarding slang returning 0, no that does not seem like a bug. The `TensorView` is designed as a `__magic_type` which is designed to be as an opaque type in slang. I also looked at another type `DifferentialPair` and that is also similar. I'm guessing the reason is becuause the layout size maybe target dependent? (cuda/spv?)
I've started a thread to confirm this and get more clarity.

2. If we get rid of this struct, yes we need to make changes in the shader object logic to add weird checks.

Will continue this once I get more clarity.

slangpy/tests/slangpy_tests/test_tensorview.py
mkeshavaNV: Indeed. Set that dimension to 0. It was indeed dispatching 5 threads. Added an extra test. Let me know if that's in-line with what you were thinking.

slangpy/tests/slangpy_tests/test_type_resolution.py
mkeshavaNV: Done

slangpy/tests/slangpy_tests/test_tensorview.py
mkeshavaNV: Ah, yes, done.

slangpy/builtin/tensorcommon.py
mkeshavaNV: Ah, yes set to 0.

mkeshavaNV:
@ccummingsNV - PR https://github.com/shader-slang/slang/pull/9944 was approved. I will rebase this branch and ping once CI passes. I believe I will need to make a slang release for CI here to pick up the new build?

slangpy/core/callsignature.py
ccummingsNV: I guess the bespoke code gen for tensorview is back because we're using it directly internally?

Typically, we'd want to avoid this - the whole point of SlangPy is to make it extendable, and for the marshalls to define any 'per type' behaviour. With this bespoke code gen you're bypassing that entire process.

However, long term we do want to be able to do this in general - i.e. pass things directly where they don't need 'wrapping up'. This is part of the GPU optimization we were talking about with @szihs  To take a first step, why don't you put some logic on to BoundVariable (eg 'is_uniform' that says "this can be added directly as a uniform instead of needing a slang-side marshall".

Your BoundVariable then call 'self.python.is_uniform' which can be overriden for all types as 'False' other than TensorView. It's a tiny start, but sets us down the right path.

vcpkg.json
```diff
@@ -5,8 +5,7 @@
         "libjpeg-turbo",
         "libpng",
         "openexr",
-        "asmjit",
-        "crashpad"
+        "asmjit"
```
ccummingsNV: What's going on here?

src/slangpy_ext/utils/slangpytensor.cpp
```diff
@@ -194,6 +200,20 @@ NativeTensorMarshall::CachedOffsets NativeTensorMarshall::extract_offsets(Shader
         if (grad_out_field.is_valid()) {
             offsets.grad_out = extract_tensor_field_offsets(grad_out_field);
         }
+    } else {
+        // Check for DiffTensorViewData's primal/diff pattern (no underscore prefix)
+        ShaderCursor dtv_primal_field = field.find_field("primal");
```
ccummingsNV: This seems pretty fragile. Why not just get the type information from the shader cursor if you want to know if its a TensorView/DiffTensorView?

slangpy/torchintegration/__init__.py
```diff
@@ -1 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+try:
+    import torch
+    from slangpy.core.native import NativeTorchTensorDiffPair
+
+    class DiffPair(NativeTorchTensorDiffPair):
```
ccummingsNV: I think you've overcomplicated this. All you need is a function:

def diffPair(primal,grad):
    return NativeTorchTensorDiffPair(primal,grad)

Then you don't have to worry about inheritance or registering new types or anything

slangpy/bindings/boundvariable.py
```diff
@@ -501,6 +507,11 @@ def _calculate_differentiability(self, mode: CallMode):
         """
         Calculates access types based on differentiability, call mode and io type
         """
+        # DiffTensorView manages differentiation internally via load()/store().
+        # In backward mode, treat it as differentiable so its data pointers are accessible.
+        if mode == CallMode.bwds and isinstance(self.vector_type, DiffTensorViewType):
+            self.differentiable = True
```
ccummingsNV: This we can't do. A way to think of slangpy is that it is designed to be extendable. That is - we want users to be able to easily extend it with their own types in the long run. So if we find ourselves having to write type-specific code inside the core binding system (rather than in the marshalls), that should ring alarm bells. Users won't have this option, so neither should we :)

In theory if:
- your DiffTensorViewType in reflectiontypes.py returns True for differentiable
- the tensor marshall returns True for has_derivative (which should be the case if it has a python tensor with a gradient)

Then self.differentiable should end up true. If not, we've got a bug in that logic we need to solve, which is probably impacting other areas.

vcpkg.json
mkeshavaNV: removed. Slang ToT on Linux was broken and I had this change locally as a WAR. Included by mistake. Removed it.

slangpy/bindings/boundvariable.py
mkeshavaNV: So, self.Differentiable is set to False for DiffTensorView. Now, that didn't seem like a bug in slang because slang internally handles it through explicit custom derivatives. 

I added debug prints to check this and `self.differentiable` is always set to False. 
```
DEBUG input: no_diff=True, vector_type.differentiable=False, python.has_derivative=True => differentiable=False
DEBUG output: no_diff=True, vector_type.differentiable=False, python.has_derivative=True => differentiable=False
```

Now, with this, we have:
`differentiable=False, access = (read, none) == no derivative writes`

This was why the hack was needed. This was needed earlier when I was adding the slangpy_load/store for TensorView/DiffTensorview.

Removed this completely. It is not needed anymore with the direct assignment approach.

slangpy/torchintegration/__init__.py
mkeshavaNV: Ah, ok simplified this.

src/slangpy_ext/utils/slangpytensor.cpp
mkeshavaNV: Ok I suppose you meant to use type name from `field.slang_type_layout()->getName();`? I didn't find any other way to do this.

ccummingsNV: Yup - to my knowledge there's no other way, and this is done once so should be fine.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -415,6 +457,22 @@ void NativeTorchTensorMarshall::write_torch_tensor_fields(
...
+        // TensorView strides are in bytes, PyTorch strides are in elements
+        for (int i = 0; i < info.ndim && i < kSlangPyTensorViewMaxDim; i++) {
+            tvd.strides[i] = static_cast<uint32_t>(strides[i] * info.element_size);
+            tvd.sizes[i] = static_cast<uint32_t>(shape[i]);
+        }
+        tvd.dimensionCount = static_cast<uint32_t>(info.ndim);
+
+        shader_object->set_data(m_cached_offsets.field_offset, &tvd, sizeof(TensorViewData));
+        return;
+    }
```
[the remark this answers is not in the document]
mkeshavaNV: We don't have support for non-cuda backends for tensorview/difftensorview. So this seems like we can ignore.

ccummingsNV (approved):
LGTM - great job

Issue #387, opened 2025-07-29
Pytorch slice causes runtime error


KeKsBoTer:
Using a pytorch slice causes a runtime error.

Python / Pytorch code:

```python
import torch
import slangpy as spy

device = spy.create_device(type=spy.DeviceType.automatic, enable_cuda_interop=True)
module = spy.TorchModule.load_from_file(device, "vecmul.slang")

a = torch.rand(100).cuda()
b = torch.rand(4).cuda()

module.vecmul(a,b[:3])
```

cloning the tensor before passing it to `vecmul` is a workaround:
```python
module.vecmul(a,b[:3].clone())
```

vecmul.slang
```slang
float vecmul(float a, float[3] b)
{
    float result = 0.0f;
    for (int i = 0; i < 3; i++) {
        result += b[i] * a;
    }
    return result;
}
```

Error:
<details>

<summary>
RuntimeError: size does not match, self was 12 bytes but src was 16 bytes
</summary>

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[1], line 10
      7 a = torch.rand(100).cuda()
      8 b = torch.rand(4).cuda()
---> 10 module.vecmul(a,b[:3])

File ~/miniconda3/envs/my_env/lib/python3.13/site-packages/torch/nn/modules/module.py:1751, in Module._wrapped_call_impl(self, *args, **kwargs)
   1749     return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
   1750 else:
-> 1751     return self._call_impl(*args, **kwargs)

File ~/miniconda3/envs/my_env/lib/python3.13/site-packages/torch/nn/modules/module.py:1762, in Module._call_impl(self, *args, **kwargs)
   1757 # If we don't have any hooks, we want to skip the rest of the logic in
   1758 # this function, and just call forward.
   1759 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1760         or _global_backward_pre_hooks or _global_backward_hooks
   1761         or _global_forward_hooks or _global_forward_pre_hooks):
-> 1762     return forward_call(*args, **kwargs)
   1764 result = None
   1765 called_always_called_hooks = set()

File ~/miniconda3/envs/my_env/lib/python3.13/site-packages/slangpy/torchintegration/torchfunction.py:243, in TorchFunction.forward(self, *args, **kwargs)
[...]

PR #781, opened 2026-02-05
Add PyTorch autograd tests and fix interop/copy-back bugs

### LINKED ISSUE #733: Add tests: Complex use of torch tensors
Chris' comment from Discord:

> one set of things I'm nervous about is how more complex use of torch tensors works - i.e. passing none-contigous torch tensors, or torch tensors that are slices of others into slangpy. Have some test cases that exercise more 'real world' use of pytorch would be well worth it. Right now our tests verify the pytorch integration works but they certainly don't exercise it in anger
> A test case that operates on a sliced up torch tensor, or another that implements a loss function to stick on the end of some torch linear layers, or a bespoke slang adam optimizer applied to the parameters of a torch NN would all be valuable things that represent our user's predicted workloads.

jhelferty-nv:
Fixes #733, #387

### Bugfixes

- **CUDA interop memory leak**: free mapped `CUdeviceptr` with `cuMemFree` before destroying external memory, preventing OOM after many Vulkan dispatches.
- **Interop backward crash**: create zeroed interop buffer when primal tensor is `None` (backward output slot), instead of crashing.
- **Broadcast stride zeroing**: apply after contiguous stride computation for interop buffers so broadcast dimensions correctly use stride 0.
- **Copy-back correctness**: decide whether to copy interop buffers back based on the Slang uniform type name (`W`/`RW` prefix), cached at bind time. Prevents spurious `copy_()` on read-only inputs that break autograd's version tracking.
- **CUDA stream synchronization**: thread the CUDA stream from `NativeCallRuntimeOptions` through `CallContext` so `memset_device_async` and dispatch use the correct stream instead of stream 0.
- **Async memset**: replace synchronous `cuMemsetD8` with `cuMemsetD8Async` for zeroed interop buffers.

### Refactoring

- Rename `CachedOffsets` → `CachedBindingInfo`; now holds copy-back decision flags alongside shader offsets.
- `CallContext` constructor accepts an optional `NativeHandle cuda_stream`.

### New tests

- **test_pytorch_gradient_parity.py**: gradient correctness across activations, losses, slicing, transposes, strided ops, and copy-back semantics.
- **test_torch_autograd_workflows.py**: optimizer loops, chained kernels, gradient accumulation, mixed SlangPy + PyTorch graphs.

## Release Notes

* **New Features**
  * Added asynchronous CUDA memory set operation for improved performance.
  * Enhanced CallContext with CUDA stream support for better control over asynchronous execution.

* **Tests**
  * Added comprehensive PyTorch gradient parity tests validating consistency across activations, losses, and tensor operations.
  * Added autograd integration tests covering optimization workflows including polynomial fitting, curve fitting, and multi-layer networks.


src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -576,7 +576,15 @@ void NativeTorchTensorMarshall::write_shader_cursor_with_interop(
...
+    // Only copy back if:
+    // 1. The marshall allows writing (m_writable)
+    // 2. The binding's primal access includes write (not read-only)
+    // 3. There's data to copy (numel > 0) and an interop buffer
+    // This prevents incrementing PyTorch tensor _version for read-only inputs,
+    // which would break autograd's version tracking and cause RuntimeError during backward pass.
+    AccessType primal_access = binding->access().first;
+    bool primal_is_writable = (primal_access == AccessType::write || primal_access == AccessType::readwrite);
+    bool needs_primal_copyback = m_writable && primal_is_writable && primal_info.numel > 0 && primal_interop_buffer;
     bool needs_grad_copyback = has_grad && grad_info.numel > 0 && grad_interop_buffer;
```
ccummingsNV: I'm not convinced this logic is correct. It works for a scalar during broadcasting (i.e. pass tensor to float), but what if it's passing a tensor to a tensor. the access type would probably be 'read' (its just reading the tensor info) even though the tensor is written to.

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
```diff
@@ -0,0 +1,819 @@
...
+        nn.ReLU(),
+        nn.Tanh(),
+        nn.Sigmoid(),
+        nn.Linear(hidden_features, out_features),
+    ).cuda()
+
+    # Test: three consecutive SlangPy ops
+    test_model = nn.Sequential(
+        nn.Linear(in_features, hidden_features),
+        SlangpyActivation(slang_module, "slang_relu"),
```
ccummingsNV, line 821: is there a reason this needs wrapping - can it just be 'slang_module.slang_relu'?

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
```diff
@@ -0,0 +1,819 @@
...
+
+    def __init__(self, slang_module: Any, func_name: str):
+        super().__init__()
+        self.slang_module = slang_module
+        self.func_name = func_name
+        # Get the function from the module
+        self.func = getattr(slang_module, func_name)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.func(x)
```
ccummingsNV, line 161: it's a shape we have to do this. the slangpy function is callable, and we could add a '.forward' helper (that just called it). if we did that, can we remove the need to wrap things in modules

ccummingsNV (changes requested):
Good set of tests, though obviously they need to pass before we can merge. 

We should add tests that pass full tensors in as well as tests that just take scalar variables, as the interop logic is subtly different.

It would also be nice to see if we can get rid of the need to wrap slangpy functions in torch modules to work. Our goal is for a user to integrate seemlessly - creating custom wrapper classes isn't ideal.

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
jhelferty-nv: `nn.Sequential` requires `nn.Module` instances; there are explicit `isinstance` checks in the PyTorch source code.

Claude suggests we could try making `SlangFunction` conditionally inherit from `nn.Module` if PyTorch is available. In general, inheriting from `nn.Module` seems to be the suggested integration point in the [Modules documentation](https://docs.pytorch.org/docs/stable/notes/modules.html). It doesn't look like Duck typing is going to help us here.

Is it more palatable to have a `.as_module()` method on `SlangFunction`?

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
jhelferty-nv: (see other comment about needing to inherit from `nn.Module`)

src/slangpy_ext/utils/slangpytorchtensor.cpp
jhelferty-nv: Confirmed, I'll push a fix and tests for this.

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
ccummingsNV: Hmmm - that's a shame. I think for this PR let's not worry about it then. We can consider how to make it cleverer for another time. I've hit a few areas now where having something in 'torch mode' would be useful (eg all return values from functions being torch tensors by default).

src/slangpy_ext/utils/slangpytorchtensor.cpp
ccummingsNV: I'm not happy with this - it's getting too complex, and will slow down the dispatch. The simplest is just to check if it is writable, though as you point out that could result in redundant copying.

Probably the correct logic, if we want to do it, would be to adjust the logic that caches offsets to look at the slang uniform type that is being bound to, and identify it is  a writable tensor type (by name). This could work for gradient based tensors too. i.e. in that situation we're saying:
- if the user has explicitly written a function that takes a writable tensor, assume it is being written to
- otherwise, if the user has written a scalar function we're vectorizing across, and it has an 'out/inout' param, assume it is being written to

Exactly the same principle could be applied to the gradients, making that logic more robust/optimal as well.

As that code only happens once, it would not slow down dispatches. (though we'd want to rename some bits so it didn't claim to just be caching offsets when it in fact was caching more than that!).

Ultimately the actual decision making for this is in tensorcommon.py, where we generate the call data for the tensor. This is where we decide what actual tensor type the generated kernel uses. Another option would be to write some boolean value 'needs copying' inside gen_calldata, but that feels wrong to me - gen_calldata by implication is immutable, and many design decisions are based on the idea that a marshall is complete from the moment it's created.

We probably also need to apply this logic to the normal tensor right?

ccummingsNV (changes requested):
Hi. This is great work - sorry to be picky on the binding - it's just important we absolutely nail that particular logic, both in terms of correctness and performance.

I've detailed what I think is the only robust path for us now - ultimately looking at the type that is being bound to, which we conveniently do have access to when calculating offsets. I think what the actual slang side type that we're binding the tensor to (note *not* the same as the argument type of the user's function) is the concrete answer to when/whether copies are needed.

src/slangpy_ext/utils/slangpytorchtensor.cpp
jhelferty-nv: Okay, I've gone ahead and moved the logic to cache-time. We do need this fixed, because more complex cases where a Sequential jumps back and forth between SlangPy and PyTorch will generate an assert if we end up copying a read-only input tensor. e.g.,

```
RuntimeError: one of the variables needed for gradient computation has been modified 
by an inplace operation: [torch.cuda.FloatTensor [32, 16]], which is output 0 of 
TanhBackward0, is at version 2; expected version 0 instead.
```

The cache-time logic uses `TypeReflection::Kind` instead of the type name ("RWTensor" or "WTensor").. hopefully this aligns with what you were thinking?

As for normal tensors, IIUC, the normal SlangPy Tensor doesn't need copy-back because the shader writes directly to the tensor's GPU buffer (via device_address()). The interop buffer / copy-back mechanism is specifically for PyTorch tensors on non-CUDA backends, where SlangPy can't write directly to PyTorch's CUDA memory from a Vulkan/D3D12 shader.

Or as Claude puts it, the asymmetry is:
- **SlangPy buffer on Vulkan/D3D12**: The buffer is native to Vulkan/D3D12. The shader writes directly to it. cuda_memory() provides a CUDA view of this same memory for interop (if needed).
- **PyTorch tensor on Vulkan/D3D12**: The tensor is native to CUDA (PyTorch created it). Vulkan/D3D12 shaders can't write to CUDA-owned memory. So we create a SlangPy interop buffer (which Vulkan/D3D12 can access), run the shader, then copy the results back to PyTorch via CUDA.

slangpy/tests/slangpy_tests/test_pytorch_gradient_parity.py
jhelferty-nv: Created #790 for follow-up.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -235,11 +235,38 @@ Shape NativeTorchTensorMarshall::get_shape(nb::object data) const
...
+        m_cached_binding_info = NativeTensorMarshall::extract_binding_info(field);
+
+        // Compute primal copy-back flag based on binding type and access mode.
+        //
+        // For simple types (scalar/vector/matrix), the tensor is "broadcast" per-thread.
+        // Read-only inputs must NOT copy back - this would increment PyTorch's _version
+        // counter via copy_(), breaking autograd's version tracking.
+        // Writable outputs MUST copy back to return results.
+        //
+        // For tensor types (Tensor, RWTensor, etc.), copy back if marshall is writable.
```
ccummingsNV: Good that it's being cached, but I honestly think the solution here, as I said below, is to check the type of the tensor being bound to, not attempt to infer writability from the vector type. To clarify:
- vector type == type being passed to the user's function
- slang type == the uniform type stored in the generated kernel's call data, which this marshall represents (this is what the cursor is pointing at)

Python side, a lot of logic goes into working out whether the slang type should be Tensor, WTensor, RWTensor, etc etc. This is the 'ground truth' wtih regards to whether copying is needed. A read-only tensor uniform does not need copy-back. A write-only tensor uniform technically does not need initial copy.

The same principle applies to differentiable tensors, though is more complex:
- DiffTensor == read only primal, writable gradient (grad out)
- WDiffTensor == writable primal, readable gradient (grad in)
- RWDiffTensor == rw primal, rw gradient (grad out + grad in)

So by trying to work things out from the vector type here, you're effectively trying  to replicate logic that already happened python side. Even if you get it right for now, it'll mean 2 places this logic needs to be maintained.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -478,59 +512,73 @@ void NativeTorchTensorMarshall::write_shader_cursor_with_interop(
...
 
+        ref<Buffer> interop_buffer = context->device()->create_buffer({
+            .size = buffer_size,
+            .struct_size = static_cast<size_t>(info.element_size),
+            .usage = BufferUsage::unordered_access | BufferUsage::shader_resource | BufferUsage::shared,
+            .default_state = ResourceState::shader_resource,
+        });
+        void* cuda_ptr = interop_buffer->cuda_memory();
+        if (cuda_ptr && buffer_size > 0)
+            cuda::memset_device(static_cast<uint8_t*>(cuda_ptr), 0, buffer_size);
```
ccummingsNV: This is a none-async call, so would block the full gfx pipeline on execution. It should utilize a cuMemSetAsync, and provide the stream this operation is being run on.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -478,59 +512,73 @@ void NativeTorchTensorMarshall::write_shader_cursor_with_interop(
...
             .struct_size = static_cast<size_t>(info.element_size),
             .usage = BufferUsage::unordered_access | BufferUsage::shader_resource | BufferUsage::shared,
             .default_state = writable ? ResourceState::unordered_access : ResourceState::shader_resource,
         });
-
-        // Copy data from PyTorch tensor to interop buffer using TorchBridge
-        // This handles non-contiguous tensors via PyTorch's copy mechanism
-        // copy_to_buffer() now throws on error with detailed message
         if (info.numel > 0 && info.data_ptr != nullptr) {
             TorchBridge::instance().copy_to_buffer(tensor_value, interop_buffer->cuda_memory(), buffer_size);
```
ccummingsNV, line 585: I know you didn't write this, but just noted - for related reasons to the cuda::memset_device point, this could cause a bug. If the copy_to_buffer I wrote isn't running on the correct cuda stream, we could end up with the copy occuring after the actual dispatch.

ccummingsNV (changes requested):
Nice catch on the memory leak.

I think the binding is still fragile, and needs to piggy back off what the system already decided rather than attempt to re-calculate writability. The shader cursor already has enough information to say _exactly_ whether the tensor is writable.

We also need to make that memset async if it is to be used in a performant way.

src/slangpy_ext/utils/slangpytorchtensor.cpp
jhelferty-nv: Should be addressed.

src/slangpy_ext/utils/slangpytorchtensor.cpp
jhelferty-nv: I think this has been addressed?

src/slangpy_ext/utils/slangpytorchtensor.cpp
jhelferty-nv: Logic should be simplified; I'm now just using the Tensor type name chosen by the python code to determine things.

ccummingsNV (dismissed):
Looks great - nice work

kaizhangNV (dismissed):

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -455,29 +485,24 @@ void NativeTorchTensorMarshall::write_torch_tensor_fields(
...
-            tvd.strides[i] = static_cast<uint32_t>(strides[i] * info.element_size);
-            tvd.sizes[i] = static_cast<uint32_t>(shape[i]);
-        }
-        tvd.dimensionCount = static_cast<uint32_t>(std::min(info.ndim, kSlangPyTensorViewMaxDim));
-
-        shader_object->set_data(m_cached_offsets.field_offset, &tvd, sizeof(TensorViewData));
+        // TensorView path: build TensorViewData struct and write via set_data()
+        TensorViewData tvd = populate_tensorview_data(info, shape, strides);
+        shader_object->set_data(m_cached_binding_info.field_offset, &tvd, sizeof(TensorViewData));
         return;
```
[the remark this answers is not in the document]
jhelferty-nv: This is already in origin/main. I think coderabbit may be having trouble distinguishing past the merge commit, to understand which changes are coming from my PR vs upstream. I've created a separate issue for Mukund to follow up.

kaizhangNV (approved):

PR #789, opened 2026-02-08
reinterpret cursor

fknfilewalker:
Adds a way to reinterpret a BufferElementCursor object. This can be helpful for situations like writing a uint64 as two uint32 or when writing implementation data of dynamic interfaces within a struct.

```slang
public struct ConformanceWrapper<T> {
    uint _;
    uint __;
    uint id;
    uint ___;
    T data;
};

interface IFoo;
struct Bar : IFoo {
    float c;
}

struct MyData {
    float a;
    IFoo b;
}
```
```python
# shortened pseudocode
cursor = spy.BufferCursor("StructuredBuffer<MyData>")
cursor[0].a = 2.0
cursor[0].b.reinterpret("ConformanceWrapper<Bar>").id = 0
cursor[0].b.reinterpret("ConformanceWrapper<Bar>").data.c = 10
```

The manual id part is not nice, but currently that is the best we can do.

## Summary by CodeRabbit

* **New Features**
  * Cursor reinterpretation: view and operate on buffer and shader cursors using alternative type layouts from both C++ and Python.

* **Tests**
  * New tests validating reinterpretation round-trips and memory consistency when fields are accessed under different layouts.

* **Documentation**
  * Added docstrings describing the reinterpret operations for cursor APIs.


src/slangpy_ext/device/buffer_cursor.cpp
```diff
@@ -23,6 +23,10 @@ SGL_PY_EXPORT(device_buffer_cursor)
     nb::class_<BufferElementCursor> buffer_element_cursor(m, "BufferElementCursor", D(BufferElementCursor));
 
     buffer_element_cursor //
+        .def("reinterpret", [](BufferElementCursor& self, ref<TypeLayoutReflection> new_layout)
```
ccummingsNV: Any reason this is done with a lamda function rather than just binding the address of the new function>

ccummingsNV (changes requested):
Needs tests and the function needs doxygen comments.

src/slangpy_ext/device/buffer_cursor.cpp
fknfilewalker: nanobind has complains about the `ref<const TypeLayoutReflection>` type and says it is ambiguous, removing the const works but at the same time c++ side needs the const. So the direct function signatur can not be used.

tests/sgl/device/test_cursors.cpp
```diff
@@ -90,6 +96,20 @@ struct TestStruct
         (*tocursor_cursor)[0] = cpu_struct;
 
         CHECK(memcmp(from_direct.data(), from_tocursor.data(), from_direct.size()) == 0);
+
+        // reinterpret uint32_t data as uint16_t[2]
```
ccummingsNV, line 127: This needs to be a separate test (eg buffer_cursor_reinterpret). Our tests (where possible) should generally each test one thing - that way we know what's broken when one fails.

Ideally though we need a python test for this (see slangpy\tests\device\test_buffer_cursor.py). Whilst native tests are good, we make a point of ensuring every element of the python api is tested, as that's what we're expecting users to interact with most.

src/slangpy_ext/device/buffer_cursor.cpp
```diff
@@ -23,6 +23,10 @@ SGL_PY_EXPORT(device_buffer_cursor)
     nb::class_<BufferElementCursor> buffer_element_cursor(m, "BufferElementCursor", D(BufferElementCursor));
 
     buffer_element_cursor //
+        .def("reinterpret", [](const BufferElementCursor& self, ref<TypeLayoutReflection> new_layout)
```
ccummingsNV: Does this actually belong on bind_readable_cursor or bind_writable_cursor (src\slangpy_ext\device\cursor_utils.h) which are available to both shader cursor and buffer cursors, or is it fundamentally buffer specific?

ccummingsNV (changes requested):
Thanks for adding the test. Few more comments - I'm wondering if this is a general cursor thing, rather than just buffer cursor?

If not, let's get a python side test in and then good to go.

src/slangpy_ext/device/buffer_cursor.cpp
fknfilewalker: I guess one could also find this useful for shader cursors.
I don't think this directly relates to bind_readable_cursor/bind_writable_cursor as reinterpret just mirrors the properties of the original cursor and can be useful for either read, write or rw cursors. But maybe I am just getting something wrong.

ccummingsNV (changes requested):
fyi, this looks good to me once tests pass

jkwak-work:
@fknfilewalker , can you make some changes to make the tests pass?

fknfilewalker:
@ccummingsNV the issue is how D3D packs data. It does not pack a uint16_t array tightly. 
So when we reinterpret a uint32 into two uint16, we actually access [0] at 0 bytes and [1] at 4 bytes offset in the D3D case.
I fixed it by accounting for this now, but I am not sure if this is the best solution. Maybe we should expose an option to allow choosing tight reinterpretation or not.

ccummingsNV (dismissed):
This looks fine to me. I think that if the user choose to perform reinterprets and wants their code to be cross platform, it's also their responsibility to verify this. If we start worrying about layouts this PR will grow a lot - uint16 arrays are the least of your worries :)

src/sgl/device/buffer_cursor.cpp
```diff
@@ -101,6 +114,11 @@ BufferElementCursor BufferElementCursor::find_element(uint32_t index) const
     return {};
 }
 
+size_t BufferElementCursor::element_stride() const
+{
+    return m_is_reinterpreted ? m_type_layout->element_type_layout()->size() : m_type_layout->element_stride();
```
ccummingsNV: This seems very dodgy to me - we shouldn't have to be storing whether a cursor has been reinterpreted to know how to treat it.

ccummingsNV (changes requested):
The new addition of an m_is_reinterpreted for me is a step backwards, and I suspect is brittle. What now happens if you reinterpret a reinterpreted thing? I have no idea :) 

Reinterpeting a cursor should either leave you with 1 of 2 things:
- an error, because the conversion is invalid for some reason
- a valid cursor with no need for extra data

fknfilewalker:
You are very right. Now I fixed the test by wrapping the reinterpret type in the desired buffer layout. 

```python
reinterpret_type_layout = module.layout.get_type_layout(
    module.layout.find_type_by_name("StructuredBuffer<uint16_t[2]>")
).element_type_layout
```

src/sgl/device/buffer_cursor.cpp
```diff
@@ -21,6 +21,15 @@ BufferElementCursor::BufferElementCursor(ref<const TypeLayoutReflection> layout,
 {
 }
 
+BufferElementCursor BufferElementCursor::reinterpret(ref<const TypeLayoutReflection> new_layout) const
+{
+    BufferElementCursor element_cursor;
+    element_cursor.m_buffer = m_buffer;
+    element_cursor.m_type_layout = new_layout;
```
ccummingsNV, line 29: Should we check if new_layout is null before making the cursor? I don't think there's anything to stop the user passing in 'None' accidentally, and then they'd get weird late-in-the-day crashes.

We could arguably verify the size hasn't changed too, though I'm 50/50 on that one.

src/slangpy_ext/device/buffer_cursor.cpp
```diff
@@ -23,6 +23,15 @@ SGL_PY_EXPORT(device_buffer_cursor)
     nb::class_<BufferElementCursor> buffer_element_cursor(m, "BufferElementCursor", D(BufferElementCursor));
 
     buffer_element_cursor //
+        .def(
+            "reinterpret",
+            [](const BufferElementCursor& self, ref<TypeLayoutReflection> new_layout)
```
ccummingsNV, line 28: I may have asked this before - does this need to be a lambda? Or can it just be a simple binding.

ccummingsNV (changes requested):
Sorry for repeated change suggestions! :)

A few comments - I'd strongly suggest the null check in the 2 cursors. Other 2 are your call.

src/slangpy_ext/device/buffer_cursor.cpp
fknfilewalker: unfortunately yes, otherwise it does not compile. There is a `ref<const TypeLayoutReflection>` vs `ref<TypeLayoutReflection>` problem otherwise.

src/sgl/device/buffer_cursor.cpp
fknfilewalker: I would say one option would be that it must be smaller or equal to the original size. Still I would rather not restrict this too much.

ccummingsNV (approved):
LGTM - thanks for the extra work neatening it up!

fknfilewalker:
Happy to contribute. 
I think the auto merge got stuck.

jkwak-work:
It seems like there is a problem on Github system.
I cannot "update branch" at the moment; I will retry it later.
image

jkwak-work:
"Update branch" works now.

fknfilewalker:
@jkwak-work can we merge this now?

jkwak-work:
> @jkwak-work can we merge this now?

I am merging the ToT again.

Note that this was supposed to be merged already but we were having intermittent CI failure problems, which is supposed to be resolve now.

PR #814, opened 2026-02-25
Optimize value types

ccummingsNV:
## Release Notes

* **New Features**
  * Enhanced tensor creation capabilities through improved bridge APIs with support for multiple scalar types.
  * Added caching mechanism for improved performance in value marshalling operations.

* **Bug Fixes**
  * Optimized benchmark tests with tensor reuse and improved gradient handling.
  * Corrected dtype mapping and scalar type code handling.

* **Chores**
  * Updated package version to 0.6.0.
  * Refactored internal hook invocation to use centralized bridge abstraction.


slangpy/benchmarks/test_benchmark_autograd.py
```diff
@@ -48,8 +49,8 @@
 # =============================================================================
 
 RUN_PURE_TORCH_BENCHMARK = False
-RUN_SLANGTORCH_BENCHMARK = True
-RUN_SLANGPY_MANUAL_HOOK_BENCHMARK = False
+RUN_SLANGTORCH_BENCHMARK = False
+RUN_SLANGPY_MANUAL_HOOK_BENCHMARK = True
```
tdavidovicNV, line 53: Is this swap intentional?

tdavidovicNV (approved):
Makes sense, the only thing that raised a red flag is the swap of which benchmarks are True and False.

PR #815, opened 2026-02-25
Pack/unpack optimization

ccummingsNV:
## Release Notes

* **New Features**
  * Added test coverage for argument packing and caching behavior.

* **Bug Fixes**
  * Improved parameter handling in debug mode type checking.

* **Performance**
  * Optimized function call performance by making argument unpacking conditional—skipped when not required for specific object types.

* **Tests**
  * Enhanced debug-only type validation for cursor operations.


src/sgl/device/cursor_access_wrappers.h
```diff
@@ -7,10 +7,11 @@
...
 #include "sgl/device/cursor_utils.h"
 
-// TODO: Decide if we want to disable / optimize type checks
-// currently can represent 50% of the cost of writes in
-// certain situations.
+// These checks are expensive and involve re-examining reflection data, so only
+// enable in debug builds.
+#ifdef _DEBUG
 #define SGL_ENABLE_CURSOR_TYPE_CHECKS
+#endif
```
tdavidovicNV, line 14: Every project I worked on ended up having "light" and "heavy" checks in the end.. maybe we should just formalize this?

PR #816, opened 2026-02-25
Native torch auto-grad

ccummingsNV:
## Summary by CodeRabbit

* **New Features**
  * Added zeros-like tensor creation API for creating tensors matching source shape, dtype, and device.
  * Enhanced PyTorch autograd integration with native backend delegation and backward pass support.

* **Tests**
  * Comprehensive test coverage for zeros-like tensor creation across various dtypes and shapes.

* **Chores**
  * Version bump to 0.7.0.

### DIFF
--- a/slangpy/tests/utils/test_torch_bridge.py
+++ b/slangpy/tests/utils/test_torch_bridge.py
@@ -539,3 +539,114 @@ class TestTorchBridgeCreateEmptyTensor:
         assert info["element_size"] == 4
         assert info["is_cuda"] is True
         assert info["is_contiguous"] is True
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+class TestTorchBridgeCreateZerosLikeTensor:
+    """Tests for create_torch_zeros_like_tensor via the TorchBridge.
+
+    Verifies that tensors created through the bridge (both native and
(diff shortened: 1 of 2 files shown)


slangpy/core/calldata.py
```diff
@@ -478,93 +480,52 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
...
                 raise
 
-    def find_torch_tensors(
-        self, args: list[Any], kwargs: dict[str, Any]
-    ) -> list[NativeTorchTensorDiffPair]:
+    def _build_autograd_access_list(self, args: list[Any], kwargs: dict[str, Any]) -> None:
         """
-        Find all torch tensors in args/kwargs, wrap them in NativeTorchTensorDiffPair,
-        and replace the tensors in args/kwargs with the pairs.
+        Walk args/kwargs in the same recursive order as find_torch_tensors,
```
tdavidovicNV, line 485: What's the danger here of the two ways of walking it getting out of sync?

ccummingsNV: Well, one function builds a list, and the other iterates over that list. The optimization works on the assumption that both functions visit each tensor in the same order. I can't really think of a safer way of writing it that isn't also a lot slower though

PR #817, opened 2026-02-25
Cursor: read ptrs as uint 

fknfilewalker:
When constructing objects via `module.find_struct("")()` it currently results in an error if the objects contains a ptrs to some type. The reason for this is that read is not implemented for ptrs. The best way would be to just handle pointers similar to how they are handled for writes, by just treating them as uint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of pointer-typed data so pointer values are read/written as raw addresses and offsets from views are applied correctly when reconstructing device pointers.
* **Tests**
  * Added tests that validate construction and assignment of pointer fields on devices that support pointers; pointer-incapable devices are skipped.


ccummingsNV (commented):
Looks good. Before adding, does the opposite work? i.e. if we can read pointers, can we also write them? If not, we should add that too - important this is 2 way.

Otherwise looks good - just need to verify that first.

ccummingsNV (changes requested):
Maybe add some tests for this to slangpy\tests\device\test_buffer_cursor.py

Once that's in, we can go ahead and merge this

jkwak-work:
@fknfilewalker , can you review the last comment from Chirs?

fknfilewalker:
@jkwak-work @ccummingsNV
I added a test which on one hand constructs a struct with a ptr field but also reads/writes the field through a cursor. Additional I made sure that StridedBufferViews (Tensors) are properly written when the reflection type is a ptr, instead of turning into uniform.

ccummingsNV (dismissed):

ccummingsNV (dismissed):

fknfilewalker:
@coderabbitai autofix

fknfilewalker:
@coderabbitai fix clang formating

fknfilewalker:
@coderabbitai autofix

fknfilewalker:
@coderabbitai apply both fixes

ccummingsNV:
Looks like you need to solve formatting.

fyi, if you run 'pre-commit install', it will install git hooks that validate/fix formatting on every commit. Locally if you need to (as with now), run 'pre-commit run --all-files' and it will fix all formatting.

definitely recommend the hook though!

if you get formatting fixed I'll approve + merge for you next time I'm online.

ccummingsNV (approved):

PR #818, opened 2026-02-26
Support _threadcount and float<N> for TensorView/DiffTensorview

### LINKED ISSUE #819: take a thread count, in the event the kernel has a dimensionality of 0 (no arguments are vectorized)

mkeshavaNV:
This CL adds support for 2 features.

Fixes #819

### 1. Optional _threadcount argument when launching kernels.

Slangpy's internal thrad count calculation is based on argument shapes. For eg: if we have a normal slangpy function with vectorizable arguments:

`void add(float a, float b) {}`

then slangpy launches 1000 threads if the user passed in two 1000 sized tensors. Now, when all arguments have call_dim = 0, i.e no arguments are vectorized, slangpy cannot determine how many threads to launch. In this case, a _threadcount helps slangpy to launch proper number of threads.
`[CUDAKernel]` functions typically take TensorView<T> / DiffTensorView<T> and this _threadcount helps in that case.

### 2. Support float<N> with TensorView and DiffTensorView

```
void _trampoline(Context __slangpy_context__, CallData __calldata__)
  {
      TensorView<vector<float,2>> input;
      TensorView<vector<float,2>> output;
      input = __calldata__.input;
      output = __calldata__.output;
      copy_tensorview_float2(input, output);
  }
```
Also notice from the bindings:
```
input  | RWTensor<float, 2> | TensorView<vector<float,2>> | [-1, -1] | Call Dim: 0
output | RWTensor<float, 2> | TensorView<vector<float,2>> | [-1, -1] | Call Dim: 0

```

## Summary by CodeRabbit

* **New Features**
  * TensorView resolution now accepts vector element types (e.g., float2/float4) for compatible scalar→vector bindings.
  * Expose a writable thread-count override on runtime options and a flag to indicate when it’s used.

* **Bug Fixes**
  * Validation added to reject thread-count usage for dispatch shapes where it’s invalid (clearer error).

* **Tests**
  * Added CUDA/differentiable kernel tests plus TensorView/vector-resolution tests covering forward/backward and dispatch paths.

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -114,6 +114,21 @@ class CallData(NativeCallData):
         self.build(build_info, *args, **kwargs)
 
     def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
+        # The follow section explain how _thread_count is handled.
+        # Phase 1 (C++, every call): read_thread_count_kwarg() reads the value from kwargs
+        #   into opts->m_thread_count, but leaves _thread_count in kwargs so Phase 2 can
+        #   see it. The signature naturally includes "_thread_count:int" when present,
+        #   creating a separate cache entry for calls with vs. without it.
+        #
+        # Phase 2 (Python, cache miss only): pop _thread_count before the
+        #   shader binding machinery runs, since it is not a Slang parameter and would
+        #   cause a binding error if left in kwargs.
+        #
+        # Phase 3 (C++, every call): exec() uses opts->has_thread_count() / thread_count()
+        #   to set total_threads. opts was already populated in Phase 1, so exec() never
+        #   needs to look at kwargs for this value.
+        has_thread_count = "_thread_count" in kwargs
+        kwargs.pop("_thread_count", None)
 
         try:
 
@@ -241,6 +256,18 @@ class CallData(NativeCallData):
             self.call_dimensionality = calculate_call_dimensionality(bindings)
             context.call_dimensionality = self.call_dimensionality
             self.log_debug(f"  Call dimensionality: {self.call_dimensionality}")
+
+            # _thread_count is only valid when call_dimensionality is 0 (all args are whole
+            # buffers/values and the kernel manages its own thread indexing). For auto-vectorized
+            # kernels (call_dimensionality > 0) the thread count is inferred from arg shapes.
+            if has_thread_count and self.call_dimensionality > 0:
+                raise ValueError(
+                    f"_thread_count is only valid for kernels with call dimensionality 0 "
+                    f"(i.e., all parameters are passed as whole buffers/values and the kernel "
+                    f"manages its own thread indexing). This kernel has call dimensionality "
+                    f"{self.call_dimensionality}, meaning the thread count is automatically "
+                    f"inferred from the shapes of the vectorized arguments."
+                )
 
             # If necessary, create return value node once call dimensionality is known.
             create_return_value_binding(context, bindings, return_type)
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -114,21 +114,7 @@ class CallData(NativeCallData):
         self.build(build_info, *args, **kwargs)
 
     def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
-        # The follow section explain how _thread_count is handled.
-        # Phase 1 (C++, every call): read_thread_count_kwarg() reads the value from kwargs
-        #   into opts->m_thread_count, but leaves _thread_count in kwargs so Phase 2 can
-        #   see it. The signature naturally includes "_thread_count:int" when present,
-        #   creating a separate cache entry for calls with vs. without it.
-        #
-        # Phase 2 (Python, cache miss only): pop _thread_count before the
-        #   shader binding machinery runs, since it is not a Slang parameter and would
-        #   cause a binding error if left in kwargs.
-        #
-        # Phase 3 (C++, every call): exec() uses opts->has_thread_count() / thread_count()
-        #   to set total_threads. opts was already populated in Phase 1, so exec() never
-        #   needs to look at kwargs for this value.
-        has_thread_count = "_thread_count" in kwargs
-        kwargs.pop("_thread_count", None)
+        self.has_thread_count = "_thread_count" in kwargs
 
         try:
 
@@ -170,6 +156,7 @@ class CallData(NativeCallData):
             # Unpack args (handles IThis wrappers)
             unpacked_args, args_had_unpack = unpack_args(*args)
             unpacked_kwargs, kwargs_had_unpack = unpack_kwargs(**kwargs)
+            unpacked_kwargs.pop("_thread_count", None)  # not a Slang parameter
             self.needs_unpack = args_had_unpack or kwargs_had_unpack
 
             # If we have torch tensors, enable torch integration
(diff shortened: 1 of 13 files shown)


src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -767,6 +767,11 @@ nb::object NativeCallData::exec(
         current_stride *= call_grid_data[i];
     }
 
+    // Override total_threads if _threadcount was explicitly provided.
+    if (opts->has_threadcount()) {
```
ccummingsNV: We should check here that the call dimensionality is 0. It'd be an error for the user to try to override the thread count in an auto vectorized case. 

For example, if the user passes a Tensor<float,2> of shape [32,32] to a function, `void foo(float x)`, the only valid call shape would be [32,32], meaning the call's dimensionality is 2, and the only valid thread count is 32*32.

Some sort of nice clear error stating that it is only valid to override thread count for 0 dimensional kernels (and what that actually means) would be worth having.

src/slangpy_ext/utils/slangpy.h
```diff
@@ -642,6 +642,15 @@ class NativeCallRuntimeOptions : Object {
     /// Set ray tracing pipeline flag.
     void set_is_ray_tracing(bool is_ray_tracing) { m_is_ray_tracing = is_ray_tracing; }
 
+    /// Get the thread count override.
+    int threadcount() const { return m_threadcount; }
```
ccummingsNV: as per current naming conventions, this should be 'thread_count', not 'threadcount' in all uses

src/slangpy_ext/utils/slangpyfunction.h
```diff
@@ -91,6 +91,20 @@ class NativeFunctionNode : NativeObject {
         }
     }
 
+    /// Extract runtime options from special underscore kwargs (e.g. _threadcount).
+    static void gather_kwargs_runtime_options(ref<NativeCallRuntimeOptions> options, nb::kwargs& kwargs)
```
ccummingsNV: This is a tricky one - it'll cause some perf issues. 

What actually happens with respect to this variable during kernel build? I assume it must be getting stripped out somewhere though I can't see where! 

I'd suggest we need to:
- make the fact that _thread_count is specified part of the cache signature (if it isn't already), 
- store on the call data at build time whether a thread count is expected. 
- then this if statement can just be 'if thread count expected' instead of a string lookup

That way this can boil down to just a bool check when the variable isn't present, with negligiable impact on performance. And when it is present, it's a single string lookup, instead of contains+lookup.

ccummingsNV (changes requested):
Great start. Can we change the naming to 'thread_count' to match existing work + add error checks.

My main concern here is:
- I'm not quite clear on how this variable is getting plumbed through / stripped out during kernel generation. Would be good to know
- I think we need to piggy back off the caching mechanism to ensure users only pay for this when they need it

src/slangpy_ext/utils/slangpy.cpp
ccummingsNV: This call shape calculation and everything below related to shapes and threads really does contribute to overall cost of the kernel. Can we skip a tonne of this if thread count has been specified? If so it might get us back an extra microsecond or 2, and compensate for the extra string variable checking.

src/slangpy_ext/utils/slangpy.cpp
mkeshavaNV: I moved this validation to python land. Hope it works there? It's an easy check to add there?

src/slangpy_ext/utils/slangpy.h
mkeshavaNV: Done

src/slangpy_ext/utils/slangpyfunction.h
mkeshavaNV: > I assume it must be getting stripped out somewhere though I can't see where!
yes, this variable is getting stripped out via `nb::del(kwargs["_thread_count"]);`.  I've tweaked the logic a little bit now to add this as part of the cache signature, but gets deleted when we have a cache hit.

> make the fact that _thread_count is specified part of the cache signature (if it isn't already),

Yes, this is handled now. I did some quick test to ensure cache hit with this signature. Here is some debug output of signature and whether cache hit happens. So when variable is used,
```
 [slangpy] MISS sig=fill_thread_ids\x0a{'strict_broadcasting': False}\x0aargs\x0akwargs\x0acount:int\x0aoutput:torch\x0a[D1,S3]_thread_count:int\x0a
     [slangpy] HIT  sig=fill_thread_ids\x0a{'strict_broadcasting': False}\x0aargs\x0akwargs\x0acount:int\x0aoutput:torch\x0a[D1,S3]_thread_count:int\x0a
```

> That way this can boil down to just a bool check when the variable isn't present, with negligiable impact on performance. And when it is present, it's a single string lookup, instead of contains+lookup.

So we are now down to 1 string look up per call at least, and multiple if the variable is present. The code is relatively simple this way. I am not sure if this 1 string look could be replaced as well. Does this suffice?

src/slangpy_ext/utils/slangpy.cpp
mkeshavaNV: Done, refactored a bit and added a separate method to handle this.

slangpy/core/calldata.py
```diff
@@ -170,6 +156,7 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
             # Unpack args (handles IThis wrappers)
             unpacked_args, args_had_unpack = unpack_args(*args)
             unpacked_kwargs, kwargs_had_unpack = unpack_kwargs(**kwargs)
+            unpacked_kwargs.pop("_thread_count", None)  # not a Slang parameter
```
mkeshavaNV, line 159: @ccummingsNV - Now, with the new code, the C++ deletion happens after generate_call_data returns. So this is needed.

src/slangpy_ext/utils/slangpyfunction.cpp
```diff
@@ -23,12 +23,29 @@ struct GcHelper<slangpy::NativeFunctionNode> {
...
+// _thread_count is included in the signature (kwargs are signature-scanned before this is called),
+// so calls with vs. without it are separate cache entries.
+// Only called when call_data->has_thread_count() is true (cheap bool check), so the
+// kwargs.contains() string lookup is avoided entirely on the common no-thread-count path.
+static bool read_thread_count_kwarg(ref<NativeCallRuntimeOptions> options, nb::kwargs& kwargs)
+{
+    if (!kwargs.contains("_thread_count"))
+        return false;
+    int thread_count = nb::cast<int>(kwargs["_thread_count"]);
+    if (thread_count <= 0)
```
ccummingsNV: Use the 'SGL_CHECK' macro

src/slangpy_ext/utils/slangpyfunction.cpp
```diff
@@ -23,12 +23,29 @@ struct GcHelper<slangpy::NativeFunctionNode> {
...
 namespace sgl::slangpy {
 
+// Read _thread_count from kwargs into options and remove it from kwargs.
+// _thread_count is included in the signature (kwargs are signature-scanned before this is called),
+// so calls with vs. without it are separate cache entries.
+// Only called when call_data->has_thread_count() is true (cheap bool check), so the
+// kwargs.contains() string lookup is avoided entirely on the common no-thread-count path.
+static bool read_thread_count_kwarg(ref<NativeCallRuntimeOptions> options, nb::kwargs& kwargs)
+{
+    if (!kwargs.contains("_thread_count"))
```
ccummingsNV: This should be an assert or a throw - by definition its a bug in the caching system if we end up trying to read a thread count arg when one doesn't exist

ccummingsNV (changes requested):
Couple of minor comments, and this feature needs to go in the documentation (I suggest getting it right in at the start in the 'firstfunctions.rst' section, after 'Automatic vectorization'.

Otherwise, looks great.

src/slangpy_ext/utils/slangpyfunction.cpp
mkeshavaNV: done, updated docs as well.

src/slangpy_ext/utils/slangpyfunction.cpp
mkeshavaNV: done.

ccummingsNV (approved):
Looks good. This does highlight though that we need to get in the ability for the user to somehow access dispatch thread id etc on none-cuda platforms. This is a **really** useful feature in principle - we regularly see users wanting to just call a function with N threads and pass it, eg, a texture. However they'd need some way of identifying the thread index without using the cuda block primitives. We should get this working asap (and update the docs once done).

Nice work though - great step forwards!

PR #780, opened 2026-02-05
Support cross-repo CI testing from Slang PRs

jkiviluoto-nv:
Modify ci-latest-slang workflow to accept repository_dispatch events from Slang repository PRs. This allows testing SlangPy against Slang changes before they are merged.

Changes:
- Add repository_dispatch trigger for slang-pr-test events
- Use conditional Slang ref (PR ref from payload or branch input)
- Add status reporting back to Slang PRs via GitHub Status API
- Update run name to show Slang PR number when triggered by dispatch

Relates to shader-slang/slang#9220

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI workflow to support programmatic test triggering and automated status reporting for Slang pull requests.
  * Updated workflow permissions and checkout logic to enable improved test tracking and result notifications.

PR #846, opened 2026-03-03
Run slang repo integration tests in ci-latest-slang

### LINKED ISSUE #845: Run slang repo integration tests in ci-latest-slang workflow
### Summary

The `ci-latest-slang.yml` workflow builds slang from source and runs slangpy tests against it. However, it does not run slangpy-based integration tests that live in the **slang** repo under `tests/integration/slangpy/`.

These tests verify neural module functionality via slangpy and are maintained in the slang repo because they test slang compiler features. They need a working slangpy + slang build to run, which `ci-latest-slang` already provides.

### Proposed Change

Add a step to `ci-latest-slang.yml` to discover and run any slangpy integration tests from the slang checkout:

```yaml
- name: Run slang integration tests
  if: contains(matrix.flags, 'unit-test')
  run: |
    if [ -d "slang/tests/integration/slangpy" ]; then
      python -m pytest slang/tests/integration/slangpy/ -ra -v
    fi
```

This is guarded by an `if [ -d ... ]` check so it's a no-op for slang branches that don't have integration tests.

### Context

- Slang PR #9738 adds neural module integration tests in `tests/integration/slangpy/neural/`
- Slang PR #10298 adds neural large-model stress tests in `tests/integration/slangpy/neural-large-model/`
- These tests currently have no CI coverage because:
  - Slang's embedded slangpy CI was removed by PR #9900 (replaced by cross-repo dispatch)
  - The cross-repo dispatch triggers `ci-latest-slang.yml`, but it only runs slangpy's own tests
  - `pip install slangpy` doesn't work because the released slangpy may not be compatible with the current slang being tested
- The `ci-latest-slang` workflow already clones the slang repo and builds slangpy against it, so adding this step has minimal cost (just a few seconds of pytest)

### Related

- shader-slang/slang#9738 — Neural.slang integration test placement
- shader-slang/slang#10298 — Neural.slang integration/stress tests
- shader-slang/slang#9900 — Decouple SlangPy CI testing via cross-repo dispatch

NV-xiaoyongs:
Fixes #845 

Add a step to discover and run slangpy-based integration tests from the slang repo checkout. These tests verify neural module functionality via slangpy and are maintained in the slang repo under tests/integration/slangpy/.

The step is guarded by a directory existence check so it's a no-op for slang branches without integration tests.

## Summary by CodeRabbit

* **Chores**
  * Added conditional execution of integration tests based on repository content detection in the build workflow.


NV-xiaoyongs:
> Can you clarify why these tests shouldn't be part of neural.slang's ci system? I'm happy to be convinced, I'm just currently concerned about this process - are we going to end up with a bespoke integration test system for every new slang module?
> 
> I'm not clear on why we aren't simply running neural.slang tests in neural.slang CI

**The concern is mainly about circular dependencies and cost.**
Running the slangpy tests that live in the slang repo inside slang CI introduces a couple of issues:

- Circular dependency / build cost

If we run these tests in slang CI, we either:

Depend on a released slangpy, or
Build slangpy from source as part of slang CI.

I experimented with the first option in this PR by adding a dedicated workflow:
https://github.com/shader-slang/slang/pull/9825
However, this approach is inherently limited to testing against the latest released slangpy. Building slangpy inside slang CI would significantly increase CI cost and complexity, which we want to avoid.


- Why slangpy CI is a better fit

- Running these tests in slangpy CI is much more straightforward:

  - slangpy already depends on slang, so there’s no circularity.
  - No additional build cost is introduced.
  - The tests still exercise the same integration path, just from the correct dependency direction.

This setup is now possible thanks to Janne’s recent changes:

  - https://github.com/shader-slang/slangpy/pull/780
  - https://github.com/shader-slang/slang/pull/9900

So the intent isn’t to create bespoke integration test systems per module, but rather to place integration tests in the CI of the repo that owns the dependency edge. In this case, that’s slangpy, not slang.
Let me know if that framing makes sense or if there’s a specific concern I’m missing.

Issue #600, opened 2025-10-24
[Feature Request] Upstream ImGui Render Backend Support


fangjunzhou:
Currently, slangpy implements a minimal ImGui support under https://github.com/shader-slang/slangpy/tree/main/src/sgl/ui. There is a window example in https://github.com/shader-slang/slangpy/tree/main/examples/window but only a small number of widgets are supported (basic ImGui components like buttons, sliders, etc.). Many advanced features in the upstream ImGui project as well as various third party libraries are not compatible with the SGL ImGui because SGL handles all the components in the context internally.

An alternative solution would be exposing the render pipeline to render arbitrary ImDrawData here: https://github.com/shader-slang/slangpy/blob/64d1e8b9e667cc8c12bab11b5349a799a5ea1e76/src/sgl/ui/ui.cpp#L386

I've implemented an SGL ImGui render backend in python here: https://github.com/fangjunzhou/slangpy-imgui-bundle/blob/main/src/slangpy_imgui_bundle/imgui_adapter.py and it now supports the full upstream ImGui features as well as all the third party libraries such as ImPlot3D, ImGuizmo, etc.

The following demo is rendered completely in SGL:

https://github.com/user-attachments/assets/6953bba9-8939-4f90-8e2e-dbe29badc52e

bmillsNV:
@kaizhangNV FYI on this.

MichaelMoroz:
The external imgui draw data support is merged into main, so I suppose this feature request can be closed?

PR #858, opened 2026-03-10
Add external ImGui draw data rendering and imgui_bundle integration

MichaelMoroz:
This PR adds support for rendering Dear ImGui draw data produced by external Python ImGui bindings (specifically `imgui_bundle`) through slangpy's existing UI rendering pipeline.

Slangpy currently provides a limited set of UI widgets, which is fine for basic examples but becomes a pain point for more complex projects. This change lets users leverage the full Dear ImGui widget set via `imgui_bundle` while still rendering through Slangpy's GPU backend.

This imlpements a feature requested by https://github.com/shader-slang/slangpy/issues/600

As an example I've been using this in my custom gaussian splat training pipeline written in Slang/Slangpy:
image

## Changes

**C++ core** (`src/sgl/ui/ui.h`, `src/sgl/ui/ui.cpp`):
- `DrawCommand`/`DrawList`/`DrawData` structs for marshalling external draw data across the Python/C++ boundary.
- `Context::render_draw_data()` to feed external draw data into the existing render pipeline.
- `Context::texture_id()` to obtain `ImTextureID` values for slangpy-managed textures.

**Python bindings** (`src/slangpy_ext/ui/ui.cpp`):
- `_render_marshaled_draw_data()` — marshals `imgui_bundle` Python draw data into the C++ structs and renders.
- `texture_id()` exposed on `ui.Context`.

**Python helpers** (`slangpy/ui/imgui_bundle.py`):
- `create_imgui_context()` — standalone offscreen ImGui context setup.
- `sync_draw_data_textures()` — uploads font/image atlases to GPU and registers texture IDs.
- `render_imgui_draw_data()` — high-level render helper.
- `handle_keyboard_event()` / `handle_mouse_event()` — forward `AppWindow` input events to ImGui.
- `begin_frame()` — frame setup (display size, delta time, `new_frame()`).

**Example & tests**:
- Offscreen rendering demo (`examples/imgui_bundle/imgui_bundle_offscreen.py`).
- Parametrized tests across D3D12/Vulkan/CUDA (`slangpy/tests/device/test_external_imgui.py`).

## Summary by CodeRabbit

* **New Features**
  * ImGui integration: runtime helpers for forwarding input, managing UI textures, syncing draw-data, and rendering ImGui draw calls to device targets.
* **Examples**
  * Added an offscreen example that renders an ImGui UI to an image for preview/export.
* **Tests**
  * End-to-end tests validating external ImGui rendering, input/event forwarding, texture lifecycle (creation/reuse/release), and pixel-level output verification.


slangpy/ui/imgui_bundle.py
```diff
@@ -0,0 +1,330 @@
...
+            texture = ui_context.get_texture(texture_id)
+            if texture is not None:
+                bpp = tex.bytes_per_pixel
+                if bpp not in _BPP_TO_FORMAT:
+                    raise ValueError(
+                        f"Unsupported bytes_per_pixel={bpp} for texture update at index {idx}"
+                    )
+                pixels = tex.get_pixels_array().reshape(
+                    (tex.height, tex.width, bpp)
+                )
```
aidanfnv: The slangpy tests are failing a pre-test formatting check here. This assignment needs to be on a single line, i.e.
```
pixels = tex.get_pixels_array().reshape((tex.height, tex.width, bpp))
```

slangpy/ui/imgui_bundle.py
```diff
@@ -0,0 +1,330 @@
...
+                continue
+
+        if status == imgui.ImTextureStatus.destroyed:
+            continue
+
+        bpp = tex.bytes_per_pixel
+        if bpp not in _BPP_TO_FORMAT:
+            raise ValueError(
+                f"Unsupported bytes_per_pixel={bpp} for texture at index {idx}"
+            )
```
aidanfnv: The slangpy tests are failing a pre-test formatting check here as well. This error needs to be on a single line, i.e.
```
raise ValueError(f"Unsupported bytes_per_pixel={bpp} for texture at index {idx}")
```

slangpy/ui/imgui_bundle.py
MichaelMoroz: Fixed.

slangpy/ui/imgui_bundle.py
MichaelMoroz: Fixed.

src/sgl/ui/ui.cpp
```diff
@@ -490,6 +492,303 @@ void Context::end_frame(Texture* texture, CommandEncoder* command_encoder)
     end_frame(texture->create_view({}), command_encoder);
 }
 
+void Context::render_draw_data(const DrawData& draw_data, TextureView* texture_view, CommandEncoder* command_encoder)
```
ccummingsNV: How much of this function is a copy/paste of the core end_frame? Ideally we share code between these where its reasonable, to avoid maintenance in future.

ccummingsNV (commented):
Hi Michael - just saw I didn't finish my review previously. I'm keen to get this useful work in, but my concern here is the level of code duplication in the UI which has the potential to create hard to maintain code with duplicate bugs in. 

Is it feasible to reduce duplication within the render function, or is this duplicated by design?

ccummingsNV (changes requested):

src/sgl/ui/ui.cpp
MichaelMoroz: Yes, for the most part `Context::end_frame` and `Context::render_draw_data` are quite similar with the exception of error handling and the way textures are provided. I was initially too afraid to modify already existing code paths since I didn't want to accidentally introduce any regressions in the current UI code. 
But in general the renderer code can be shared, `end_frame` can for example use a shared `render_draw_data` and marshall the internal UI draw data into a common DrawData.
The only concern I have is that `render_draw_data` is a user facing method, which means it requires much more thorough input validation, which `end_frame` currently does not do. If we merge these 2 paths native UI will also go though all these checks and safeties. 

For example a working solution is this:
```c++
void Context::end_frame(TextureView* texture_view, CommandEncoder* command_encoder)
{
    ImGui::SetCurrentContext(m_imgui_context);

    ImGui::Render();

    // Skip actual rendering device does not support it.
    if (!m_device->has_feature(Feature::rasterization))
        return;

    ImDrawData* imgui_draw_data = ImGui::GetDrawData();
    if (!imgui_draw_data)
        return;

    std::vector<std::vector<DrawCommand>> commands;
    std::vector<DrawList> draw_lists;
    commands.reserve(imgui_draw_data->CmdListsCount);
    draw_lists.reserve(imgui_draw_data->CmdListsCount);

    DrawData draw_data{
        .display_pos = float2(imgui_draw_data->DisplayPos.x, imgui_draw_data->DisplayPos.y),
        .display_size = float2(imgui_draw_data->DisplaySize.x, imgui_draw_data->DisplaySize.y),
        .framebuffer_scale = float2(imgui_draw_data->FramebufferScale.x, imgui_draw_data->FramebufferScale.y),
        .total_vtx_count = narrow_cast<uint32_t>(imgui_draw_data->TotalVtxCount),
        .total_idx_count = narrow_cast<uint32_t>(imgui_draw_data->TotalIdxCount),
        .index_size = sizeof(ImDrawIdx),
    };

    for (int i = 0; i < imgui_draw_data->CmdListsCount; ++i) {
        const ImDrawList* cmd_list = imgui_draw_data->CmdLists[i];
        auto& draw_commands = commands.emplace_back();
        draw_commands.reserve(cmd_list->CmdBuffer.Size);

        for (const ImDrawCmd& cmd : cmd_list->CmdBuffer) {
            SGL_CHECK(
                cmd.UserCallback == nullptr,
                "ImGui draw callbacks are not supported by the shared draw-data renderer"
            );

            draw_commands.push_back(
                DrawCommand{
                    .clip_rect = float4(cmd.ClipRect.x, cmd.ClipRect.y, cmd.ClipRect.z, cmd.ClipRect.w),
                    .elem_count = cmd.ElemCount,
                    .idx_offset = cmd.IdxOffset,
                    .vtx_offset = cmd.VtxOffset,
                    .texture_id = texture_id(static_cast<Texture*>(cmd.GetTexID())),
                }
            );
        }

        draw_lists.push_back(
            DrawList{
                .vertex_data = reinterpret_cast<uintptr_t>(cmd_list->VtxBuffer.Data),
                .vertex_count = narrow_cast<uint32_t>(cmd_list->VtxBuffer.Size),
                .index_data = reinterpret_cast<uintptr_t>(cmd_list->IdxBuffer.Data),
                .index_count = narrow_cast<uint32_t>(cmd_list->IdxBuffer.Size),
                .commands = std::span<const DrawCommand>(draw_commands.data(), draw_commands.size()),
            }
        );
    }

    draw_data.draw_lists = std::span<const DrawList>(draw_lists.data(), draw_lists.size());
    render_draw_data(draw_data, texture_view, command_encoder);
}

```
From what I've checked it seems to work fine with the old UI examples.

MichaelMoroz: Another concern I can see is that we now also register the internal UI textures via `texture_id`. 
I think its better to have `render_draw_data` take texture references directly, and only registering the textures when used from outside. 
I'll look into that.

MichaelMoroz: I actually think it might make sense to remove the texture registry from `Context` entirely. It was only really needed for storing the `imgui_bundle` textures somewhere, but it can just as well be done on the python side. In that case `DrawData` will hold `ref<Texture>` and `render_draw_data` will use it directly for rendering.

MichaelMoroz:
@ccummingsNV any news?

ccummingsNV (approved):
This looks good to me. @skallweitNV feel free to chip in, but it's good my end

src/sgl/ui/ui.h
```diff
@@ -10,15 +10,43 @@
...
+struct DrawCommand {
+    float4 clip_rect;
+    uint32_t elem_count;
+    uint32_t idx_offset;
+    uint32_t vtx_offset;
+    ref<Texture> texture;
+};
+
+struct DrawList {
+    uintptr_t vertex_data;
```
skallweitNV: Shouldn't these be pointer types? Like `const void*`?

src/slangpy_ext/ui/ui.cpp
```diff
@@ -24,6 +24,95 @@ struct GcHelper<ui::Context> {
 
 } // namespace sgl
 
+namespace {
+
+struct MarshaledDrawData {
+    std::vector<std::vector<sgl::ui::DrawCommand>> commands;
+    std::vector<sgl::ui::DrawList> draw_lists;
+    sgl::ui::DrawData draw_data;
```
skallweitNV: minor inconsistency, maybe `commands` should be `draw_commands`, or the three should just be `commands`, `lists` and `data`.

src/slangpy_ext/ui/ui.cpp
```diff
@@ -54,6 +143,28 @@ SGL_PY_EXPORT(ui)
             "command_encoder"_a,
             D(Context, end_frame, 2)
         )
+        .def(
+            "_render_marshaled_draw_data",
```
skallweitNV: I would probably just expose these as `render_draw_data`, because that's what they wrap. Maybe we can use `nb::sig` to provide a better type signature for the `draw_data` as well.

src/sgl/ui/ui.cpp
```diff
@@ -514,6 +468,270 @@ void Context::end_frame(Texture* texture, CommandEncoder* command_encoder)
...
```
skallweitNV: I think this is kinda reversed from what I would expect. We now convert to our internal draw command format even for the native imgui case. So we make the default path slower for no reason. I think the both `end_frame` and `render_draw_data` should call a draw implementation that operates on the ImGui types. The `render_draw_data` function would then convert into the ImGui draw command format instead of the other way around. This way, the direct path has zero conversions back and forth and we only do the conversion where it matters. Sounds reasonable? Or do I miss something?

skallweitNV (commented):
Sorry for chiming in so late. I do have a few comments. I'm mostly concerned about the extra conversion back and forth for the direct ImGui path in `end_frame`. I think that is worth addressing. Otherwise this is a nice addition and great to have!

ccummingsNV (changes requested):
Good work reducing code, but see Simon's comments before we merge this. He'd be more keen to flip this around unless you have a solid reason not to.

src/sgl/ui/ui.cpp
MichaelMoroz: Yeah, I did consider using the internal ImGUI draw data format, i.e. marshall outside drawdata into internal ImGUI drawdata when using from python. The only reason why I chose to stick with this approach is that conversion into third-party data types (ImDrawData) will be required for marshalling from python, and that might introduce compatibility issues any time ImGUI internal data structure changes on a version update, and perhaps might be less safe? 
But on the other hand, we will still need to update the rendering code if that happens regardless, and I do see that it would likely be both simpler and faster to use it directly.

src/sgl/ui/ui.h
MichaelMoroz: Yeah I agree, this should be `const void*`.

jkwak-work:
@MichaelMoroz , I don't see any changes after the conversation with Simon.
I assume it is not ready for a review yet.
Please leave a comment when this PR is ready for another review.

MichaelMoroz:
@jkwak-work yes, I haven't done the changes yet. I'll comment when its ready.

src/sgl/ui/ui.h
MichaelMoroz: Done.

src/slangpy_ext/ui/ui.cpp
MichaelMoroz: Done

src/sgl/ui/ui.cpp
MichaelMoroz: Done

MichaelMoroz:
@ccummingsNV @skallweitNV You can review now. I have refactored `render_draw_data` to directly take `ImDrawData`. This significantly simplifies the code in `‎src/sgl/ui/ui.cpp`.
* Now `end_frame` directly calls `render_draw_data` with the `ImDrawData` without any changes or copies.
* I removed most of the heavy input validation code from `‎src/sgl/ui/ui.cpp` and moved it to `src/slangpy_ext/ui/ui.cpp` as its not really needed for internal UI rendering and is faster.
* Now `src/slangpy_ext/ui/ui.cpp` takes the python input and directly validates and translates it into `ImDrawData`.
* I renamed the exposed python method from `_render_marshaled_draw_data` to just `render_draw_data`
* Also the shader change from https://github.com/shader-slang/slangpy/pull/949 tripped one of my tests which was made to pass with the old srgb blend, so I fixed that.

I've checked precommit and tests, all seems to pass. Internal UI works fine. I've also checked these changes on my personal project https://github.com/MichaelMoroz/slang-splat/tree/main which heavily uses this UI, and it works fine as well.
image

ccummingsNV (dismissed):
LGTM - thanks for all the iterations on this one!

ccummingsNV (approved):

PR #869, opened 2026-03-13
Move function call path from Python to C++ for cached calls

szihs:
- Add m_cache member to NativeFunctionNode with set_cache/resolve_cache
- Implement full_call() in C++ handling _result, _append_to, and error formatting
- Register full_call as __call__ via nanobind, eliminating Python frames on hot path
- Set cache pointer in Function.__init__, refresh on hot reload
- Remove Python __call__, simplify call() to delegate to C++ __call__
- Clean up dead code: _handle_error, _cache_value_to_id, unused imports

## Summary by CodeRabbit

* **New Features**
  * Functions are callable from Python using normal call syntax with a writable native cache pointer accessible for advanced use.

* **Bug Fixes**
  * Runtime call errors now include richer, reformatted messages with additional context to aid debugging.

* **Refactor**
  * Call and append behavior and cache lifecycle reworked to improve call reliability and ensure hot-reloads propagate to existing function wrappers.

## Before/After Timing

| Benchmark | Before (main) | After (this PR) | Improvement |
|---|---|---|---|
| slangpy_manual_hook | 0.158ms | 0.153ms | **+3.2%** |
| slangpy_automatic | 0.167ms | 0.164ms | **+1.8%** |

Measured with `pytest slangpy/benchmarks/test_benchmark_autograd.py` on CUDA, 1000 iterations after 10 warmups.

The improvement comes from eliminating Python frames on the cached call path — `__call__` now goes directly to C++ instead of through Python `call()` → `_native_call()`.

### DIFF
--- a/.github/actions/build-and-test-with-slang/action.yml
+++ b/.github/actions/build-and-test-with-slang/action.yml
@@ -179,9 +179,13 @@ runs:
       run: python -c "import pathlib, os; f=open(os.environ['GITHUB_OUTPUT'],'a'); f.write('has_tests=true\n' if pathlib.Path('slang/tests/integration/slangpy').is_dir() else 'has_tests=false\n'); f.close()"
 
     - name: Slang integration tests
-      if: steps.detect-slang-integration.outputs.has_tests == 'true'
+      if: steps.detect-slang-integration.outputs.has_tests == 'true' && startsWith(inputs.os, 'windows')
       shell: ${{ inputs.shell }}
-      run: python -m pytest slang/tests/integration/slangpy/ -ra -v
+      env:
+        NEURAL_MODULE_PATH: ${{ github.workspace }}/slang/build/${{ inputs.config }}/bin
+      run: |
+        python -m pip install Pillow
+        python -m pytest slang/tests/integration/slangpy/ -ra -v
 
     # Test Examples
     - name: Test Examples
--- a/slangpy/builtin/array.py
+++ b/slangpy/builtin/array.py
@@ -77,6 +77,24 @@ class ArrayMarshall(ValueMarshall):
         if as_array is not None:
             return [as_array]
 
+        # Support array-of-scalar-array matching array-of-vector
+        # e.g. float[4][2] (Python list [[1,2],[3,4],...]) -> float2[4]
+        if (
+            isinstance(bound_type, kfr.ArrayType)
+            and isinstance(bound_type.element_type, kfr.VectorType)
+            and (bound_type.num_elements == 0 or st.num_elements == bound_type.num_elements)
+        ):
+            as_vector = spyvec.array_to_vector_scalarconvertable(
+                cast(kfr.SlangType, st.element_type), bound_type.element_type
+            )
+            if as_vector is not None:
+                return [
+                    st.program.find_type_by_name(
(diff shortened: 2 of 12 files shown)


slangpy/core/function.py
```diff
@@ -253,74 +243,27 @@ def debug_build_call_data(self, *args: Any, **kwargs: Any):
...
         Debug helper to build call data without dispatching the kernel.
         """
+        cache = self._find_native_root()._native_cache
         return cast(
             "CallData",
-            self._native_build_call_data(self.module.call_data_cache, *args, **kwargs),
+            self._native_build_call_data(cache, *args, **kwargs),
         )
 
     def call(self, *args: Any, **kwargs: Any) -> Any:
```
ccummingsNV, line 262: If we can, I would prefer we maintain the same pattern, as before. This would mean:
- implementing 'call' in NativeCallData
- make __call__ call 'self.call' (either keeping '__call__' in python or making it native side)

I think this is just minor tweak - it's fairly standard practice where you have 2 ways of doing something, where one of them is the operator.

slangpy/core/function.py
```diff
@@ -664,6 +592,12 @@ def __init__(
         lines.append(str(self._options))
         self.slangpy_signature = "\n".join(lines)
 
+        # Store native cache pointer for fast C++ call path
+        self._native_set_cache(module.call_data_cache)
+
+        # Track this function so hot reload can update its cache pointer
```
ccummingsNV: This is a bit backwards - if the module wants to have a list of functions, it should add to that list when it creates/returns one.

We don't want a pattern where:
- a module creates a function, and calls its constructor
- the constructor calls back into the module to tell it about the function it just created

src/slangpy_ext/utils/slangpyfunction.cpp
```diff
@@ -211,6 +270,8 @@ SGL_PY_EXPORT(utils_slangpy_function)
         .def_prop_ro("_native_type", &NativeFunctionNode::type)
         .def_prop_ro("_native_data", &NativeFunctionNode::data)
         .def("_find_native_root", &NativeFunctionNode::find_root, D_NA(NativeFunctionNode, find_root))
+        .def("_native_set_cache", &NativeFunctionNode::set_cache, "cache"_a, D_NA(NativeFunctionNode, set_cache))
+        .def_prop_ro("_native_cache", &NativeFunctionNode::cache)
```
ccummingsNV: why have 'set_cache' and a read-only property instead of just a read-write property.

src/slangpy_ext/utils/slangpyfunction.cpp
```diff
@@ -181,6 +181,65 @@ nb::object NativeFunctionNode::call_bwds(NativeCallData* fwds_call_data, nb::arg
     return bwds_cd->call(options, args, kwargs);
 }
 
+nb::object NativeFunctionNode::full_call(nb::args args, nb::kwargs kwargs)
```
ccummingsNV: This naming of 'full_call' calling a thing called 'call' all feels a bit half done

I feel like what you're actually trying to switch to here is:
- the existing function's native side 'call' becomes 'invoke' (or similar)
- the python 'call' function becomes completely native, and is this 'full_call' (renamed to actually be 'call')
- the python '__call__' operator remains similar to original functionality - it's just syntactic suger for 'call'

PR #870, opened 2026-03-13
Kernel gen uses entry point args when possible

ccummingsNV:
Updated SlangPy kernel generation system so it can attempt to identify when it is safe to pass data directly via entry point arguments instead of a call data structure.

Few things to improve but wanted to do in future PR to avoid overlapping functionality:
- identifying max size for entry point uniforms is currently hardcoded - could be potentially read from slang-rhi
- the heuristic for when entry point args can be used is normally accurate but occasionally wrong. only way I've found to reliably handle this is to fallback if use of entry point args failed to compile. 
- slangpy.cpp and callsignature.py are getting big. will split these out into separate files in PR that doesn't affect functionality.

## Summary by CodeRabbit

* **New Features**
  * Automatic fast-path for passing entry-point uniforms with automatic fallback to parameter blocks when needed.
  * Exposed per-device maximum inline-uniform size to the API (max_entry_point_uniform_size).
  * Added a public toggle to opt into the entry-point-argument fast path.

* **Chores**
  * Removed the legacy CallDataMode enum and related mode property from the public API; mode selection is now handled by the new approach.

### DIFF
--- a/.github/actions/build-and-test-with-slang/action.yml
+++ b/.github/actions/build-and-test-with-slang/action.yml
@@ -23,6 +23,12 @@ inputs:
   slang-pr-number:
     description: "Slang PR number to checkout (pr mode)"
     default: ""
+  slang-ref:
+    description: "Slang ref to checkout (ref mode)"
+    default: ""
+  slang-repo:
+    description: "Slang repository to clone for ref mode"
+    default: "shader-slang/slang"
   shell:
     description: "Shell to use for run steps (bash on Linux/macOS, pwsh on Windows)"
     required: true
@@ -90,6 +96,16 @@ runs:
         git clone https://github.com/shader-slang/slang.git
         cd slang
         git fetch origin pull/${{ inputs.slang-pr-number }}/head
+        git checkout FETCH_HEAD
+        git submodule update --init --recursive
+
+    - name: Checkout Slang (ref)
+      if: inputs.slang-checkout-mode == 'ref'
+      shell: ${{ inputs.shell }}
+      run: |
+        git clone https://github.com/${{ inputs.slang-repo }}.git slang
+        cd slang
+        git fetch origin ${{ inputs.slang-ref }}
         git checkout FETCH_HEAD
         git submodule update --init --recursive
 
--- a/.github/workflows/ci-latest-slang.yml
+++ b/.github/workflows/ci-latest-slang.yml
@@ -1,5 +1,5 @@
 name: ci-latest-slang
-run-name: "${{ github.event_name == 'repository_dispatch' && format('Slang PR #{0}', github.event.client_payload.slang_pr_number) || format('Slang branch: {0}', github.event.inputs.slang_branch || 'master') }}"
+run-name: "${{ github.event_name == 'repository_dispatch' && (github.event.client_payload.slang_pr_number && format('Slang PR #{0}', github.event.client_payload.slang_pr_number) || format('Slang ref: {0}', github.event.client_payload.slang_ref || github.event.client_payload.slang_commit_sha)) || format('Slang branch: {0}', github.event.inputs.slang_branch || 'master') }}"
 
 on:
   schedule:
@@ -20,7 +20,7 @@ permissions:
   statuses: write
 
 concurrency:
-  group: ${{ github.event_name == 'repository_dispatch' && format('slang-pr-{0}', github.event.client_payload.slang_pr_number) || github.run_id }}
+  group: ${{ github.event_name == 'repository_dispatch' && format('slang-{0}', github.event.client_payload.slang_pr_number || github.event.client_payload.slang_commit_sha) || github.run_id }}
   cancel-in-progress: true
 
 jobs:
@@ -105,8 +105,10 @@ jobs:
           python: ${{ matrix.python }}
           flags: ${{ matrix.flags }}
           shell: ${{ matrix.os == 'windows' && 'pwsh' || 'bash' }}
-          slang-checkout-mode: pr
+          slang-checkout-mode: ${{ github.event.client_payload.slang_checkout_mode || 'pr' }}
           slang-pr-number: ${{ github.event.client_payload.slang_pr_number }}
+          slang-ref: ${{ github.event.client_payload.slang_ref }}
+          slang-repo: ${{ github.event.client_payload.slang_checkout_repo || 'shader-slang/slang' }}
 
   # Post aggregate status to Slang PR after all matrix jobs complete
   report-status:
--- a/.github/workflows/wheels-dev.yml
+++ b/.github/workflows/wheels-dev.yml
@@ -1,9 +1,14 @@
 name: wheels-dev
 
 on:
-  push:
-    branches: [main]
   workflow_dispatch:
+    inputs:
+      build_type:
+        description: 'dev: next dev version → Artifactory | release: current version → Artifactory'
+        required: true
+        default: 'dev'
+        type: choice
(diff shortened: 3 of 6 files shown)


slangpy/bindings/codegen.py
```diff
@@ -214,9 +221,10 @@ def finish(
         if call_data_structs:
             all_code = all_code + self.call_data_structs.code
             all_code.append("\n")
```
skallweitNV, line 223: shouldn't this be skipped if call data is not needed? like should this whole block be restructured around an initial `if not self.skip_call_data:`

slangpy/bindings/marshall.py
```diff
@@ -37,8 +36,8 @@ def __init__(
         #: Call mode (prim/bwds/fwds).
         self.call_mode = call_mode
 
-        #: Call data mode (global_data/entry_point).
-        self.call_data_mode = call_data_mode
+        #: Whether to use direct entry-point params (fast path) vs ParameterBlock<CallData> (fallback).
+        self.use_entrypoint_args = False
```
skallweitNV, line 40: in the other place you call this `skip_call_data` and here `use_entrypoint_args` .. it's kinda using two names that mean the same/similar thing. might be better to align the names.

slangpy/core/calldata.py
```diff
@@ -113,7 +113,7 @@ def __init__(
         build_info = func.calc_build_info()
         self.build(build_info, *args, **kwargs)
 
-    def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
+    def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any) -> None:
```
skallweitNV, line 116: i thought we typically don't have `-> None:` in the typings

slangpy/core/calldata.py
```diff
@@ -269,174 +259,58 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
...
-                print()
-
-            # Hash the code to get a unique identifier for the module.
-            # We add type conformances to the start of the code to ensure that the hash is unique
-            code_minus_header = (
-                "[CallData]\n" + str(build_info.type_conformances) + code[len(codegen.header) :]
+            # Determine fast path (entry-point params) vs fallback (ParameterBlock<CallData>).
+            # Sum inline-uniform byte size and compare against per-device threshold.
+            inline_size = estimate_entrypoint_arguments_size(bindings, self.call_dimensionality)
+            threshold = build_info.module.device.info.limits.max_entry_point_uniform_size
```
skallweitNV, line 265: where is `max_entry_point_uniform_size` coming from?

slangpy/core/calldata.py
```diff
@@ -269,174 +259,58 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
...
-                        miss_entry_points=build_info.ray_tracing_miss_entry_points,
-                        hit_group_names=hit_group_names,
-                        callable_entry_points=build_info.ray_tracing_callable_entry_points,
-                    )
-                    build_info.module.shader_table_cache[hash] = self.shader_table
-                else:
-                    raise RuntimeError("Unknown pipeline type")
-                self.device = device
-                self.log_debug(f"  Build succesful")
+            # Until https://github.com/shader-slang/slang-rhi/pull/676, Vk RTP can't use entry point args
```
skallweitNV, line 273: it's also questionable if the entrypoint params will end up faster in the shader table or as a param block .. might be worth measuring

slangpy/core/calldata.py
```diff
@@ -269,174 +259,58 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any):
...
+                build_info.pipeline_type == PipelineType.ray_tracing
+                and build_info.module.device.info.type == DeviceType.vulkan
+            ):
+                use_entrypoint_args = False
+
+            # Disable for Metal until I can figure out how entry point args work properly
+            if build_info.module.device.info.type == DeviceType.metal:
+                use_entrypoint_args = False
+
+            # Try building the shader. If direct args compilation fails (the
```
skallweitNV, line 284: this is a bit nasty, but i guess it's a good fallback to have always. we should just minimize the amount this triggers.

slangpy/core/callsignature.py
```diff
@@ -55,7 +55,7 @@ def is_generic_vector(type: TypeReflection) -> bool:
     try:
         if type.scalar_type != TypeReflection.Kind.none and type.col_count > 0:  # @IgnoreException
             return False
-    except:
+    except Exception:
```
skallweitNV, line 58: why this?

src/sgl/device/device.cpp
```diff
@@ -274,6 +274,27 @@ Device::Device(const DeviceDesc& desc)
     );
     m_info.limits.max_shader_visible_samplers = rhi_device_info.limits.maxShaderVisibleSamplers;
 
+    // TODO: These are known safe limits based on API spec, but could be increased based on
+    // platform (eg early Vk==128, CUDA 12.1+ supports 32k etc). Either this or the relevant
+    // information needs to be exposed by slang-rhi.
+    switch (m_desc.type) {
+    case DeviceType::vulkan:
+        // Vulkan spec minimum maxPushConstantsSize is 128 bytes.
```
skallweitNV, line 282: we should query that, in reality this might be much larger.

src/sgl/device/device.cpp
```diff
@@ -274,6 +274,27 @@ Device::Device(const DeviceDesc& desc)
...
+        m_info.limits.max_entry_point_uniform_size = 128;
+        break;
+    case DeviceType::d3d12:
+        // D3D12 root signature allows 64 DWORDs
+        m_info.limits.max_entry_point_uniform_size = 256;
```
skallweitNV, line 287: does slang put entry point parameters into the root signature by default? isn't it always using a descriptor set?

slangpy/core/calldata.py
ccummingsNV: yeah I hate it - it gets hit by 2 test runs in the whole 4k test bed at current

slangpy/core/calldata.py
ccummingsNV: hard coded, though my plan was to try and extract useful info from the lower level api. for example, vk can tell you the max push constant size

slangpy/bindings/codegen.py
ccummingsNV: yeah I am restructuring the generator code now - it was getting messy

slangpy/bindings/marshall.py
ccummingsNV: agreed - the old name is left over from when it was just cuda doing it, and only for a whole CallData structure. this is being cleaned now though

slangpy/core/callsignature.py
ccummingsNV: I am told by Claude (and have confirmed) that this is better practice, because certain key Python exceptions such as the one to break out of an iterator, or the one that fires on Ctrl+C, inherit from BaseException, not Exception

src/sgl/device/device.cpp
ccummingsNV: Yeah I'm going to add querying it into slang-rhi. Definitely not much bigger though or the ugly retry would never fire

src/sgl/device/device.cpp
ccummingsNV: I'll check

PR #872, opened 2026-03-13
Eliminate heap allocations from cached dispatch hot path

### LINKED ISSUE #690: Remove Heap Usage in Dispatch Path
## Estimated saving
1–3 µs

## Goal
Eliminate overhead from frequent short-lived allocations driven by Python-style APIs.

## Steps
- [ ] Implement either:
  - a global block allocator for fixed-size small allocations, or
  - a per-class pool allocator.
- [ ] Ensure implementation is lock-free and thread-safe.
- [ ] Remove temporary `std::vector` and `std::string` allocations in dispatch.

## Exit Criteria
- **Zero** heap allocations during typical dispatch.

szihs:
## Summary
- Stack-allocated `SignatureBuffer` (backed by `short_vector<uint8_t, 1024>`) replaces heap-allocated `SignatureBuilder` on the C++ fast path
- Stack-allocated `RuntimeOptions` struct replaces heap-allocated `NativeCallRuntimeOptions` (heap version only used for rare torch autograd path)
- `static_vector<nb::object, 4>` for `RuntimeOptions::uniforms` instead of raw C array + manual counter
- Transparent `string_view` lookup on signature cache avoids `std::string` copy on every cache hit
- Cache `CallContext` on `NativeCallData`, reuse via `reinitialize()`
- Skip unpack fast path: borrow args/kwargs directly when `m_needs_unpack == false`
- Extract `apply_torch_cuda_stream()` helper to deduplicate stream setup

Reduces per-call heap allocations from 6 to 1 (excluding RHI driver calls).

Closes #690

## Test plan
- [ ] Build on Windows (`cmake --build --preset windows-msvc-debug`)
- [ ] `pytest slangpy/tests -v` — all tests pass
- [ ] `pytest slangpy/benchmarks/test_benchmark_autograd.py -v` — no regression
- [ ] `pre-commit run --all-files`

## Summary by CodeRabbit

* **Refactor**
  * Faster, lower‑latency dispatch with reduced allocations and improved signature caching; streamlined runtime option handling and Python-facing bindings.

* **Bug Fixes**
  * More reliable GPU execution via cached call context and centralized CUDA/Torch stream setup.
  * Safer positional-argument handling to reduce edge-case dispatch errors and avoid unnecessary Python object copying.

* **API**
  * Simplified Python-facing call-context creation/initialization.

### DIFF
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -738,8 +738,11 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
     if (total_threads == 0) {
         // Still allocate and return the (empty) result if needed.
         if (!command_encoder && m_call_mode == CallMode::prim) {
-            NativeHandle cuda_stream = opts->cuda_stream();
-            auto context = make_ref<CallContext>(m_device, call_shape, m_call_mode, cuda_stream);
+            NativeHandle cuda_stream = opts.cuda_stream;
+            if (!m_cached_context)
+                m_cached_context = make_ref<CallContext>(m_device, m_call_mode);
+            m_cached_context->init(call_shape, cuda_stream);
+            auto& context = m_cached_context;
             ref<NativeBoundVariableRuntime> rv_node = m_runtime->find_kwarg("_result");
             if (rv_node && (!kwargs.contains("_result") || kwargs["_result"].is_none())) {
                 nb::object output = rv_node->python_type()->create_output(context, rv_node.get());
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -709,9 +709,7 @@ CallShapeInfo NativeCallData::compute_call_shape_info(
 nb::object
 NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_encoder, nb::args args, nb::kwargs kwargs)
 {
-    // Unpack args and kwargs (skip if no args have get_this/update_this).
-    // On the fast path (no unpack needed), avoid creating Python list/dict copies
-    // by using args (nb::tuple) and kwargs (nb::dict) directly.
+    // Skip unpacking when no args use get_this/update_this (avoids Python list/dict copy).
     nb::object unpacked_args;
     nb::dict unpacked_kwargs;
     if (m_needs_unpack) {
@@ -719,7 +717,6 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
         unpacked_args = unpack_args(args, had_unpack);
         unpacked_kwargs = unpack_kwargs(kwargs, had_unpack);
     } else {
-        // Fast path: borrow args/kwargs directly — zero allocation.
         unpacked_args = nb::borrow<nb::object>(args);
         unpacked_kwargs = nb::borrow<nb::dict>(kwargs);
     }
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -96,43 +96,15 @@ std::string NativeSlangType::to_string() const
     }
 }
 
-static constexpr std::array<char, 16> HEX_CHARS
-    = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'};
-
-void SignatureBuilder::add(const std::string& value)
-{
-    add_bytes((const uint8_t*)value.data(), (int)value.length());
-}
-void SignatureBuilder::add(const char* value)
-{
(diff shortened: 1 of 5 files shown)


src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -766,10 +746,14 @@ nb::object NativeCallData::exec(
     int total_threads = si.total_threads;
 
     // Extract CUDA stream handle for interop operations and command buffer submission.
-    NativeHandle cuda_stream = opts->cuda_stream();
+    NativeHandle cuda_stream = opts.cuda_stream;
 
-    // Setup context.
-    auto context = make_ref<CallContext>(m_device, call_shape, m_call_mode, cuda_stream);
+    // Setup context (reuse cached context to avoid heap allocation on repeat calls).
+    if (!m_cached_context)
```
ccummingsNV, line 763: To avoid cude duplication, if we want this approach we should have a constructor that just takes device and call mode, then an init that takes call shape and cuda stream. Having 2 ways to initialize call shape and cuda stream will make things harder to maintain.

src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -1668,15 +1645,20 @@ SGL_PY_EXPORT(utils_slangpy)
         )
         .def(
             "call",
-            &NativeCallData::call,
+            static_cast<nb::object (NativeCallData::*)(ref<NativeCallRuntimeOptions>, nb::args, nb::kwargs)>(
```
ccummingsNV: There's a lot of these static casts in bindings - what are they doing exactly? If it's to get around some odd templated/arg type issues it may be better solved with lambda functions as wrappers (though often once you do that, it becomes apparent the warnings are valid!).

src/slangpy_ext/utils/slangpy.h
```diff
@@ -20,6 +21,71 @@
...
+    SignatureBuffer& operator<<(const T& v)
+    {
+        add(v);
+        return *this;
+    }
+
+    std::string_view view() const { return {reinterpret_cast<const char*>(m_buf.data()), m_buf.size()}; }
+
+private:
+    short_vector<uint8_t, 1024> m_buf;
```
ccummingsNV, line 79: Does this really make it faster? I found that short vector was slower in many cases than explicitly writing that logic, as it does extra checks and handles none-POD types.

src/slangpy_ext/utils/slangpy.h
```diff
@@ -131,6 +176,10 @@ class NativeObject : public Object {
 
     virtual void read_signature(SignatureBuilder* builder) const { builder->add(m_signature); }
 
+    /// Write signature into a stack-allocated SignatureBuffer (hot path).
+    /// Non-virtual: delegates to the virtual read_signature(SignatureBuilder*) via a temporary wrapper.
+    /// For NativeObject base class, we can write directly without the virtual call.
+    void read_signature_fast(SignatureBuffer& builder) const { builder.add(m_signature); }
```
ccummingsNV: Should this just be 'read_signature' - i.e. is there any reason it's not just an overload?

src/slangpy_ext/utils/slangpyfunction.cpp
```diff
@@ -63,41 +71,36 @@ ref<NativeCallData> NativeFunctionNode::build_call_data(NativeCallDataCache* cac
...
-        NativeHandle stream_handle(NativeHandleType::CUstream, reinterpret_cast<uint64_t>(stream_ptr));
-        options->set_cuda_stream(stream_handle);
-    }
+    apply_torch_cuda_stream(call_data, options);
 
-    // If torch auto grad required, go via autograd hook
+    // If torch auto grad required, go via autograd hook (rare path — heap-allocate NativeCallRuntimeOptions)
     if (call_data->is_torch_autograd()) {
-        // Use TorchBridge to call the autograd hook - handles caching and cleanup
+        ref<NativeCallRuntimeOptions> ref_opts = NativeCallRuntimeOptions::from_runtime_options(options);
```
ccummingsNV: We've gone to a lot of trouble here to 'preserve' this nativecallruntimeoptions structure. If it is only needed in the autograd case, we may just be able to make it a none-ref counted type and pass it by value, rather than have this double-wrapper situation.

ccummingsNV (dismissed):
A few comments on where it'd be worth simplifying these changes, but by and large they look like the right ones. In future I'd try and split up large scale optimizations like this, as it's possible some of your changes made things slower, whilst others made it faster, but hard to tell with them all combined.

src/slangpy_ext/utils/slangpy.cpp
szihs: Ack

src/slangpy_ext/utils/slangpy.h
szihs: Yes it does

ccummingsNV (dismissed):
LGTM

szihs:
Update to resolve conflict with main branch

ccummingsNV (approved):

PR #894, opened 2026-03-24
Add `SlangSession::compose_modules` API

skallweitNV:
This PR adds a new `compose_modules` API to `SlangSession` and `Device`. This allows composing multiple Slang modules into a composed module, optionally with type conformances.

## Summary by CodeRabbit

* **New Features**
  * Compose multiple shader modules into a single composed module; APIs and Python bindings to compose, inspect composition state, and access source modules; entry-point creation supports type conformances.

* **Improvements**
  * Hot-reload and reflection invalidation scoped per device for more targeted updates; reflection objects expose their owner; hot-reload layout updates no longer rebuild linked programs.

* **Tests**
  * Added tests for module composition, type conformance behavior, entry-point discovery, linking composed modules, and hot-reload of composed sources.

* **Documentation**
  * Python docstrings updated to document composed-module and reflection APIs.

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -397,6 +397,10 @@ class CallData(NativeCallData):
         # Generate code.
         codegen = CodeGen()
         generate_code(context, build_info, bindings, codegen)
+        # TODO: This is not technically needed anymore because in the case of
+        # linking multiple modules, we already have composed the module to begin with.
+        for link in build_info.module.link:
+            codegen.add_import(link.name)
         code = codegen.finish(
             call_data=True,
             input_load_store=True,


slangpy/core/calldata.py
```diff
@@ -397,8 +397,6 @@ def _try_build_shader(
         # Generate code.
         codegen = CodeGen()
         generate_code(context, build_info, bindings, codegen)
-        for link in build_info.module.link:
```
ccummingsNV, line 400: I'm in 2 minds about removing this. It means the generated kernel is now 'further from' a correct kernel. i.e. it relies on special linking below. Unless it's causing harm, I'd rather leave it in - it makes reading the generated kernel a lot easier as slang's IDE highlights errors correctly.

src/sgl/device/reflection.cpp
```diff
@@ -17,6 +18,29 @@ namespace detail {
 
     static std::map<void*, const BaseReflectionObject*> g_slang_reflection_to_sgl_reflection;
 
+    /// Extract device from an owner object.
+    /// The possible owner types are known so we can query the device.
+    static Device* get_device_from_owner(const Object* owner)
```
ccummingsNV, line 23: I think this horror is why I ended up just invalidating all reflection data

ccummingsNV (changes requested):
Only point is it might be worth us leaving that code gen in. Otherwise this is the complexity / implementation I'd anticipated and all looks solid.

src/sgl/device/reflection.cpp
skallweitNV: Yeah but the side effect of not doing this is that you can get invalidation on an unrelated device. If we were lazily recreating the reflection data, that would be one thing. But we're not and just ending up with unrelated invalid reflection data. Do you suggest we remove this again?

ccummingsNV: We don't have to. My original design was based on the idea that it was (and always has been) fundamentally 'unsafe' to hold on to reflection data unless you go out of your way to handle a hot reload event. On that basis, it made just as much sense to blindly clear all data as to filter it by device - either you're being safe or you aren't.

But I'm happy to make it more precise as above - there's no 'clean' approach here without adding significant general perf overhead.

skallweitNV: Ok, I'll leave it in. We're only paying for this for hot-reload, which is fine IMO.

PR #898, opened 2026-03-30
Add torch.bool support for TensorView (#884)

### LINKED ISSUE #884: torch.bool tensors not supported in slangpy TensorView
Passing a torch.bool tensor to a slangpy kernel with TensorView<bool> parameter raises ValueError: Unsupported torch dtype torch.bool.
```
  out = torch.zeros(N, device="cuda", dtype=torch.bool)
  module.my_kernel(_thread_count=N, out=out)
  # ValueError: Unsupported torch dtype torch.bool
```

The dtype mapping in slangpy/torchintegration/torchtensormarshall.py does not include torch.bool. Supported dtypes are: `float32, float16, int32, int64, float64, int16, int8, uint8.`

Current Workaround: Use TensorView<int32_t> in .slang and dtype=torch.int32 in Python, cast back with .bool().

mkeshavaNV:
Fixes #884

Add bool dtype mapping across Python and C++ layers so that torch.bool tensors can be passed directly to TensorView<bool> Slang kernels without requiring a manual int32 workaround.

## Summary by CodeRabbit

* **New Features**
  * End-to-end boolean tensor support in the PyTorch integration, including CUDA zero-copy views and dtype mappings so boolean tensors are handled consistently for inputs, outputs, and operations.
* **Tests**
  * Added CUDA-only tests validating boolean tensor copy and logical-negation behavior across device types.


slangpy/tests/slangpy_tests/test_tensorview.py
```diff
@@ -137,6 +137,45 @@ def test_tensorview_grid_dispatch(device_type: DeviceType):
...
+def test_tensorview_negate_bool_torch(device_type: DeviceType):
+    """Test negate_tensorview_bool with torch.bool tensors."""
+    device = helpers.get_torch_device(device_type)
+    module = load_module(device)
+
+    input_tensor = torch.tensor([True, False, True, False, True], device="cuda", dtype=torch.bool)
+    output_tensor = torch.zeros(5, device="cuda", dtype=torch.bool)
+
+    module.negate_tensorview_bool(input_tensor, output_tensor)
+    torch.cuda.synchronize()
```
ccummingsNV: why is a sync needed here (and above?)

mkeshavaNV: 1. For sync: It is to just wait for the kernel to complete. Is there an issue with that? I remember adding these to  all tests in this file prior as well, just to be sure that the kernel is complete. (isn't this equivalent to a `glFinish` ?

2. For slangpy native Tensor consistency: The native Tensor does not support bool today. 
Test code:
```
import slangpy as spy
  import numpy as np
  device = spy.Device()
  arr = np.array([True, False, True], dtype=np.bool_)
  t = spy.Tensor.from_numpy(device, arr)
```
Tensor.from_numpy with np.bool_ fails with ```Unsupported numpy dtype bool``` 
because the _numpy_to_sgl mapping in slangpy/reflection/lookup.py doesn't include "bool". 

So maybe we should add the support for new Tensors as well?

ccummingsNV: We should certainly add it for the new Tensors too - we can make it a separate PR, but it'll need to be a high priority one. We can't have functionality that only works for TensorView in SlangPy - that's the start of a bad path :)

You shouldn't need to call synchronize - read back does that implicitly. 

A genuine need to call syncronize is pretty rare. Situations you might use it:
- a situation like shutdown, where you just want to make sure nothing is going on that you don't know about. 
- it can be useful if you're trying to verify if a problem is due to some race condition and want to see if it goes away when you wait for work to finish
- it can be useful to get a rough sanity check of gpu timings - i.e. sync gpu (so its idle), start timer, run some work, sync again, stop timer
- if we didn't have sync_to_device and sync_to_cuda (which syncronize the graphics pipes with cuda, not the cpu), it could be used to create hard sync points by blocking cuda on cpu before submitting work from the cpu

ccummingsNV: I'd remove the sync, and verify it works. If it does, let's get this in and add a high priority task to get the new Tensors handling bool as well.

mkeshavaNV: - Removed sync for all tests
- I filed https://github.com/shader-slang/slangpy/issues/899. I will check if we can get this planned in the coming sprint.

slangpy/tests/slangpy_tests/test_tensorview.py
```diff
@@ -137,6 +135,39 @@ def test_tensorview_grid_dispatch(device_type: DeviceType):
 # ============================================================================
 # Tests for TensorView<float2> / <float4> (vector element types)
```
skallweitNV: comment seems wrong now

mkeshavaNV: fixed

PR #906, opened 2026-03-31
Support reinterpreting torch.Tensor as Tensor<StructType, N> (#522)

### LINKED ISSUE #522: Support `reinterpret` of a torchTensor as `Tensor<MyStructType, N>`

NV-xiaoyongs:
Fixes: #522 

Add Tensor.from_torch() for struct tensor reinterpretation (#522)

Add a simple factory method that reinterprets a torch.Tensor as a slangpy Tensor
with struct element type by treating the last dimension as packed struct storage.
This reuses the existing Tensor marshalling pipeline with zero changes to type
resolution, codegen, or C++ dispatch.

Example usage:

    input = spy.Tensor.from_torch(device, torch_tensor, dtype=module.Vec2)
    output = spy.Tensor.from_torch(device, output_tensor, dtype=module.Vec2)
    module.my_kernel(input, output)

Where the Slang kernel uses structured access:

    void my_kernel(Tensor<Vec2, 1> input, WTensor<Vec2, 1> output) {
        Vec2 v = input[i];  // named field access: v.x, v.y
    }

Note: as stated in the parent issue, struct layout may vary between platforms
due to padding/alignment. The user is at their own risk when using this feature.

Tests: PackedFloat2 copy, wrong-dimension validation, Particle struct
(float3+float3) update. All pass on CUDA and Vulkan backends.

## Summary by CodeRabbit

* **New Features**
  * Added `Tensor.from_torch()` static method to convert PyTorch tensors to slangpy tensors with support for struct reinterpretation and custom buffer usage, enabling efficient tensor data interchange between frameworks.

* **Tests**
  * Added parametrized PyTorch integration tests for tensor struct reinterpretation, including byte-exact data validation and particle physics update computations.

### DIFF
--- a/slangpy/torchintegration/torchtensormarshall.py
+++ b/slangpy/torchintegration/torchtensormarshall.py
@@ -22,6 +22,8 @@ from slangpy.reflection.reflectiontypes import (
     ScalarType,
     VectorType,
     MatrixType,
+    StructType,
+    ITensorType,
     TensorType,
     TensorAccess,
     DiffTensorViewType,
@@ -180,6 +182,12 @@ class TorchTensorMarshall(NativeTorchTensorMarshall):
         """Resolve types during binding phase."""
         if isinstance(bound_type, DiffTensorViewType):
             return self._resolve_difftensorview(context, bound_type)
+        packed_struct_type = self._resolve_packed_struct_tensor(context, bound_type)
+        if packed_struct_type is not None:
+            return packed_struct_type
+        resolved_struct = self._resolve_bare_struct(bound_type)
+        if resolved_struct is not None:
+            return resolved_struct
         return spytc.resolve_types(self, context, bound_type)
 
     def _resolve_difftensorview(self, context: BindContext, bound_type: DiffTensorViewType):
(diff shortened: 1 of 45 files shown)


src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -482,8 +621,9 @@ void NativeTorchTensorMarshall::write_torch_tensor_fields(
     Buffer* interop_buffer
 ) const
 {
-    Shape shape = shape_from_bridge_info(info);
-    Shape strides = strides_from_bridge_info(info);
+    Shape shape;
+    Shape strides;
+    compute_effective_shape_and_strides(info, binding, shape, strides);
```
ccummingsNV: We can't make this code slower at all unfortunately - we measure the cost of dispatch in microseconds, and this is likely to add substantial overheads

slangpy/torchintegration/torchtensormarshall.py
```diff
@@ -197,6 +205,82 @@ def _resolve_difftensorview(self, context: BindContext, bound_type: DiffTensorVi
...
+    def _resolve_packed_struct_tensor(
+        self, context: BindContext, bound_type: SlangType
+    ) -> Optional[list[SlangType]]:
+        if not isinstance(bound_type, ITensorType):
+            return None
+        if not isinstance(bound_type.element_type, StructType):
+            return None
+        if bound_type.element_type.is_generic:
+            return None
+        if bound_type.tensor_type not in (TensorType.tensor, TensorType.itensor):
```
ccummingsNV: Does this mean the feature can't work for diff tensor?

ccummingsNV (changes requested):
I'm really not happy with the complexity or performance of this change. It seems like we're trying to do something relatively simple (treat the last dimension of a tensor as a struct), but have ended up with some very extremely complex type resolution logic and a lot of new torch specific duplicate logic - something we've worked pretty hard to get away from.

@csyonghe I can see this came from a request your end. Shall we discuss in SW meeting to see if we can come up with a way of writing it that balances ease of use against complexity/cost a little better?

src/slangpy_ext/utils/slangpytorchtensor.cpp
NV-xiaoyongs: Agreed — this is now removed entirely. The new approach has zero C++ changes. The torch tensor is converted to a slangpy Tensor before dispatch, so the existing native tensor marshalling path runs unchanged with no additional overhead.

slangpy/torchintegration/torchtensormarshall.py
NV-xiaoyongs: The new Tensor.from_torch() approach works with any tensor type the existing marshalling supports, including DiffTensor — the user just calls .with_grads() on the resulting Tensor as they normally would. No tensor type restriction needed.

ccummingsNV (approved):
Looks good - pity we have to copy, but I think that's a deeper change that isn't necessary for now

PR #907, opened 2026-03-31
Support structured numpy dtypes in NDBuffer and Tensor (#636)

### LINKED ISSUE #636: Numpy arrays  created with np.dtype can't be passed to slangpy

NV-xiaoyongs:
Fixes: #636 

Handle numpy arrays created with np.dtype (structured/record dtypes) that previously failed at the nanobind binding layer due to DLPack not supporting void dtype kinds. Structured arrays are now automatically reinterpreted as uint8 byte views in the Python layer before passing to C++.

- Add NDBuffer.copy_from_numpy override with structured dtype detection, size validation, and automatic uint8 reinterpretation
- Extend Tensor.from_numpy with optional dtype parameter for structured arrays
- Add clear error messages when structured dtypes are used without explicit Slang type (NDBuffer.from_numpy, Tensor.from_numpy, direct function calls)
- Add 18 test cases covering copy roundtrip, from_numpy, size mismatch detection, and error messaging on both Vulkan and CUDA backends

## Summary by CodeRabbit

* **New Features**
  * Tensor conversion now supports NumPy structured (named-field) arrays when an explicit target type is provided.
  * Added an optional parameter to specify the target type for structured-array conversions.

* **Bug Fixes**
  * Rejects structured NumPy input when no explicit target type is given to avoid unsafe conversions.
  * Validates layout/size compatibility to prevent silent data corruption.
  * Improved error messages guiding correct conversion workflow.


slangpy/types/tensor.py
```diff
@@ -160,14 +160,59 @@ def from_numpy(
...
+                )
+            dtype = numpy_to_slang(ndarray.dtype, device, program_layout)
+            if dtype is None:
+                raise ValueError(f"Unsupported numpy dtype {ndarray.dtype}")
+        else:
+            if not isinstance(dtype, SlangType):
+                program_layout = resolve_program_layout(device, dtype, program_layout)
+                dtype = resolve_element_type(program_layout, dtype)
+
+        if is_structured:
```
ccummingsNV, line 191: This is not a well maintainable modification to the initialization of the buffer - we've ended up with the classic addition to a function of:

```
if <some new beaviour>:
     do_similar_stuff()
     return it

<old behaviour>
return it
```

That's a recipe for bugs / confusing code. This if statement should be restructured to be:

```
if is_structured:
   <calculate information required to initialize from structured>
else
   <calculate information required to initialize from none-structured>
<initialize/return buffer>
```

i.e. we minimize logic duplication, making the code easier to maintain and parse visually.

slangpy/types/tensor.py
```diff
@@ -160,14 +160,59 @@ def from_numpy(
         usage: BufferUsage = BufferUsage.shader_resource | BufferUsage.unordered_access,
         memory_type: MemoryType = MemoryType.device_local,
         program_layout: Optional[SlangProgramLayout] = None,
+        dtype: Optional[Any] = None,
```
ccummingsNV: Can we call this 'target_slang_dtype' (or similar). i.e. make it very clear to the user what dtype refers to, given this is in the context of numpy, which has its own dtype

slangpy/types/buffer.py
```diff
@@ -94,6 +94,31 @@ def view(self, shape: TShapeOrTuple, strides: TShapeOrTuple = Shape(), offset: i
         """
         return super().view(Shape(shape), Shape(strides), offset)
 
+    def copy_from_numpy(self, data: np.ndarray[Any, Any]) -> None:
```
ccummingsNV: Let's drop the changes to buffer.py - it is due to be deprecated. We don't need to confuse the issue by expanding on it

slangpy/builtin/numpy.py
```diff
@@ -80,6 +80,12 @@ def gen_calldata(self, cgb: CodeGenBlock, context: BindContext, binding: "BoundV
 
 def create_vr_type_for_value(layout: SlangProgramLayout, value: Any):
     if isinstance(value, np.ndarray):
+        if value.dtype.names is not None:
+            raise ValueError(
+                f"Structured numpy dtype {value.dtype} cannot be passed directly to a Slang "
+                f"function. Use NDBuffer.from_numpy(device, data, dtype=...) or "
```
ccummingsNV: Remove the suggestion to use NDBuffer

slangpy/tests/slangpy_tests/test_structured_numpy.py
```diff
@@ -0,0 +1,264 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+Tests for issue #636: Numpy arrays created with np.dtype can't be passed to slangpy.
+
+Verifies that structured numpy arrays (with named fields) can be:
+  - Copied into NDBuffer via copy_from_numpy
```
ccummingsNV: Need to remove the NDBuffer references / tests, and where possible replace them with Tensor versions.

ccummingsNV (changes requested):
The feature makes sense but we need to remove the use of NDBuffer, and polish up the implementation for the tensor a bit

slangpy/tests/slangpy_tests/test_structured_numpy.py
NV-xiaoyongs: done

slangpy/builtin/numpy.py
NV-xiaoyongs: done

slangpy/types/buffer.py
NV-xiaoyongs: done

slangpy/types/tensor.py
NV-xiaoyongs: done

slangpy/types/tensor.py
NV-xiaoyongs: done

ccummingsNV (approved):
This looks good to me - can you review code rabbit's comments RE structured arrays before merging though

PR #915, opened 2026-04-02
Fix VRAM leak in PyTorch autograd backward pass

### LINKED ISSUE #896: Memory Leak in differentiable kernels during backward() using Pytorch interop
When running a simple optimization loop using `IDiffTensor` and `IWDiffTensor`, the VRAM usage seems to linearly increase until the system runs out of memory. Specifically, it seems to be the backwards call on the result that causes the leak every iteration.

Libraries:
slangpy commit 1ea6c4e (latest as of writing)
torch 2.10.0+cu130
*libraries were added using uv*

System:
Windows 11
RTX 3090 Ti
CUDA Version: 13.0

Example:
```slang
// example.slang
import slangpy;

[Differentiable]
void forward(uint index, IDiffTensor<float, 1> params, IWDiffTensor<float, 1> y)
{
    float w_i = params[index];
    y[index] = w_i * w_i;
}
```
```py
import slangpy as spy
import torch
import torch.nn as nn

spy_device = spy.create_torch_device(type=spy.DeviceType.cuda)
module = spy.Module.load_from_file(spy_device, "example.slang")

size = 256 * 256 # larger size leads to larger memory leak, but it happens even with a small size

y_true = torch.rand(size=(size,), dtype=torch.float32, device='cuda')

params = torch.rand(size=(size,), dtype=torch.float32, device='cuda', requires_grad=True)
y_pred = torch.zeros(size=(size,), dtype=torch.float32, device='cuda', requires_grad=True)

# memory leak happens regardless of using an optimizer or not
optimizer = torch.optim.Adam([params], lr=1e-3)

for _ in range(100000):
    optimizer.zero_grad()
    module.forward(index=spy.grid(shape=(size,)), params=params, y=y_pred)
    loss = nn.MSELoss()(y_pred, y_true)
    loss.backward()
    optimizer.step()
```

I also tried moving the output tensor initialization inside the loop and manually clearing the cache (e.g. `del loss, y_pred` and/or `torch.cuda.empty_cache()`), but the leak persisted. Installing and using the `slangpy-torch` also did not help. Commenting out the `loss.backward()` line stops the leak.

jhelferty-nv:
TorchAutoGradHook.backward() held references to function, call data,
args, kwargs, and diff pairs on the autograd ctx after the backward
pass completed. These kept GPU resources alive until the next forward
pass replaced them, causing VRAM to grow ~0.5MB per iteration.

Clear ctx attributes after backward finishes. Add a guard that raises
a clear error if backward is called twice on the same graph
(retain_graph=True), which is currently unsupported.

Fixes #896

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent errors when backward propagation is invoked multiple times on the same computation graph.
  * Improved memory management by explicitly releasing GPU and native resource references between iterations, enabling more efficient garbage collection.

### DIFF
--- a/slangpy/torchintegration/autogradhook.py
+++ b/slangpy/torchintegration/autogradhook.py
@@ -81,6 +81,13 @@ class TorchAutoGradHook(torch.autograd.Function):
         :param grad_outputs: Gradients for each output tensor.
         :return: Tuple of gradients for each input, with None for the options arg.
         """
+        if ctx.forwards_cd is None:
+            raise RuntimeError(
+                "TorchAutoGradHook.backward() called more than once on the same "
+                "graph. This can happen when backward(retain_graph=True) is used, "
+                "which is not supported by SlangPy's autograd integration."
+            )
+
         # Native C++ handles: tensor restoration, grad creation, bwds dispatch
         input_grads = ctx.forwards_cd.autograd_backward(
             ctx.function,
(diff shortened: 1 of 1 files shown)


ccummingsNV (approved):

jhelferty-nv:
Sounds good, thanks @szihs

PR #918, opened 2026-04-02
Add regression test for ITensor with [BackwardDerivative]

### LINKED ISSUE #639: Simple function taking ITensor argument segfaults in forward pass if a custom bwds function is defined
The function `simple` defined below will segfault if a backwards function is defined. It segfaults in the forward pass, not when running backwards, and does not segfault if using the generated derivatives via [Differentiable]. Here we wrap a call to the automatic backwards pass in a manual backwards definition, which should behave the same as if `process_element` were the main function.

```C++
[Differentiable]
float process_element(int coord, ITensor<float, 1> x) {
    return x[coord];
}

[BackwardDerivative(simple_bwd)]
float simple(int coord, ITensor<float, 1> x) {
    return process_element(coord, x);
}

void simple_bwd(int coord, ITensor<float, 1> x, float df) {
    bwd_diff(process_element)(coord, x, df);
}
```

The python code:
```py
N = 13
xt = spy.Tensor.empty(device, (N,), dtype=float).with_grads(zero=True)
xt.copy_from_numpy(np.linspace(-1, 1.1, N, dtype=np.float32))
f = slang_module.simple(coord=spy.grid((N,)), x=xt)
```

This segfaults consistently. If changing `[BackwardDerivative(simple_bwd)]` to `[Differentiable]`, it no longer segfaults.

This appears to happen on Metal as well as CUDA, if passing in a torch tensor.

After some investigation, it appears this can be partially mitigated by using concrete types (GradInOutTensor, RWTensor), but such a kernel will only work with `slangpy.Tensor` or `torch.Tensor`, respectively. Unfortunately, using concrete types doesn't seem to work with the backwards pass in pytorch, as the forward pass and backwards pass are called with different tensor types (`RWTensor` and `GradOutTensor`) and one cannot use different concrete types for the forward and backwards Slang definitions. The one example I have found (https://github.com/shader-slang/slangpy-samples/tree/main/examples/soft-rasterizer-pytorch) demonstrates torch interop using ITensor arguments.

jhelferty-nv:
Add regression test for slangpy#639. Tests forward and backward passes
for a function using ITensor interface parameters with
[BackwardDerivative], the pattern from the original issue report. The
underlying compiler crash was fixed by the auto-diff refactor in
shader-slang/slang#9808.

Fixes #639

## Summary by CodeRabbit

* **Tests**
  * Added regression test coverage for custom backward derivative support in differentiable tensor operations, validating forward computation correctness and backward gradient computation execution.


ccummingsNV (approved):
Looks good to me - I'm guessing it has to wait for fixed slang

jhelferty-nv:
Correct, we don't have a public Slang release with shader-slang/slang#9808 in it yet.

jhelferty-nv:
Moving this back to draft because it's taking a while to get a new slang update into slangpy with shader-slang/slang#9808.

PR #927, opened 2026-04-09
Added <=> for quat, matrix, and vector

tdavidovicNV:
Added `<=>` operator for matrix, vector, and quaternions. Without an explicit `<=>`, types containing these cannot have default generated `<=>` of their own.

## Summary by CodeRabbit

* **New Features**
  * Math types (vectors, matrices, quaternions) now support lexicographic three-way and relational comparisons, enabling ordering operations (<, >, <=, >=, <=>).

* **Tests**
  * Added comprehensive unit tests validating equality, inequality, lexicographic ordering, and three-way comparison behavior for vectors, matrices, and quaternions.

### DIFF
--- a/src/sgl/math/matrix_types.h
+++ b/src/sgl/math/matrix_types.h
@@ -7,6 +7,7 @@
 #include "sgl/core/error.h"
 
 #include <array>
+#include <compare>
 #include <initializer_list>
 #include <limits>
 
@@ -151,9 +152,6 @@ public:
             m_rows[r][col] = v[r];
     }
 
-    bool operator==(const matrix& rhs) const { return std::memcmp(this, &rhs, sizeof(*this)) == 0; }
-    bool operator!=(const matrix& rhs) const { return !(*this == rhs); }
-
 private:
     enum class Form {
         Undefined,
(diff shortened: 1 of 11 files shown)


src/sgl/math/matrix_types.h
```diff
@@ -184,6 +182,31 @@ class matrix {
...
+[[nodiscard]] auto operator<=>(const matrix<T, RowCount, ColCount>& lhs, const matrix<T, RowCount, ColCount>& rhs)
+{
+    for (int r = 0; r < RowCount; ++r) {
+        for (int c = 0; c < ColCount; ++c) {
+            auto cmp = lhs[r][c] <=> rhs[r][c];
+            if (cmp != 0)
+                return cmp;
+        }
+    }
+    return lhs[0][0] <=> rhs[0][0];
```
skallweitNV: This can't just return `0` I guess?

skallweitNV (dismissed):

src/sgl/math/matrix_types.h
tdavidovicNV: Lemme check.. I thought it required some declspec magic to determine which ordering is actually being returned, for the given underlying numeric types, but maybe it got better since I last tried.

tdavidovicNV: Reshuffled the code so we just return the honest comparison of the last elements, instead of explicit `equal`

skallweitNV (approved):

PR #934, opened 2026-04-10
Fix CUDA interop buffer race and add targeted CUDA↔graphics sync (#929)

### LINKED ISSUE #929: Interop buffer cleanup race: cuImportExternalMemory fails after async memset on Vulkan/D3D12 interop buffers
## Summary

When a `diff_pair(None, grad)` is dispatched through a Vulkan or D3D12 device with CUDA interop, `create_zeroed_interop_buffer` in `slangpytorchtensor.cpp` issues a `cuda::memset_device_async` on the interop buffer's CUDA-mapped memory. After dispatch completes, the interop buffer is destroyed — but nothing ensures the async CUDA memset has drained before the Vulkan/D3D12 memory backing the external mapping is freed.

On the next dispatch that creates a new interop buffer, `cuImportExternalMemory` fails with `CUDA_ERROR_ILLEGAL_ADDRESS` because the CUDA context was corrupted by the dangling async operation on the now-freed external memory.

## Tests to re-enable

Once this is fixed, the following test should be re-enabled:

- **`test_diffpair_get_shape_grad_only`** in `slangpy/tests/slangpy_tests/test_torchintegration.py` — currently `@pytest.mark.skip` (PR #928). Should be restored with `DEVICE_TYPES` parametrization (CUDA, Vulkan, D3D12). The test exercises `NativeTorchTensorMarshall::get_shape` falling back to grad shape when primal is None, which is otherwise untested. Even the CUDA-only + fallback bridge mode variant triggers the race.

## Reproduction

The issue is deterministic when two dispatches happen on the same cached Vulkan/D3D12 device. In CI, it manifests as:

1. `test_diffpair_get_shape_grad_only[native-DeviceType.vulkan]` — **PASSES** (first invocation)
2. `test_diffpair_get_shape_grad_only[fallback-DeviceType.vulkan]` — **CRASHES** with:

```
RuntimeError: CUDA call cuImportExternalMemory(&ext_mem, &desc) failed with error
CUDA_ERROR_ILLEGAL_ADDRESS (an illegal memory access was encountered).
sgl/device/cuda_utils.cpp:100 in function import_external_memory
```

The crash cascades to all subsequent CUDA operations on the same (and sometimes other) pytest-xdist workers, since `CUDA_ERROR_ILLEGAL_ADDRESS` is sticky and GPU-wide.

Tests that use `diff_pair(primal, grad)` (both tensors non-None) are unaffected because they go through `create_interop_buffer_from_tensor` instead of `create_zeroed_interop_buffer`.

## Root cause

`create_zeroed_interop_buffer` (slangpytorchtensor.cpp:564–583):

```cpp
void* cuda_ptr = interop_buffer->cuda_memory();
if (cuda_ptr && buffer_size > 0) {
    CUstream stream = context->cuda_stream().is_valid()
        ? reinterpret_cast<CUstream>(context->cuda_stream().value())
        : nullptr;
    cuda::memset_device_async(static_cast<uint8_t*>(cuda_ptr), 0, buffer_size, stream);
}
```

The async memset is launched but the interop buffer may be destroyed (releasing the Vulkan memory and its CUDA external memory mapping) before the memset completes on the CUDA stream.

## Suggested fix

Ensure a CUDA stream synchronization (or at minimum a stream-ordered fence) before destroying interop buffers that had async CUDA work issued against their mapped memory. Possible approaches:

1. Add a `cudaStreamSynchronize` in the interop buffer destructor when external CUDA memory is mapped.
2. Use `cudaFreeAsync` / stream-ordered cleanup so destruction is deferred until the stream catches up.
3. Track whether async CUDA work was issued against an interop buffer and only sync when necessary.

## Affected platforms

Observed on all CI runners with Vulkan (Linux gcc Debug/Release, Windows MSVC Debug/Release). D3D12 variants on Windows show a related `createBuffer` failure.

jhelferty-nv:
## Summary

- **Fix interop buffer lifetime race**: `diff_pair(None, grad)` caused `CUDA_ERROR_ILLEGAL_ADDRESS` because zeroed interop buffers were destroyed before async CUDA work completed. Now always store interop buffer refs in `read_back` to extend lifetime past the dispatch.
- **Fix CUDA direct path null-pointer crash**: On CUDA devices, `diff_pair(None, grad)` wrote a null device address into the shader. Now allocates a zeroed buffer and waits on the submit fence before freeing it.
- **Add targeted CUDA↔graphics synchronization**: Marshalls that perform CUDA work on shared interop buffers (via `copy_to_buffer` / `memset_device_async`) now record PyTorch's current CUDA stream on the `CallContext`. `exec()` passes this stream to `submit_command_buffer`, activating the existing `sync_to_cuda` / `sync_to_device` fence mechanisms only when needed.

Closes #929

## Changes

| File | What |
|------|------|
| `slangpytorchtensor.cpp` | Interop buffer keepalive in `read_back`; zeroed buffer for CUDA direct path; `mark_interop_stream` to record PyTorch's CUDA stream; fix `TensorView` path to use interop buffer address |
| `slangpy.cpp` | Capture `submit_id`; `wait_for_submit` on CUDA when `read_back` non-empty; pass interop CUDA stream to `submit_command_buffer` |
| `slangpy.h` | Add `mark_interop_cuda_stream()` / `interop_cuda_stream()` to `CallContext` |
| `test_torchintegration.py` | Add `test_diffpair_get_shape_grad_only` reproducer |

## Test plan

- [x] `test_diffpair_get_shape_grad_only` — all 4 variants (native/fallback x vulkan/cuda) pass
- [x] Full torch integration suite — 334 passed, 84 skipped, 0 failed
- [x] No runtime regression (625s, same as baseline)

## Summary by CodeRabbit

* **New Features**
  * Enabled gradient-only differentiation support for PyTorch–Slang interop.

* **Bug Fixes**
  * Prevented premature deallocation of temporary CUDA/graphics buffers during queued GPU work.
  * Improved CUDA stream selection and fallback for command submissions.
  * Ensured temporary buffer lifetimes cover async dispatch and read-back flows.

* **Tests**
  * Added tests validating gradient-only computation behavior across device types.

### DIFF
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -975,9 +975,14 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
     }
 
     // If we created a temporary command encoder, we need to submit it.
+    // Use the interop CUDA stream recorded by marshalls (if any) so that
+    // submit_command_buffers enables CUDA↔graphics synchronization.
     uint64_t submit_id = 0;
     if (temp_command_encoder) {
-        submit_id = m_device->submit_command_buffer(temp_command_encoder->finish(), CommandQueueType::graphics, cuda_stream);
+        NativeHandle submit_cuda_stream = cuda_stream;
+        if (!submit_cuda_stream.is_valid())
+            submit_cuda_stream = context->interop_cuda_stream();
+        submit_id = m_device->submit_command_buffer(temp_command_encoder->finish(), CommandQueueType::graphics, submit_cuda_stream);
         command_encoder = nullptr;
     }
 
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -975,8 +975,9 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
     }
 
     // If we created a temporary command encoder, we need to submit it.
+    uint64_t submit_id = 0;
     if (temp_command_encoder) {
-        m_device->submit_command_buffer(temp_command_encoder->finish(), CommandQueueType::graphics, cuda_stream);
+        submit_id = m_device->submit_command_buffer(temp_command_encoder->finish(), CommandQueueType::graphics, cuda_stream);
         command_encoder = nullptr;
     }
 
@@ -993,6 +994,14 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
         auto rb_val = t[1];
         auto rb_data = t[2];
         bvr->python_type()->read_calldata(context, bvr.get(), rb_val, rb_data);
+    }
+
+    // On CUDA devices, temporary buffers in read_back (e.g. zeroed primal for
+    // diff_pair(None, grad)) are referenced by raw device address — the RHI has
+    // no ref to defer their deletion. Wait for the submit to finish so the GPU
+    // dispatch is complete before read_back goes out of scope and frees them.
+    if (submit_id && m_device->type() == DeviceType::cuda && nb::len(read_back) > 0) {
+        m_device->wait_for_submit(submit_id);
     }
 
     // Pack updated 'this' values back (skip if no args needed unpacking).
--- a/src/slangpy_ext/utils/slangpy.cpp
+++ b/src/slangpy_ext/utils/slangpy.cpp
@@ -418,7 +418,7 @@ nb::object NativeCallData::find_torch_tensors_recurse(nb::object arg, nb::list& 
         // Read access from pre-built list
         if (access_idx >= m_autograd_access_list.size()) {
             throw std::runtime_error(
-                "Autograd access list index out of bounds — "
+                "Autograd access list index out of bounds - "
                 "argument structure doesn't match build-time bindings."
             );
         }
@@ -976,13 +976,17 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
 
     // If we created a temporary command encoder, we need to submit it.
     // Use the interop CUDA stream recorded by marshalls (if any) so that
-    // submit_command_buffers enables CUDA↔graphics synchronization.
+    // submit_command_buffers enables CUDA<->graphics synchronization.
     uint64_t submit_id = 0;
     if (temp_command_encoder) {
         NativeHandle submit_cuda_stream = cuda_stream;
         if (!submit_cuda_stream.is_valid())
             submit_cuda_stream = context->interop_cuda_stream();
-        submit_id = m_device->submit_command_buffer(temp_command_encoder->finish(), CommandQueueType::graphics, submit_cuda_stream);
+        submit_id = m_device->submit_command_buffer(
+            temp_command_encoder->finish(),
+            CommandQueueType::graphics,
+            submit_cuda_stream
+        );
         command_encoder = nullptr;
     }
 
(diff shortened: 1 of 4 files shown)


src/slangpy_ext/utils/slangpy.cpp
```diff
@@ -995,6 +1005,14 @@ NativeCallData::exec(NativeCallRuntimeOptions& opts, CommandEncoder* command_enc
         bvr->python_type()->read_calldata(context, bvr.get(), rb_val, rb_data);
     }
 
+    // On CUDA devices, temporary buffers in read_back (e.g. zeroed primal for
+    // diff_pair(None, grad)) are referenced by raw device address -- the RHI has
+    // no ref to defer their deletion. Wait for the submit to finish so the GPU
+    // dispatch is complete before read_back goes out of scope and frees them.
+    if (submit_id && m_device->type() == DeviceType::cuda && nb::len(read_back) > 0) {
```
ccummingsNV, line 1012: If this logic is correct, there is a serious problem deeper in the RHI - it should be impossible for resources to be freed before the gpu has finished with them. We can't add a cpu/gpu sync to solve this.

ccummingsNV: If the RHI is working correctly, then the requirement is that you should not allow the buffer object's destructor to be called until after the dispatch has been submitted. If this is done, the fence value will be tracked, and the actual device memory will not be freed until that submit is complete.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -461,8 +490,20 @@ void NativeTorchTensorMarshall::write_torch_tensor_fields(
     strides = apply_broadcast_stride_zeroing(strides, shape, binding->transform(), context->call_shape());
 
     if (offsets.is_tensorview) {
-        // TensorView path: build TensorViewData struct and write via set_data()
         TensorViewData tvd = populate_tensorview_data(info, shape, strides);
+        if (interop_buffer) {
```
ccummingsNV, line 494: We made the concious decision to say TensorView should not need to support interop, as slang treats it as a cuda only construct

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -393,31 +393,54 @@ void NativeTorchTensorMarshall::write_shader_cursor_pre_dispatch(
...
     } else {
-        // CUDA device - direct pointer access
+        // CUDA device - direct pointer access.
+        // When primal is None (backward pass, grad-only diff_pair), allocate a
+        // zeroed buffer so the shader has a valid device address instead of null.
+        // The buffer is stored in read_back; exec() will wait_for_submit on CUDA
+        // devices before read_back goes out of scope, ensuring the GPU dispatch
+        // is complete before the buffer is freed.
+        ref<Buffer> primal_zeroed_buffer;
+        if (primal_info.data_ptr == nullptr && primal_info.numel > 0) {
```
ccummingsNV, line 403: Is there no way for us to support a null binding and handle this. It doesn't feel right to me that if  a user has, eg, a 1GB tensor that we allocate 1GB of zeros to avoid passing nulls.

src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -538,8 +579,21 @@ void NativeTorchTensorMarshall::write_shader_cursor_with_interop(
...
 ) const
 {
-    // Helper: create interop buffer with given info. If tensor_value has data, copy from it;
-    // otherwise (e.g. backward pass output slot, primal is None) leave buffer uninitialized.
+    // Record PyTorch's current CUDA stream on the context so that exec() can
+    // pass it to submit_command_buffer for CUDA<->graphics synchronization.
+    // Called after any CUDA work on shared interop buffers (copy_to_buffer,
+    // memset_device_async). Only needed once per dispatch -- subsequent calls
+    // are no-ops if the stream is already recorded.
+    auto mark_interop_stream = [&](int32_t device_index)
```
ccummingsNV, line 587: This mechanism seems a bit shonky and has already suffered from errors relating to missmatched streams (note the one coderabbit highlights).

Is the correct fix not to simply add a device->sync_to_cuda after the copy, passing in the stream on which the copy was performed?

ccummingsNV (changes requested):
Quite a bit of work to go on this one I think - I don't think we've fixed the correct bits with the correct approaches. Have a read through my comments and see what you think. We can discuss on Teams if need be

PR #935, opened 2026-04-13
Remove dead NativeTorchTensorDiffPair::read_signature override

### LINKED ISSUE #923: NativeTorchTensorDiffPair::read_signature bypassed by SignatureBuffer optimization
## Summary

`NativeTorchTensorDiffPair::read_signature(SignatureBuilder*)` is never called during normal dispatch. PR #872 changed `get_value_signature` from taking `ref<SignatureBuilder>` to `SignatureBuffer&`, which changed the call at line 1074 from virtual dispatch (calling the override) to the non-virtual `NativeObject::read_signature(SignatureBuffer&)` overload that just reads the fixed `m_signature` string.

## Timeline

- **`73473aa2b`** (Chris Cummings) — Added `NativeTorchTensorDiffPair::read_signature(SignatureBuilder*)` to generate per-instance cache keys from tensor shape/dtype. At this point, `get_value_signature` took `ref<SignatureBuilder>`, so the virtual override was called correctly.
- **`77d5b0dfe`** (PR #872) — Changed `get_value_signature` to take `SignatureBuffer&` for performance. The call `native_object->read_signature(builder)` now resolves to the non-virtual `NativeObject::read_signature(SignatureBuffer&)` instead of the virtual `read_signature(SignatureBuilder*)`. Comment says "this is safe for all NativeObjects" but doesn't account for `NativeTorchTensorDiffPair`.

## Impact

All `NativeTorchTensorDiffPair` instances contribute an empty string to the cache key (since `m_signature` is never set). The cache still works because the Python type name differentiates DiffPairs from raw tensors, and `CallData` handles shape/dtype differences dynamically at dispatch time. However, the original intent was to produce distinct cache keys for different primal/grad tensor configurations.

## Affected code

**`slangpy.cpp` line 1074:**
```cpp
// get_value_signature(SignatureBuffer& builder, ...)
const NativeObject* native_object;
if (nb::try_cast<const NativeObject*>(o, native_object)) {
    builder << type_info.name() << "\n";
    native_object->read_signature(builder);  // Non-virtual: reads empty m_signature
    return;
}
```

**`slangpytorchtensor.cpp` line 164 — never reached:**
```cpp
void NativeTorchTensorDiffPair::read_signature(SignatureBuilder* builder) const
{
    *builder << "TorchDiffPair\n";
    // ... builds signature from primal/grad tensor shape/dtype
}
```

## Possible fixes

1. Make `read_signature(SignatureBuffer&)` virtual on `NativeObject`
2. Add a `read_signature(SignatureBuffer&)` override on `NativeTorchTensorDiffPair`
3. Have `get_value_signature` call the virtual `read_signature(SignatureBuilder*)` path for objects that don't have a fixed `m_signature`

Option 2 is probably the least invasive — keep the fast path non-virtual for most objects, but let DiffPair override it too.

szihs:
Fixes #923

The `read_signature(SignatureBuilder*)` override on `NativeTorchTensorDiffPair` was never called from `get_value_signature` because DiffPair objects are created AFTER the cache key signature is built. The call ordering has been this way since the override was first added (`73473aa2`):

1. `get_args_signature(builder, args, kwargs)` — signature built from raw torch tensors
2. `if is_torch_autograd()` → `find_torch_tensors(args, kwargs)` — DiffPairs created here

The cache key from the raw tensor (dtype, shape, strides) is sufficient for kernel selection. DiffPair primal/grad configuration only affects dispatch-time data binding, not which kernel gets compiled.

Removes the dead override (-38 lines).

## Summary by CodeRabbit

* **Refactor**
  * Removed signature-based cache key generation functionality for native torch tensor diff pairs.

### DIFF
--- a/src/slangpy_ext/utils/slangpytorchtensor.cpp
+++ b/src/slangpy_ext/utils/slangpytorchtensor.cpp
@@ -159,41 +159,6 @@ namespace {
 
 } // anonymous namespace
 
-// NativeTorchTensorDiffPair implementation
-
-void NativeTorchTensorDiffPair::read_signature(SignatureBuilder* builder) const
-{
-    // Write signature that combines both primal and grad tensor signatures
-    // This ensures that different primal/grad combinations get different cache keys
-    char buffer[128];
-
-    *builder << "TorchDiffPair\n";
-
-    // Add primal signature
-    // get_signature() returns 0 on success, non-zero on failure (does not throw)
-    if (!primal.is_none()) {
-        if (TorchBridge::instance().get_signature(primal.ptr(), buffer, sizeof(buffer)) == 0) {
-            *builder << "primal:" << buffer << "\n";
-        } else {
-            *builder << "primal:none\n";
-        }
-    } else {
-        *builder << "primal:none\n";
-    }
-
-    // Add grad signature
-    if (!grad.is_none()) {
-        if (TorchBridge::instance().get_signature(grad.ptr(), buffer, sizeof(buffer)) == 0) {
-            *builder << "grad:" << buffer << "\n";
-        } else {
-            *builder << "grad:none\n";
-        }
-    } else {
-        *builder << "grad:none\n";
-    }
-}
-
-
 NativeTorchTensorMarshall::NativeTorchTensorMarshall(
     int dims,
     bool writable,
--- a/src/slangpy_ext/utils/slangpytorchtensor.h
+++ b/src/slangpy_ext/utils/slangpytorchtensor.h
@@ -58,9 +58,6 @@ public:
     /// True if this is an input tensor, false if it's an output tensor.
     /// Determines which saved tensor list to index into.
     bool is_input = true;
-
-    /// Read signature for cache key generation
-    void read_signature(SignatureBuilder* builder) const override;
 };
 
 /// Native marshall for torch.Tensor objects.


src/slangpy_ext/utils/slangpytorchtensor.cpp
```diff
@@ -159,41 +159,6 @@ namespace {
 
 } // anonymous namespace
 
-// NativeTorchTensorDiffPair implementation
-
-void NativeTorchTensorDiffPair::read_signature(SignatureBuilder* builder) const
```
jhelferty-nv, line 164: @szihs  So the reason I didn't do this myself was I wanted someone to make sure that we weren't supposed to be using the `virtual` stuff after all. My thought had been, sure we might be able to remove this and the other path might work as-is, but is it less efficient than having this signature stuff hooked up?

szihs:
The override is unreachable because the call ordering has DiffPairs created AFTER signature building — confirmed by tracing the code at the original commit (73473aa2):

```
NativeFunctionNode::call()
  1. get_args_signature(builder, args, kwargs)   ← raw torch tensors
  2. if is_torch_autograd():
       find_torch_tensors(args, kwargs)          ← DiffPairs created here
```

The cache key from raw tensors already includes dtype/shape/strides via `TorchBridge::get_signature()`, which is what matters for kernel selection. DiffPair primal/grad configuration only affects dispatch-time data binding.

To make the override reachable, we'd need to move `find_torch_tensors` before signature building — a larger refactor with no correctness benefit since the raw tensor signature is sufficient.

src/slangpy_ext/utils/slangpytorchtensor.cpp
szihs: It would actually be less efficient to hook it up. The DiffPair config (which tensors have grads, input vs output) only affects buffer bindings at dispatch time — the compiled kernel is the same regardless. For example, `square(x)` with `requires_grad=True` and `requires_grad=False` compile to the same kernel — only the dispatch wrapper changes.

Adding DiffPair info to the cache key would create duplicate cache entries for identical kernels. The current separation (raw tensor signature for kernel selection, DiffPair for dispatch binding) is correct.

jhelferty-nv (approved):
Sounds good!

PR #936, opened 2026-04-13
Change missing slangpy-torch warning to error with opt-in fallback

### LINKED ISSUE #900: Design: visibility of user notification when missing extension
Review behavior when the pytorch integration extension isn't installed.

szihs:
Fixes #900
- Raise `RuntimeError` when torch tensors are used without `slangpy-torch` installed (was a `UserWarning` that was easy to miss)
- Differentiate between "missing" and "incompatible" `slangpy-torch` with tailored error messages
- Set `SLANGPY_ALLOW_TORCH_FALLBACK=1` to opt into the slower Python fallback
- Add `slangpy[torch]` optional dependency group in `pyproject.toml` (installs `slangpy-torch>=0.7.0`)
- Expose `get_torch_bridge_fallback_reason()` from the native layer to Python

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,9 @@ Repository = "https://github.com/shader-slang/slangpy"
 Issues = "https://github.com/shader-slang/slangpy/issues"
 Changelog = "https://slangpy.shader-slang.org/en/latest/changelog.html"
 
+[project.optional-dependencies]
+torch = ["slangpy-torch>=0.7.0"]
+
 [tool.setuptools]
 include-package-data = true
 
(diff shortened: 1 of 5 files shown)


szihs:
@ccummingsNV can we get the slangpy-torch installed on CI infrastructure?

src/slangpy_torch/pyproject.toml
```diff
@@ -21,6 +21,9 @@ authors = [
     { name = "Yong He", email = "yhe@nvidia.com" },
 ]
 license = {text = "Apache-2.0 WITH LLVM-exception"}
+dependencies = [
+    "slangpy>=0.40.0",
```
jhelferty-nv: I think @ccummingsNV was suggesting that the dependency should be the other way around, an optional dependency in the `slangpy` manifest on `slangpy_torch`.

szihs: Updated. @jhelferty-nv

jhelferty-nv (approved):
Looks good to me.

PR #942, opened 2026-04-15
Fix aarch64 wheel build: override nanothread -march=native under QEMU

kaizhangNV:
nanothread's cmake-defaults uses -march=native, which under QEMU aarch64 emulation expands to an architecture string with extensions (lse128, gcs) that the container toolchain's assembler doesn't support. Override with -march=armv8-a for portable aarch64 wheels.

Made-with: Cursor

## Summary by CodeRabbit

* **Chores**
  * Optimized compilation settings for ARM64 Linux builds to ensure consistent native flags during build.
  * Updated CI to run Linux ARM64 wheel builds on a native ARM runner instead of using QEMU emulation.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -44,16 +44,11 @@ jobs:
         include:
           # Specify runners
           - { os: windows, runs-on: [windows-latest] }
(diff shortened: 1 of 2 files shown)


szihs (approved):
LGTM

PR #945, opened 2026-04-18
Avoid forward-referencing generic parameters

juliusikkala:
I'm working on a [Slang PR](https://github.com/shader-slang/slang/pull/10836) that (in addition to its main goal) improves certain diagnostics with generic parameters. One of them is the pre-existing check for "forward-references" in generics, which already issues errors like "generic constraint for parameter 'T' references type parameter 'U' before it is declared" for cases like `<T : U, U>`. However, due to what I _think_ is an oversight, it only checked type parameters, not value parameters; things like `<T : Foo<U>, let U : int>` slipped through the checking.

When fixing that diagnostic, all Slang tests pass, but SlangPy tests do not, due to the line that this PR changes to not forward-reference `D`.

TL;DR changed order of generic params to fix what I suspect is invalid Slang code but just isn't caught by the compiler's checking yet.


csyonghe (approved):
I agree that this needs to be diagnosed as an error by the slang compiler.

juliusikkala:
@csyonghe it seems that I'm not able to merge this PR myself even with the approval, as I'm not authorized to do so in this repo (github gives me the "You're not authorized to push to this branch" note). Could you merge it for me?

ccummingsNV:
I have updated branch + set to auto merge once tests pass

PR #946, opened 2026-04-20
Tweak torch fallback

ccummingsNV:
- change optional dependency to == instead of >=
- add support for enabling fallback from code
- remove warning that fallback is being used, as user already opted into it

## Summary by CodeRabbit

* **New Features**
  * Exposed a top-level set_allow_torch_fallback() API to control torch fallback behavior at runtime.

* **Chores**
  * Pinned the optional torch dependency to version 0.7.0.
  * Updated packaging/workflow paths for the torch integration artifacts.

* **Tests**
  * Test fixtures now use the runtime setter to toggle torch-fallback instead of manipulating environment variables.


.github/workflows/slangpy_torch.yml
```diff
@@ -27,18 +27,18 @@ jobs:
         run: python -m pip install build twine
 
       - name: Build sdist
-        working-directory: slangpy_torch
+        working-directory: src/slangpy_torch
```
skallweitNV, line 30: So this just didn't work before?!

slangpy/core/calldata.py
```diff
@@ -163,36 +177,21 @@ def build(self, build_info: "FunctionBuildInfo", *args: Any, **kwargs: Any) -> N
...
-
-                            warnings.warn(
-                                "slangpy-torch native bridge not available. Using slower Python fallback. "
-                                "Install slangpy-torch for better performance: pip install slangpy[torch]",
-                                UserWarning,
-                                stacklevel=6,
-                            )
-                            _torch_bridge_warned = True
-                    else:
+                    if not _ALLOW_TORCH_FALLBACK:
```
skallweitNV, line 180: I'd probably do `if is_torch_bridge_using_fallback() and not _ALLOW_TORCH_FALLBACK:`

skallweitNV (approved):
LGTM, minor comments

.github/workflows/slangpy_torch.yml
ccummingsNV: Nope - looks like it'd been done manually before.

PR #948, opened 2026-04-20
Use ptrs for tensor storage (vulkan/metal)

fknfilewalker:
Update the internal storage of tensors to use pointers for the Vulkan/Metal backends. This change allows tensors to be packed, enabling them for dynamic dispatch for the Vulkan/Metal backend.

## Summary by CodeRabbit

* **Improvements**
  * Unified buffer/tensor handling across GPU backends (CUDA, Vulkan, Metal) for more consistent behavior.
  * Atomic operations on GPU backends now use byte-addressing stride handling for consistent cross-backend semantics.

* **Refactor**
  * Adjusted compile-time inclusion of bespoke tensor load extensions so Vulkan and Metal follow the same loading behavior as CUDA.

### DIFF
--- a/slangpy/slang/atomics.slang
+++ b/slangpy/slang/atomics.slang
@@ -103,7 +103,7 @@ public extension<S : IAtomicAddable, T : ISizedArray<S, D>, let D : int> T : IAt
 // store a buffer of atomics, and where appropriate uses IAtomicAddable to define the atomic add operation.
 public extension<T: IAtomicAddable> StorageTraits<T>
 {
-#ifdef __TARGET_CUDA__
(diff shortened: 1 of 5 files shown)


ccummingsNV:
I don't think we can do this yet as it has a subtle knockon we need to work through. Bindful resources are state tracked, but pointers (and bindless resources) are not. As a result, this will definitely allow for race conditions to be introduced.

We want to add the ability to provide (possibly by default) a CUDA style dispatch model in which a global memory barrier is inserted between each operation. This would then make sense, though I'd want us to expose that as a define and then only have tensors use ptr storage if it was safe to do so.

fknfilewalker:
I see. I assume doing automatic synchronization for ptrs is hard and requires some sort of memory range map. 
Maybe I can roll my own tensor type in the meantime.

PR #960, opened 2026-04-27
Add free-standing device API

skallweitNV:
- Add free standing device API (mirroring the members on `Device` class)
- Add thread-local device stack and `push_current_device`, `pop_current_device` and `current_device` API
- Add context manager for Python `Device`
- Creating a device automatically pushes it to the device stack (allowing free standing API to "just work")
- Closing a device automatically pops it from the device stack if at the top
- Matching push/pop semantics to what CUDA context handling does

## Summary by CodeRabbit

* **New Features**
  - Device context management system: push, pop, and retrieve the current device across operations
  - Device context manager support enabling Python `with` statement usage
  - Free-standing creation functions for resources (buffers, textures, surfaces, samplers, pipelines, kernels, etc.) that delegate to the current device

* **Tests**
  - Comprehensive test suite for device API functionality

### DIFF
--- a/slangpy/tests/device/test_device_api.py
+++ b/slangpy/tests/device/test_device_api.py
@@ -1,0 +1,176 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+
+import slangpy as spy
+from slangpy.testing import helpers
+
+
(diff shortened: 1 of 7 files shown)


ccummingsNV (dismissed):
Suggest making the first created device the 'default', but doesn't need to be part of this pr

ccummingsNV (approved):

Issue #943, opened 2026-04-16
Unify wheels and wheels-dev workflows into a single workflow


kaizhangNV:
## Problem

We currently maintain two separate wheel-building workflows that do largely the same thing:

- **`wheels.yml`** — builds release wheels, uploads to PyPI and GitHub Releases
- **`wheels-dev.yml`** — builds dev wheels, uploads to internal Artifactory

Both workflows duplicate the same build matrix, cibuildwheel configuration, runner selection, and dependency installation. When one is updated (e.g., adding native ARM runners, fixing build dependencies), the other must be manually kept in sync. This has already caused divergence — for example, `wheels-dev.yml` was using native `ubuntu-24.04-arm` runners while `wheels.yml` was still using QEMU emulation.

## Proposal

Unify both into a single reusable workflow with a `build_type` input (or similar) that controls:

- **Version string**: release uses `sgl.h` version as-is; dev computes `.devN` suffix
- **Upload target**: release goes to PyPI + GitHub Releases; dev goes to Artifactory
- **slangpy-torch builds**: currently only in `wheels-dev.yml`, could be conditionally included

The build matrix, runner selection, cibuildwheel settings, and dependency installation should be defined once.

## Current differences to reconcile

| Aspect | `wheels.yml` | `wheels-dev.yml` |
|--------|-------------|-----------------|
| Version | From `sgl.h` | `.devN` suffix |
| Upload | PyPI, GitHub Releases | Artifactory |
| slangpy-torch | Not built | Built (sdist + per-torch-version wheels) |
| `fail-fast` | default (true) | false |
| `fetch-depth` | default (1) | 0 (needed for dev version counting) |
| virtualenv caching | No | Yes |

PR #963, opened 2026-04-28
Unify wheels and wheels-dev workflows (#943)

szihs:
@/tmp/pr-body-v4.md


szihs:
@nv-slang-bot[bot] We need to do below 

Extract the .devN version computation (currently duplicated four times) into a composite action under .github/actions/compute-dev-version/.

What does the below mean?
> Add a tag-gate to the PyPI upload (current behavior preserves the existing logic).

PR #965, opened 2026-04-30
Work on improving bindings

ccummingsNV:


src/sgl/device/buffer_cursor.h
```diff
@@ -72,16 +73,30 @@ class SGL_API BufferElementCursor : public CursorWriteWrappers<BufferElementCurs
...
 
+    template<typename T>
+        requires(!is_ref_v<std::remove_cvref_t<T>> && HasWriteToCursor<T, BufferElementCursor>)
+    void set(const T& value) const
+    {
+        cursor_utils::write_to_cursor(*this, &value);
+    }
+
+    template<typename T>
+        requires(!std::is_const_v<T> && HasWriteToCursor<T, BufferElementCursor>)
```
tdavidovicNV: If feels really weird to have two sets, one for `ref<T>` and the other for `ref<const T>`, can we add a comment on this, if it really is needed.

src/sgl/device/cursor_utils.h
```diff
@@ -6,9 +6,34 @@
...
+#include <typeinfo>
+#include <utility>
+#include <vector>
 
 namespace sgl {
 
+namespace detail {
+    template<typename T>
+    struct CursorWriterOwner {
+        using type = std::remove_const_t<std::remove_cvref_t<T>>;
```
tdavidovicNV: Isn't cvref already removing const?

src/sgl/device/cursor_utils.h
```diff
@@ -65,6 +90,192 @@ namespace cursor_utils {
...
+        detail::CursorWriterOwner<T>::type::write_to_cursor(cursor, value);
+    }
+
+    namespace detail {
+
+        /// Return the compiler-specific spelling that contains T's name.
+        template<typename T>
+        constexpr std::string_view wrapped_type_name()
+        {
+#if SGL_MSVC
```
tdavidovicNV: This whole detail{} part feels like it belongs in platform, not here.

src/sgl/device/shader_object.h
```diff
@@ -36,15 +38,17 @@ class SGL_API ShaderObject : public Object {
...
-    virtual void set_buffer(const ShaderOffset& offset, const ref<Buffer>& buffer);
-    virtual void set_buffer_view(const ShaderOffset& offset, const ref<BufferView>& buffer_view);
-    virtual void set_texture(const ShaderOffset& offset, const ref<Texture>& texture);
-    virtual void set_texture_view(const ShaderOffset& offset, const ref<TextureView>& texture_view);
-    virtual void set_sampler(const ShaderOffset& offset, const ref<Sampler>& sampler);
-    virtual void
-    set_acceleration_structure(const ShaderOffset& offset, const ref<AccelerationStructure>& acceleration_structure);
+    virtual void set_object(const ShaderOffset& offset, const ref<const ShaderObject>& object);
+
+    virtual void set_buffer(const ShaderOffset& offset, const ref<const Buffer>& buffer);
```
tdavidovicNV, line 43: should these be const for symmetry with buffer?

src/sgl/device/types.cpp
```diff
@@ -2,10 +2,26 @@
...
+{
+    SGL_CHECK(value, "Cannot write a null descriptor handle pointer to a shader cursor.");
+    cursor.set_descriptor_handle(*value);
+}
+
+void DescriptorHandle::write_to_cursor(const BufferElementCursor& cursor, const DescriptorHandle* value)
+{
+    SGL_CHECK(value, "Cannot write a null descriptor handle pointer to a buffer cursor.");
+    cursor.set_data(&value->value, sizeof(value->value));
+}
```
tdavidovicNV, line 24: Do these two actually do different thing under the hood, or is that just name asymmetry?

tdavidovicNV (dismissed):
Overall I like it, couple symmetry points (I really wish we could say those two cursors have the same interface) and couple code location notes.

src/sgl/device/buffer_cursor.h
ccummingsNV: Collapsed this to a single overload. `ref<const T>` now deduces `T = const T`, and the writer lookup normalizes via `remove_cvref_t`.

src/sgl/device/cursor_utils.h
ccummingsNV: Fixed. `remove_cvref_t` already removes const, so the extra `remove_const_t` is gone.

src/sgl/device/cursor_utils.h
ccummingsNV: Moved the compiler-specific type-name helper out of cursor utils into the core type utilities.

src/sgl/device/shader_object.h
ccummingsNV: Leaving these non-const for now. Unlike cursor writes, these calls mutate the `ShaderObject` binding state, so making them const would be misleading even though the surface API is similar.

src/sgl/device/types.cpp
ccummingsNV: They are intentionally different. The shader cursor path binds the descriptor handle through RHI, while the buffer cursor path writes the raw descriptor payload layout. I added a comment to make that explicit.

src/sgl/func/tensor.cpp
```diff
@@ -28,11 +267,65 @@ Device* Tensor::device() const
...
+        shape.size()
+    );
+
+    TensorViewData data = {};
+    data.data = storage->device_address() + static_cast<uint64_t>(offset) * element_stride;
+
+    for (size_t i = 0; i < shape.size(); ++i) {
+        data.strides[i] = static_cast<uint32_t>(static_cast<size_t>(strides[i]) * element_stride);
+        data.sizes[i] = static_cast<uint32_t>(shape[i]);
+    }
```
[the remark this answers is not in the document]
ccummingsNV: Added checked stride conversion here: negative strides are rejected, the `size_t` multiply is guarded, and narrowing now goes through `narrow_cast<uint32_t>`.

tdavidovicNV (approved):

PR #967, opened 2026-04-30
A wrapper for Cursors that allows us to write virtual write_to_cursor

tdavidovicNV:
This wraps the common interface of BufferElementCursor and ShaderCursor, allowing us to a write one virtual interface that takes this generic cursor, rather than having to have two. This is a native code only, as Python can already do that.

## Release Notes

* **New Features**
  * Added polymorphic cursor abstraction providing a unified interface for different cursor types
  * Added writable interface enabling custom cursor-writing implementations

* **Tests**
  * Expanded test coverage for polymorphic cursor writing scenarios and type interactions

### DIFF
--- a/src/sgl/CMakeLists.txt
+++ b/src/sgl/CMakeLists.txt
@@ -69,6 +69,8 @@ target_sources(sgl PRIVATE
     core/window.h
 
     device/agility_sdk.h
+    device/any_cursor.cpp
+    device/any_cursor.h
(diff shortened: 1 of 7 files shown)


ccummingsNV (commented):
Can we hold off on this until I've finished rewriting half the cursor code :)

ccummingsNV:
close in favour of newer cursor implementations

PR #974, opened 2026-05-11
Fix setting up raytracing pipeline entry points

skallweitNV:
- Filter out duplicate entrypoints
- Filter out OptiX builtin entrypoints

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -58,6 +58,14 @@ _ALLOW_TORCH_FALLBACK = os.environ.get("SLANGPY_ALLOW_TORCH_FALLBACK", "").lower
     "1",
 )
 
+# OptiX built-in intersection shaders that must not be resolved as user entry points.
+_OPTIX_BUILTIN_INTERSECTION_SHADERS = frozenset(
+    {
+        "__builtin_intersection__sphere",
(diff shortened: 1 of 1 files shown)


slangpy/core/calldata.py
```diff
@@ -519,28 +519,38 @@ def _try_build_shader(
...
-                        )
-                    if hit_group.intersection_entry_point != "":
-                        eps.append(
-                            build_info.module.device_module.entry_point(
-                                hit_group.intersection_entry_point
-                            )
-                        )
+                    if (
+                        hit_group.closest_hit_entry_point != ""
+                        and hit_group.closest_hit_entry_point not in entry_point_names
```
ccummingsNV: Aren't these extra checks in this function redundant? And empty string is Falsey. I'd have thought this could be:
if hitgroup.closest_hit_entry_point:
    entry_point_name.add(hitgroup.closest_hit_entry_point)

wouldn't even need the if, if you just ignore the empty string at the end

ccummingsNV (changes requested):
Logic looks correct but might be worth checking if you can simplify that code - I think Python probably lets you write a much simpler/more efficient version

ccummingsNV (approved):

PR #979, opened 2026-05-13
ci: add diagnostic watchdog to install-slangpy-torch

jvepsalainen-nv:
## Summary

The 2026-05-12 nightly (run [25708068653](https://github.com/shader-slang/slangpy/actions/runs/25708068653)) spent 1h25m wedged inside `pip install ./src/slangpy_torch --no-build-isolation` at `Preparing metadata (pyproject.toml): started` before the job-level timeout cancelled it, producing **zero diagnostic output**. pip captures the PEP 517 build-backend's stdio by default and only flushes it on failure — a hang isn't a failure, so the captured output never reached the log. The same wedge recurred on the run's attempt 2 on the same runner (`2u1g-b650-0468`).

This PR makes the next occurrence diagnosable, without cluttering the healthy-case log.

## Changes

`tools/ci.py` — `install_slangpy_torch()`:

- `pip -vvv` so pip relays the PEP 517 build-backend's stdio live instead of buffering it
- `PYTHONUNBUFFERED=1` so any Python subprocess in the build chain flushes stdio immediately
- `PIP_NO_INPUT=1` so pip never blocks on an interactive prompt
- On GitHub Actions, wrap the verbose output in a collapsible `::group::` so the normal log stays as short as it was on `main`, with the diagnostics one click away
- Surface pip's own `Successfully installed slangpy-torch-X.Y.Z` (or `ERROR: ...` on failure) outside the collapsed group, so the visible log still shows the result without expanding

`tools/ci.py` — `run_command()`:

- On non-zero exit, attach the captured stdout to the raised `RuntimeError` as `captured_output` so callers can inspect what pip actually printed. Additive change; existing callers are unaffected.

No behavior change on healthy installs (~2s, same line count in the visible log as on `main`).

## Test plan

- [x] `python tools/ci.py --help` still lists `install-slangpy-torch`
- [x] Smoke-tested env-merging and the `captured_output` attribute on success and failure paths
- [x] Smoke-tested that the `::group::`/`::endgroup::` markers wrap the verbose output and that pip's pass/fail summary line appears outside the group
- [ ] Next nightly that exercises this step finishes normally; visible log shows only `Successfully installed slangpy-torch-X.Y.Z` plus a collapsed `▶ pip install slangpy-torch (verbose)` toggle
- [ ] If a future hang triggers, the log contains live `setuptools` / `setup.py` output up to the wedge point (inside the group)

### DIFF
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -279,7 +279,7 @@ def install_slangpy_torch(args: Any):
         # Surface pip's own pass/fail summary outside the collapsed group
         # so the result is visible without expanding.
         for line in out.splitlines():
-            if line.startswith("Successfully installed") or line.startswith("ERROR"):
+            if line.startswith(("Successfully installed", "ERROR")):
                 print(line)
 
 
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -98,7 +98,7 @@ def run_command(
     process.communicate()
     if process.returncode != 0:
         err = RuntimeError(f'Error running "{command}"')
-        err.captured_output = out
+        err.captured_output = out  # type: ignore[attr-defined]
         raise err
 
     return out
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -245,9 +245,19 @@ def install_slangpy_torch(args: Any):
     cmd = [sys.executable, "-m", "pip", "install", "wheel"]
     run_command(cmd)
 
-    # Use --no-build-isolation to compile against the user's installed PyTorch
-    cmd = [sys.executable, "-m", "pip", "install", str(slangpy_torch_dir), "--no-build-isolation"]
-    run_command(cmd)
+    # -vvv so pip relays the PEP 517 build-backend's stdio live (a hang isn't
+    # a failure, so captured output is never flushed); env vars prevent stdio
+    # buffering and any interactive prompt that could wedge the install.
+    cmd = [
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "-vvv",
+        str(slangpy_torch_dir),
+        "--no-build-isolation",
+    ]
+    run_command(cmd, env={"PYTHONUNBUFFERED": "1", "PIP_NO_INPUT": "1"})
 
 
 def main():
--- a/tools/ci.py
+++ b/tools/ci.py
@@ -247,7 +247,9 @@ def install_slangpy_torch(args: Any):
 
     # -vvv so pip relays the PEP 517 build-backend's stdio live (a hang isn't
     # a failure, so captured output is never flushed); env vars prevent stdio
-    # buffering and any interactive prompt that could wedge the install.
+    # buffering and any interactive prompt that could wedge the install. On
+    # GitHub Actions, wrap the verbose output in a collapsible log group so
+    # the normal case stays readable while diagnostics remain one click away.
     cmd = [
         sys.executable,
         "-m",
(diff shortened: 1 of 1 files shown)


.github/actions/build-and-test-with-slang/action.yml
```diff
@@ -148,7 +148,11 @@ runs:
       shell: ${{ inputs.shell }}
       run: python tools/ci.py build
 
-    # Install slangpy-torch extension (requires MSVC on Windows, PyTorch installed)
+    # Install slangpy-torch extension (requires MSVC on Windows, PyTorch installed).
```
ccummingsNV: AI got over-keen with comment here - let's restore it to the sensible 1 liner :)

.github/workflows/ci.yml
```diff
@@ -200,9 +200,13 @@ jobs:
       - name: Build
         run: python tools/ci.py build
 
-      # Install slangpy-torch extension (requires MSVC on Windows, PyTorch installed)
+      # Install slangpy-torch extension (requires MSVC on Windows, PyTorch installed).
+      # timeout-minutes is a defense-in-depth cap above the in-script watchdog
```
ccummingsNV: Same

tools/ci.py
```diff
@@ -102,6 +104,206 @@ def run_command(
...
+            os.kill(child_pid, 9)
+        except Exception:
+            pass
+    try:
+        os.kill(pid, 9)
+    except Exception:
+        pass
+
+
+def run_with_silence_watchdog(
```
ccummingsNV: This seems like a pretty epic amount of extra machinery just to debug why pip didn't run

ccummingsNV (changes requested):
I've only reviewed ci.py + the workflow as it looks like the rest is just merge issues.

I think this PR is very overcomplex for what you're trying to do - we've seen 1 hang in pip - I don't think that puts us at the point where we need to be adding complex code to list process trees.

why not start with just -vvv, PYTHONUNBUFFERED and PIP_NO_INPUT, and see if that gets us what we need first?

ccummingsNV:
Ping on this one - what would you like to do? Close it or merge/simplify it?

ccummingsNV:
Hi - is this dead? We can close if you aren't working on it.

jvepsalainen-nv:
Hi - this got buried under other tasks, sorry for delay.
This looks over-complicated to be a simple watchdog. I'll simplify this.

tools/ci.py
```diff
@@ -102,6 +104,206 @@ def run_command(
...
+    `silence_timeout` seconds, dump diagnostics about the process tree
+    (ps/pstree/py-spy) so the next hang doesn't disappear into a silent
+    job-level timeout. After `total_timeout` seconds, kill the process tree
+    and return its exit code.
+
+    Returns the subprocess's return code (or a non-zero value on watchdog kill).
+    Does NOT raise on non-zero exit; the caller decides whether to fail.
+    """
+    if isinstance(command, str):
+        command = [command]
```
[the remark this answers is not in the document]
jvepsalainen-nv: Moot — `run_with_silence_watchdog` removed in this revision.

tools/ci.py
```diff
@@ -102,6 +104,206 @@ def run_command(
...
+            sys.stdout.flush()
+            if not diagnostics_dumped:
+                _dump_process_diagnostics(process.pid)
+            _kill_process_tree(process.pid)
+            break
+        time.sleep(5)
+
+    reader_thread.join(timeout=10)
+    process.wait()
+    return process.returncode
```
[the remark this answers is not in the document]
jvepsalainen-nv: Moot — `run_with_silence_watchdog` removed in this revision.

tools/ci.py
jvepsalainen-nv: It was. the whole watchdog/process-dump path removed

.github/actions/build-and-test-with-slang/action.yml
jvepsalainen-nv: It did.. removed

.github/workflows/ci.yml
jvepsalainen-nv: Yes, fixed

jvepsalainen-nv:
Updated the PR based on comments and resolved the conflicts.

ccummingsNV (changes requested):
Much more sensible, but looks like its logging lots of errors now - may as well get those fixed (or hidden if not valid) before merging?

jvepsalainen-nv:
Fixed logging. Now the verbose log is hidden but the result is shown without collapsing the hidden part.

ccummingsNV (approved):
Ok - this looks good

Do you have a plan for when to remove it? I'd rather we don't dump loads of diagnostic spam into the CI by default unless it's genuinely useful.

ccummingsNV:
Shall we close this now?

PR #989, opened 2026-05-18
Native conversion

ccummingsNV:
First pass moving slangpy python side code to native. This focusses on moving a substantial amount of the reflection code to pure native, with minimal changes to the python api. 

Currently implemented with 2 additional native namespaces:
- sgl::refl  - reflection types moved from slangpy
- sgl::func - start of a 'functional' namespace, currently just contains BaseModule and BaseStruct

These are bound to python submodules native_refl and native_func, parts of which are then imported into the correct pure python level modules. 

The systems moved over in this PR include the various reflection wrappers, the layout wrapper, and various auxiliary functions for accessing / looking up reflection info.

The common shared layout object is now also directly owned by Device, avoiding the need for a global cache.


skallweitNV:
@ccummingsNV I do agree with the overall migration here. This PR is too big to really dig into the details though but I'm ok with it if you are :)

skallweitNV (approved):

PR #991, opened 2026-05-19
wheels: unify nightly/release workflows + drop find-links HTML

### LINKED ISSUE #943: Unify wheels and wheels-dev workflows into a single workflow

szihs:
## Summary

Unifies the `wheels.yml` and `wheels-dev.yml` workflows into a single `wheels.yml` with a `build_type` dropdown, and eliminates the static find-links HTML pages in favor of Artifactory's auto-generated directory listings.

  Closes #943.

  ## Background

  We had two separate wheels workflows that drifted apart, and the slangpy-torch distribution relied on hand-written HTML find-links pages at `whl/torch-*.html`. Those pages had fixed names, so every re-run of the workflow tried to overwrite them — but our CI token only has DEPLOY (not DELETE), and Artifactory requires both to overwrite an existing artifact. Result: the find-links upload step has been failing since the original WIP unification went in.

This PR fixes the unification properly and removes the source of the overwrite problem altogether.

  ## Changes

  - **Unify the two workflows** into one `wheels.yml` with a `build_type` dropdown — nightly / release / build-only.
  - **Extract `.devN` computation** into a reusable composite action  (`.github/actions/compute-dev-version`).
  - **Drop the find-links HTML generation step.** Point users at the auto-generated Artifactory directory listing instead — pip's
    `--find-links` parses it identically.
  - **Unify `release` mode** to atomically publish to PyPI + GitHub Release + Artifactory in one workflow run. The old `internal` mode was redundant with this and has been removed. `nightly` covers the pre-release staging use case via `.devN` builds.
  - **Simplify the `build_type` dropdown descriptions** so contributors don't need to know about `.devN` vs `sgl.h` to pick the right mode.

  ## New build_type matrix

  | | slangpy wheel | slangpy-torch wheels | sdist | → PyPI | → GitHub Release | → Artifactory |
  |---|---|---|---|---|---|---|
  | **nightly** | ✓ `0.x.x.devN` | ✓ `0.x.x.devN` | ✓ | ✗ | ✗ | ✓ |
  | **release** | ✓ bare | ✓ bare | ✓ | ✓ | ✓ | ✓ |
  | **build-only** | ✓ bare | ✗ | ✗ | ✗ | ✗ | ✗ |


ccummingsNV (approved):
I'm not confident I can look at this and confirm it works, but it all seems sane to me, so if you're happy, looks good.

We should maybe change the torch versions that get built for now though? Could be a different PR though to avoid confusion.

PR #996, opened 2026-05-23
Add Tests for Array of `Tensor`, `RWTensor`, `DiffTensor`, `RWDiffTensor`

fangjunzhou:
This pull request expands the test coverage for array handling in `slangpy`, particularly focusing on arrays of tensor types and their variants. It also removes device-specific test skips for Metal, as related issue https://github.com/shader-slang/slang/issues/7606 has been closed.

**Expanded test coverage for arrays of tensor types:**

* Added `test_array_of_tensors_read` to verify correct summation over an array of `Tensor<float, 1>` parameters.
* Added `test_array_of_rwtensors_write` to test in-place modification of an array of `RWTensor<float, 1>` parameters.
* Added `test_array_of_difftensors_read` to check reading from an array of `DiffTensor<float, 1>` parameters (primal pass).
* Added `test_array_of_rwdifftensors_write` to test writing to an array of `RWDiffTensor<float, 1>` parameters (primal write).

The tests are expected to fail on all non-CUDA platform on current main branch. Changing the Tensor _data to T* according to ca2d5bab0010521e47a3b59719c02b77d1032ab0 fixed `test_vectorize_struct_with_tensor_array`, `test_2d_mapped_vectorize_struct_with_tensor_array`, `test_array_of_tensors_read`, and `test_array_of_rwtensors_write`. But `test_array_of_difftensors_read` and `test_array_of_rwdifftensors_write` still fail.

PR #997, opened 2026-05-23
Fix shader cursor tensor write for DiffTensor

fangjunzhou:
Resolved the array diff tensor issue in https://github.com/shader-slang/slangpy/pull/996

### DIFF
--- a/src/slangpy_ext/device/cursor_utils.h
+++ b/src/slangpy_ext/device/cursor_utils.h
@@ -681,9 +681,50 @@ private:
         auto kind = (TypeReflection::Kind)type_layout->getKind();
 
         // Read uniforms for NativeTensor unless it is being written directly to a pointer.
+        //
+        // For plain tensor targets, we pass the standard flat uniforms dictionary:
+        //   {_data, _shape, _offset, _strides}
+        //
+        // For DiffTensor-like targets, the destination layout is a struct with nested
+        // fields (_primal, optional _grad_in/_grad_out). In that case, adapt the
+        // source NativeTensor into a nested dictionary so recursive struct writing
+        // populates the correct fields.
         if (kind != TypeReflection::Kind::pointer && nb::isinstance<sgl::slangpy::NativeTensor>(nbval)) {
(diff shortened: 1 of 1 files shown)


ccummingsNV (changes requested):
This is a good piece of work, but the shader cursor binding path is performance critical, and I think this code needs refining in order to avoid a performance degredation.

As it stands, the code effectively performs the original 'tensor->uniforms()' call, then does some none-trivial (but necessary) new work, the falls through to the original code path. This will mean any dispatches involving tensors - even those not involving diff tensors - will be measurably slower. As we measure dispatch perf in microseconds this will likely be noticeable.

If we want to do this, I suggest that the path in question should bypass the tensor->uniforms() call altogether, and instead:
- use cursor.find_field_by_name to find the _data field.
- if it exists, bind data, shape, strides etc as normal
- if it doesn't, assume this is a diff tensor, find the _primal field, and proceed to bind

This should probably be done using a helper function on the NativeTensor object.

If this represents too big a piece of work, a 'low tech' approach would be to add a 'diff_uniforms' to tensor, and require the user python side to pass 'diff_uniforms' when binding. i.e. you convert to a python dictionary in python, and pass that to a cursor, rather than passing the tensor to a cursor. Less elegant but would work.

I am working on some significant changes to NativeTensor at the moment so can build this work in, though it won't be immediate.

szihs:
@nv-slang-bot can you update with the fix as Chris mentioned.
```
If we want to do this, I suggest that the path in question should bypass the tensor->uniforms() call altogether, and instead:

use cursor.find_field_by_name to find the _data field.
if it exists, bind data, shape, strides etc as normal
if it doesn't, assume this is a diff tensor, find the _primal field, and proceed to bind
This should probably be done using a helper function on the NativeTensor object.

```

szihs:
Note - Will close this issue once #996 passes - Triggered the CI

jhelferty-nv:
It looks like this PR was made obsolete by #1000 which made the change Chris mentioned above, so I'm going to go ahead and close this one.

Regardless, thanks @fangjunzhou, your contribution is appreciated.

PR #998, opened 2026-05-24
more recent cmake version

ccummingsNV:


ccummingsNV:
[self-approve: minor cmake version change]

PR #1000, opened 2026-05-26
Native tensor

ccummingsNV:


src/sgl/utils/slangpy.h
```diff
@@ -98,6 +99,12 @@ class SGL_API Shape {
         }
     }
 
+    /// Constructor from a vector of dimensions.
+    Shape(const std::vector<int>& shape)
```
skallweitNV, line 103: might also take a `std::span<const int>` instead, but if all sources are vectors then this is probably fine too

skallweitNV (dismissed):
LGTM without going into too much detail.

PR #1030, opened 2026-06-25
Route CUDA kernels with runtime-indexed resource-array params through ParameterBlock

szihs:
## Summary

On CUDA the functional API picks entry-point args (EPA) vs `ParameterBlock<CallData>` (PB) by inline-uniform size against the device `max_entry_point_uniform_size` (CUDA: 4096B), so on CUDA almost everything takes EPA. For a fixed-size resource/tensor array indexed at **runtime**, EPA packs it into the `.param` bank and the runtime index becomes a serial dynamic-address `ld.param` chain — 12–24× slower than the PB lowering (`__constant__` pointer + cached global).

This adds a targeted heuristic: when a binding's resolved type transitively contains a fixed-size array of resources/tensors, force `use_entrypoint_args=False` on CUDA.

## Measurements (NVIDIA L40S, `test_tensor_sum_indirect`, median ms)

| count | baseline | this PR | speedup |
|---|---:|---:|---|
| 16 | 0.744 | 0.031 | 24× |
| 32 | 2.585 | 0.199 | 13× |

Direct (compile-time-indexed) `test_tensor_sum` unaffected; `test_tensor_add_simple` unchanged. Tiny direct `sum<N>` regress ~5–9µs absolute (the heuristic keys on array *shape*, an accepted trade).

## Test plan
- [ ] New regression test `test_resource_array_forces_parameter_block_on_cuda` (`slangpy/tests/slangpy_tests/test_code_gen.py`)
- [ ] CI green

A principled compiler-side fix (a Slang auto-hoist pass + a slang-rhi binding change) is in progress upstream to cover user-written kernels too; this host heuristic is the immediate SlangPy win.

<sub>🤖 Generated by an automated SlangPy coworker — may be inaccurate. A human maintainer should verify.</sub>

### DIFF
--- a/slangpy/core/calldata.py
+++ b/slangpy/core/calldata.py
@@ -3,7 +3,7 @@ import hashlib
 import os
 import re
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional
 
 from slangpy.core.callsignature import *
 from slangpy.core.logging import bound_call_table, bound_exception_info, mismatch_info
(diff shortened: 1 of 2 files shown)


szihs:
Was just test for heuristic runs.

PR #1063, opened 2026-07-13
Profiler

skallweitNV:
Adds a bounded CPU/GPU instrumentation profiler to SlangPy.
- Provides C++ macros and Python context managers for profiling frames, functions, and zones.
- Automatically profiles functional API dispatches and supports GPU timestamps for command-buffer work.
- Adds live frame statistics, bounded captures, diagnostics, trace queries, and Perfetto-compatible JSON export.
- Includes an ImGui profiler window, path tracer integrations, documentation, and comprehensive C++/Python tests.

### DIFF
--- a/src/sgl/utils/profiler.cpp
+++ b/src/sgl/utils/profiler.cpp
@@ -552,7 +552,7 @@ struct ProfilerImpl {
         }
 
         auto data = std::make_unique<ThreadData>(desc.thread_event_capacity);
-        const uint64_t thread_id = std::hash<std::thread::id>()(std::this_thread::get_id());
+        const ThreadID thread_id = platform::current_thread_id();
         ThreadData* result = data.get();
         {
             std::lock_guard lock(thread_mutex);
--- a/src/sgl/utils/profiler.cpp
+++ b/src/sgl/utils/profiler.cpp
@@ -1,0 +1,2220 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "profiler.h"
+
+#include "sgl/core/error.h"
+#include "sgl/core/platform.h"
+#include "sgl/core/thread.h"
+#include "sgl/device/command.h"
+#include "sgl/device/device.h"
+#include "sgl/device/query.h"
+
+#include <algorithm>
+#include <array>
+#include <cctype>
+#include <chrono>
+#include <cmath>
+#include <condition_variable>
+#include <deque>
+#include <fstream>
+#include <iomanip>
+#include <mutex>
+#include <numeric>
+#include <sstream>
+#include <thread>
+#include <unordered_map>
+
+namespace sgl {
+
+namespace {
+
+    constexpr uint32_t ZONE_CHUNK_SIZE = 4096;
+    constexpr uint32_t MAX_ZONE_DEPTH = 64;
+    constexpr uint32_t GPU_ZONES_PER_CHUNK = 16;
+    constexpr uint32_t INVALID_INDEX = ~0u;
+    constexpr uint64_t GPU_CALIBRATION_INTERVAL_NS = 2'000'000'000ull;
+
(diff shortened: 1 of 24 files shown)


src/sgl/utils/profiler.h
```diff
@@ -0,0 +1,590 @@
...
+    /// Metadata is interned and remains valid for the lifetime of the process.
+    static uint32_t
+    register_site(std::string_view file, uint32_t line, std::string_view function, std::string_view name);
+
+    /// Whether manual and automatic profiling zones are recorded.
+    bool enabled() const { return m_enabled.load(std::memory_order_relaxed); }
+    /// Enable or disable all zone and frame recording.
+    void set_enabled(bool value) { m_enabled.store(value, std::memory_order_relaxed); }
+    /// Whether SlangPy functional dispatches create automatic zones.
+    bool enable_auto_zones() const { return m_enable_auto_zones.load(std::memory_order_relaxed); }
```
ccummingsNV, line 439: I typically favour calling these things like 'auto_zones_enabled', so we don't end up with accessors that look like they do things - i.e. a user seeing this would assume 'enable_auto_zones' meant turning auto zones on

src/sgl/utils/profiler.h
```diff
@@ -0,0 +1,590 @@
...
+    ref<ProfilerFrameStats> frame_stats_snapshot() const;
+    /// Release profiler-owned references to snapshots previously handed to Python.
+    /// Python bindings call this while holding the GIL before acquiring a new snapshot.
+    void release_retired_snapshots();
+    /// Clear completed and pending frame statistics after flushing current producer events.
+    void clear_frame_stats();
+    /// Poll submitted GPU timestamp queries and queue resolved measurements for the collector.
+    ///
+    /// The poll does not deliberately wait for profiled submissions or unresolved queries. Refreshing timestamp
+    /// calibration can nevertheless synchronize queued GPU work on some backends, notably CUDA.
```
ccummingsNV: Do you know which backends specifically - if so the comment should just say that

src/sgl/utils/profiler.cpp
```diff
@@ -0,0 +1,2240 @@
...
+        token.start_ns = start_ns;
+        token.frame_index = *frame_index;
+        return true;
+    }
+
+    void attach_zone_to_global_frame(ProfilerZoneToken& token)
+    {
+        uint64_t state = global_frame.load(std::memory_order_acquire);
+        while (global_frame_status(state) == GlobalFrameStatus::open) {
+            if (global_frame_zone_count(state) == 0x3fffffffu)
```
ccummingsNV: worth having constants for these numbers so its clearer what they are?

src/sgl/utils/profiler.cpp
```diff
@@ -0,0 +1,2240 @@
...
+    m_impl->capture_start_ns = now_ns();
+    m_impl->capture_stop_ns = 0;
+    m_impl->capture_state = ProfilerImpl::CaptureState::recording;
+}
+
+ref<ProfilerTrace> Profiler::stop_capture()
+{
+    tick();
+    flush();
+    std::lock_guard lock(m_impl->capture_mutex);
```
ccummingsNV: This (and some surrounding) seems a bit ugly. If you're calling a function on m_impl, and passing it loads of its own members, this suggests to me this code belongs in m_impl? I'd have thought most of these Profiler functions would be very minimal - locking the mutex then calling m_impl->bla()

ccummingsNV (dismissed):
A few style comments - nothing worth delaying a merge if you're wanting it in asap. Functionally looks solid.

src/sgl/utils/profiler.h
skallweitNV: That was the original name, but i think the predominant form is `enable_xxx` (for example on `DeviceDesc`).

PR #1075, opened 2026-07-27
Fix for texture loading exhausting sampler heap

ccummingsNV:
If cpu gets too far ahead of gpu in the parallel texture loading, sampler heap gets exhausted and device dies. This adds a cpu/gpu sync per batch of textures.

### DIFF
--- a/src/sgl/utils/texture_loader.cpp
+++ b/src/sgl/utils/texture_loader.cpp
@@ -260,15 +260,29 @@ inline SourceImage convert_bitmap(Device* device, ref<Bitmap> bitmap, const Text
     };
 }
 
-inline SourceImage load_source_image(const std::filesystem::path& path)
+inline SourceImage load_source_image(Stream* stream)
 {
     SourceImage source_image;
-    FileStream stream(path, FileStream::Mode::read);
-    if (DDSFile::detect_dds_file(&stream)) {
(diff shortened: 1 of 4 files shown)


src/sgl/utils/texture_loader.cpp
```diff
@@ -372,10 +369,12 @@ inline std::vector<ref<Texture>> create_textures(
         textures[i] = create_texture(device, blitter, command_encoder, source_images[i], options[i]);
         if (i && (i % BATCH_SIZE == 0)) {
             device->submit_command_buffer(command_encoder->finish());
+            device->wait();
```
tdavidovicNV, line 389: Does this mean we now always wait and don't frontload it at all?

Also, the slang-rhi has a fprintf to stderr. I don't know if that's the way we report errors there (I'd assume throwing was the way to go)

ccummingsNV: (removed slang-rhi change).

With textures, to be clear, this isn't the process of texture loading that now involves waits - it's the process of copying the loaded textures into gpu memory with a blit.

So it doesn't change whether we front load texture loading - it prevents the cpu getting too far ahead in dispatching those copies.

tdavidovicNV (dismissed):

tdavidovicNV (approved):

PR #928, opened 2026-04-09
Improve test coverage for torch integration

jhelferty-nv:
Improve overall test coverage from 73.8% to 82.8% (+9.0 pp), focusing on
torch integration paths on both the Python and C++ sides.

Python coverage: 78.7% → 82.9% (+4.2 pp, 264 newly covered lines)
C++ coverage: 69.2% → 82.7% (+13.5 pp, 1050 newly covered lines)

New and extended test files cover:

- DiffPair factory paths, property access, repr, and error guards
- TorchTensorMarshall creation, dtype conversion, and validation
- torch bridge fallback mode error paths
- torch autograd backward pass creating output DiffPairs
- torch raw dispatch and rare scalar type output (uint8, int8, int16, bool)
- torch benchmark smoke tests (slangpy_torch C API and libtorch paths)
- Tensor/TensorView/DiffTensorView forward paths and gradient handling
- ValueRef inout for scalars, vectors, and matrices (with padding)
- DispatchData error paths, writer callbacks, and cache hits
- Module load_from_source, require_struct, find_function_in_struct
- ByteAddressBuffer and RWByteAddressBuffer read/write
- StridedBufferView error guards, contiguity, and negative indexing
- PackedArg property access and repr
- C++ repr coverage for FunctionNode, SlangType, and texture shapes
- Image loading across channel counts and flip_y

Bug fixes included:

- Fix flip_y producing non-contiguous array in types/common.py
  (np.flipud result rejected by copy_from_numpy)

Bugs found and filed as separate issues:

- DescriptorMarshall hardcodes type to StructuredBuffer<Unknown>,
  preventing texture/sampler descriptor handles (#922)
- NativeTorchTensorDiffPair::read_signature bypassed by SignatureBuffer
  optimization in PR #872 (#923)

All tests avoid direct Native object construction where possible, using
Python-level APIs (diff_pair(), pack(), backward(), func.map()) to
exercise the underlying C++ code paths naturally.

Fixes #782
Fixes #783

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure image-loading returns C-contiguous arrays; improve CUDA interop synchronization and lifetime handling to prevent async-use and memory corruption.

* **Tests**
[...]

PR #761, opened 2026-01-27
Add comprehensive stride coverage tests for non-contiguous tensor inputs

jhelferty-nv:
Add tests verifying that non-contiguous PyTorch tensor views are correctly
handled when passed to Slang functions through the functional API. Covers
the marshalling and vectorization paths for multiple parameter types and
view operations.

Parameter types tested:
- float[3] fixed-size arrays (scaled_sum)
- float3 vectors (add_vectors)
- Tensor<float,2> / WTensor<float,2> (add_tensors)
- RWTensor<float,1> (copy_tensor)
- TensorView<float> (copy_tensorview, add_tensorview)
- DiffTensorView<float> (copy_difftensorview, diff_square backward)

View operations tested:
- Prefix slices (t[:3]), offset slices (t[1:4]), strided slices (t[::2])
- Transpose (.t()) — non-unit stride in binding dimension
- Diagonal (.reshape(3,3).diagonal()) — stride = sum of dim strides
- 3D permute + select (.permute(2,0,1)[0]) — both dims non-unit stride
- Zero-stride broadcast (.expand()) — stride 0 in batch dimension

Test categories:
- Forward correctness: GPU result matches PyTorch math
- Write-back correctness: sentinel-based verification that only target
  positions are written
- Gradient parity: SlangPy backward pass matches pure PyTorch autograd

Motivated by #387

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added `scaled_sum` function for scaled summation operations on value arrays.

* **Tests**
  * Added comprehensive test coverage for tensor slicing, transposition, and view operations.
  * Added verification for gradient parity with sliced and transposed tensor inputs.
  * Added tests for broadcasting behavior and PyTorch interoperability scenarios.

PR #1085, opened 2026-07-31
Consolidate test coverage from #761 and #928

ccummingsNV:
## Summary

- Merge the test coverage from #761 and #928 on top of current `main`.
- Preserve both sets of overlapping tensor, tensor-view, and PyTorch integration tests.
- Update the imported tests for current SlangPy APIs and expectations.

## Why

The two PRs overlap in several tensor and PyTorch fixtures and had drifted from current APIs. This consolidates their coverage into one conflict-resolved branch without changing runtime or SGL source code.

## Integration fixes

- Replace removed `NDBuffer` mixed-test usage with `Tensor`.
- Use the current `device_module` and `Tensor.empty` contracts.
- Remove obsolete direct `NativeSlangType` construction coverage.
- Skip optional PyTorch marshalling tests before importing the PyTorch-specific module when PyTorch is unavailable.
- Update the current `flip_y`, dtype-error, and zero-dimensional tensor expectations.

## Validation

- VS2022 17.10 / MSVC 19.40 x64 Release build through the Falcor2 superproject: passed.
- `python -m pytest slangpy/tests/slangpy_tests`: 2,649 passed, 236 skipped, 7 xfailed.
- `pre-commit run --all-files`: passed.

PyTorch is not installed in the validating repository environment, so PyTorch-specific runtime cases were skipped.

For completeness, the broader local `slangpy/tests` run had 10 failures in untouched baseline tests: five bitmap codec cases and five standalone `slangc` precompiled/standard-module lookup cases. Those failures occur outside this PR's 24-file test-only diff and before the integrated SlangPy test directory runs.


jhelferty-nv (dismissed):
I think it's fine for it to go ahead in. There's a bunch more that could be done, but I've delayed this getting in too long already. Thanks for pick this up Chris!

(For reference, there are a couple of things that an LLM prompt about maintainability will turn up. I recall dealing with this feedback to be a bit of a slog to get through last time I triaged it. It's probably better just to get it in and improve things as we go.)

PR #1090, opened 2026-08-03
Add Device::create_buffer_from_native_handle for buffer imports

fknfilewalker:
Wrap an existing native Buffer without copying, so a buffer allocated by another framework (torch) can be bound directly instead of round-tripping through host memory.

### DIFF
--- a/slangpy/tests/device/test_buffer_from_native_handle.py
+++ b/slangpy/tests/device/test_buffer_from_native_handle.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+import numpy as np
+
+import slangpy as spy
+from slangpy.testing import helpers
+
+
+# Native handle type accepted by Device.create_buffer_from_native_handle, per device type.
+# Device types missing from this table have no buffer import in slang-rhi.
+NATIVE_BUFFER_HANDLE_TYPES = {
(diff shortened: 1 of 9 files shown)


ccummingsNV (changes requested):
Looks solid. Is this Metal only, or will it work on other platforms (i.e. did slangrhi already support it on d3d/vk/cuda)?

If it is not implemented on other platforms we should probably throw not implemented.

Can we also get a test in? Even if its just creating a buffer, then getting its native handle and creating another buffer from it.

fknfilewalker:
> Looks solid. Is this Metal only, or will it work on other platforms (i.e. did slangrhi already support it on d3d/vk/cuda)?

It should be implemented for d3d/vk/webgpu. Cuda has no native handle implementation (I can add one in another pr) as well as cpu. This implementation is just for metal.
> Can we also get a test in? Even if its just creating a buffer, then getting its native handle and creating another buffer from it.

Done

fknfilewalker:
Tests fail because of slang-rhi [#813](https://github.com/shader-slang/slang-rhi/pull/813).

guoxx:
Fantastic work! I am also doing the exact same thing, and I can confirm that this implementation is feasible on Vulkan. This capability will greatly expand slangpy's ability to interact with other systems, for example, enabling slangpy to interact with NPUs through this capability.

skallweitNV:
@fknfilewalker I took the liberty to rebase/cleanup this PR so we can hopefully merge it.

PR #1094, opened 2026-08-07
Improve persistent cache robustness

skallweitNV:
- Use RAII for DB handle when creating LMDB database
- Make persistent cache creation more robust through exponential backoff when retrying
- Warn if shader cache cannot be created/initialized and run without it
- Remove deleting of the shader cache if it fails to initialize as this was not save across processes

### DIFF
--- a/src/sgl/core/lmdb_cache.cpp
+++ b/src/sgl/core/lmdb_cache.cpp
@@ -9,6 +9,8 @@
 
 #include <lmdb.h>
 
+#include <memory>
+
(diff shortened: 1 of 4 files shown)


jhelferty-nv:
**PR board sync:** auto-assigned @jkwak-work as shepherd for this Community PR.

Committer signal on the changed files is highest for ccummingsNV among collaborators other than the assignee — consider requesting a review from them. They were not auto-requested.

tdavidovicNV (approved):

PR #1099, opened 2026-08-11
Make nanothread a private SGL dependency

skallweitNV:
Route public task operations through exported SGL wrappers so downstream libraries no longer create separate nanothread default pools.

Hide nanothread task types behind opaque handles, use short_vector for dependency conversion, and share the same pool with slang-rhi. Remove the unused worker-ID API and limit tests to SGL integration behavior.


jhelferty-nv:

FYI for maintainers: committer signal on the changed files is highest for fknfilewalker among collaborators other than the assignee. They were not auto-requested; a human may optionally add them as a reviewer.

PR #1101, opened 2026-08-11
Fix logger deadlock with Python-backed outputs

skallweitNV:
Store logger outputs in immutable shared snapshots so logging only retains a native shared_ptr while holding the logger mutex. This prevents intrusive reference operations for Python-backed outputs from acquiring the GIL under the mutex.

Update output registration through copy-on-write snapshots constructed and destroyed outside the mutex. Retry publication after concurrent changes to avoid lost updates and prevent cleared or replaced outputs from being resurrected.

Make the log level atomic so filtered messages avoid mutex acquisition, and skip formatting for filtered formatted messages. Also make use_same_outputs() synchronized and safe for self-assignment.

Add coverage for callback re-entry, snapshot reference counts, removal during active callbacks, concurrent output mutations, snapshot sharing, and lazy formatting.

### DIFF
--- a/src/sgl/core/logger.cpp
+++ b/src/sgl/core/logger.cpp
@@ -179,6 +179,7 @@ std::string DebugConsoleLoggerOutput::to_string() const
 Logger::Logger(LogLevel log_level, const std::string_view name, bool use_default_outputs)
     : m_level(log_level)
     , m_name(name)
+    , m_outputs(std::make_shared<OutputSet>())
 {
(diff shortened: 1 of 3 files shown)


jhelferty-nv:

FYI for maintainers: committer signal on the changed files is highest for ccummingsNV among collaborators other than the assignee. They were not auto-requested; a human may optionally add them as a reviewer.

PR #1109, opened 2026-08-14
Add support for printing enum values

skallweitNV:
Adds `IPrintable` support for Slang enum types.

Enum values retain the printable representation of their underlying type rather than being coerced to `int32`. This includes signed, unsigned, 64-bit, and boolean-backed enums.

## Implementation

- Adds a generic `__EnumType` conformance to `IPrintable`.
- Derives the printable layout and data width from the enum’s underlying `T.__Tag` type.
- Preserves 64-bit values using two printable data words.
- Includes the preferred forwarding implementation as a TODO.

Directly forwarding through `T.__Tag : IPrintable` currently triggers an incorrect recursion diagnostic when invoked through an `IPrintable` interface value. The implementation uses a temporary workaround tracked by shader-slang/slang#12540.


jhelferty-nv:

PR #1115, opened 2026-08-18
Fix thread safety issues

skallweitNV:
## Summary

Fix thread-safety and lifetime-ordering issues found by ThreadSanitizer.

- Synchronize intrusive reference-count decrements so cross-thread final release observes writes made by previous object owners.
- Ensure slang-rhi task payload deletion completes before the task is reported as complete.
- Protect `FileSystemWatcher::m_last_event` with the queued-events mutex.
- Give each LMDB stress-test thread its own RNG to remove shared-state races.
- Add regression tests for cross-thread object destruction and task payload cleanup ordering.

## Rationale

`Object::dec_ref()` previously used only relaxed atomic operations. When the final reference was released on another thread, there was no happens-before relationship ensuring that the destructor observed writes made by earlier owners.

The nanothread backend also publishes task completion before invoking its payload deleter. This conflicts with the slang-rhi task-pool contract and allowed `waitAndReleaseTask()` to return while payload cleanup was still running. The adapter now performs user payload cleanup within the task callback, before nanothread can publish completion.

### DIFF
--- a/src/sgl/core/file_system_watcher.cpp
+++ b/src/sgl/core/file_system_watcher.cpp
@@ -213,8 +213,8 @@ void FileSystemWatcher::_notify_change(
         std::lock_guard<std::mutex> lock(m_queued_events_mutex);
 #endif
         m_queued_events.push_back(event);
+        m_last_event = now;
     }
-    m_last_event = now;
 }
 
 void FileSystemWatcher::update()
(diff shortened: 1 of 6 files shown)


jhelferty-nv:

FYI for maintainers: committer signal on the changed files is highest for ccummingsNV among collaborators other than the assignee. They were not auto-requested; a human may optionally add them as a reviewer.

PR #1119, opened 2026-08-20
Thread sanitizer

skallweitNV:
## Summary

Adds ThreadSanitizer build support and continuous testing for SlangPy’s native C++ code on Linux and macOS.

## Changes

- Add the `SGL_ENABLE_TSAN` CMake option.
- Apply TSan instrumentation to SlangPy targets and `nanothread`.
- Reject unsupported compiler, platform, and ASan/TSan combinations.
- Disable doctest’s POSIX signal handler for Linux TSan builds to avoid conflicts with TSan’s signal and `fork()` interceptors.
- Extend the sanitizer environment tooling to configure the TSan runtime and `TSAN_OPTIONS`.
- Add TSan CI jobs for:
  - Linux x86-64 using LLVM 22, with device tests skipped.
  - macOS arm64, including device tests.
- Continue after individual TSan reports while returning a failing exit code when races are detected.
- Keep sanitizer runtimes local to the instrumented test process so they are not injected into GitHub Actions’ Node-based post-job steps.

## LLVM 22 support

Linux uses LLVM 22 instead of the runner’s LLVM 18 because LLVM 18 can expose a partially initialized thread to TSan interceptors.

A temporary `libdeflate` vcpkg overlay is included to support this toolchain. It:

- Preserves externally supplied release compiler flags, including sanitizer instrumentation.
- Restricts the `evex512` workaround to compiler versions where it is supported and required.

## Scope

The TSan jobs build and run the native C++ test suite. Python bindings and examples are disabled for these jobs. Linux device tests are skipped through the `-skip-device-tests` option because the GPU stack is outside the supported TSan scope.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -174,3 +174,100 @@ jobs:
           name: sanitizer-reports-${{ matrix.os }}-${{ matrix.platform }}
           path: build/${{ matrix.preset }}/${{ matrix.config }}/sanitizer-logs/
           if-no-files-found: ignore
+
+  tsan:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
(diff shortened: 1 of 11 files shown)


jhelferty-nv:

Auto-assigned @kaizhangNV as shepherd for this Community PR.

FYI for maintainers: committer signal on the changed files is higher for ccummingsNV than for the auto-requested reviewer (kaizhangNV). They were not auto-requested; a human may optionally add them as a reviewer.

PR #1124, opened 2026-08-26
Fix profiler collector ordering races

### LINKED ISSUE #1072: Profiler can finalize a global frame before consuming attached cross-thread zones
## Summary

The profiler collector can finalize a global frame before consuming a zone that was attached to that frame and already published by another thread. This makes frame statistics nondeterministic and caused the macOS ARM64 Debug CI failure below.

Failed job: https://github.com/shader-slang/slangpy/actions/runs/30059554766/job/89378240354

SlangPy commit: `5a1b34b63a78904aeb092bc6475744edc500a83c`

## Observed failure

The C++ test `global frames include zones from other threads and exclude unframed zones` failed at `tests/sgl/device/test_profiler.cpp:371-373`:

```text
CHECK( stats->entries().front().cpu_time_per_call.count == 2 )
values: CHECK( 1 == 2 )

CHECK( stats->sample(1).call_count.front() == 1 )
values: CHECK( 0 == 1 )
```

The two expected per-frame zone counts are `[1, 1]`, but the collected result is `[0, 1]`. The worker begins its zone and signals `worker_started` before the main thread closes frame 0, so the zone is attached to frame 0 by design.

Only this test failed: 197 test cases passed, one failed, and one was skipped.

## Suspected root cause

`ProfilerImpl::drain()` currently:

1. snapshots and drains the per-thread CPU event queues;
2. consumes those events;
3. snapshots `sealed_frame_events`;
4. consumes the frame markers.

See `src/sgl/utils/profiler.cpp` around lines 739-765.

There is a race between the two snapshots:

1. The collector snapshots a worker queue before the worker publishes its zone.
2. The worker publishes the zone to its queue.
3. The worker releases the final attached-zone reference and seals frame 0.
4. The collector snapshots the sealed-frame list and consumes the frame marker.
5. `consume(frame)` calls `finalize_ready_frames()`, completing frame 0 without the worker zone.
6. The next drain consumes the worker zone, but frame 0 has already been removed from `pending_frames`.

This ordering produces exactly the observed `[0, 1]` result. Calling `flush()` after joining the worker cannot repair the already-finalized frame.

## Proposed fix

Snapshot `sealed_frame_events` before snapshotting the per-thread queues, while continuing to consume the per-thread zone events before consuming the captured frame markers.

A frame is sealed only after all attached zones have published and released their references. Acquiring the sealed-frame mutex first, followed by the queue write-index acquire loads, should therefore ensure that every zone associated with a captured frame marker is included in the same drain. If a frame is sealed after the sealed-frame snapshot, both its marker and any not-yet-snapshotted zone can safely wait for the next drain.

The existing cross-thread test should remain: it directly exercises the intended contract documented by `end_frame()`, where the collector seals a frame only after all attached zones publish.

skallweitNV:
## Summary

Fixes #1072.

Profiler inputs were collected from independent CPU-event, frame-seal, and GPU-result channels. A collector pass could therefore observe a completion record after taking an older snapshot of its prerequisite CPU events. This could prematurely finalize a frame without all of its zones, causing nondeterministic CI failures.

## Changes

- Take one causal ingress cut across CPU queues, frame seals, and GPU results.
- Process each batch in dependency order: CPU events, GPU results, frame seals, then frame settlement.
- Replace per-thread flush targets with generation-based collector-pass acknowledgement.
- Document that `flush()` waits for already-published inputs, but not active zones, unsealed frames, or unresolved GPU queries.
- Re-enable the cross-thread frame and device-close regression tests.
- Add coverage for concurrent flush requests.

Normal CPU-zone recording remains lock-free and gains no additional per-zone bookkeeping.

## Validation

- Profiler tests: 18/18 passed, 186 assertions
- Full C++ suite: 266/266 passed, 21,002 assertions
- 200 repeated runs of both CPU concurrency regressions
- All pre-commit hooks passed

### DIFF
--- a/src/sgl/utils/profiler.cpp
+++ b/src/sgl/utils/profiler.cpp
@@ -452,6 +452,11 @@ struct ProfilerImpl {
         bool missing{false};
     };
 
+    struct CpuDrainTarget {
+        ThreadData* thread{nullptr};
+        uint64_t write_index{0};
+    };
+
     struct GpuSlot {
         uint64_t correlation_id{0};
         uint64_t parent_correlation_id{0};
@@ -555,7 +560,7 @@ struct ProfilerImpl {
         const ThreadID thread_id = platform::current_thread_id();
         ThreadData* result = data.get();
         {
-            std::lock_guard lock(thread_mutex);
+            std::lock_guard lock(ingress_mutex);
             data->timeline_id = uint32_t(timelines.size());
             ProfilerTimeline timeline;
             timeline.type = ProfilerTimelineType::cpu;
@@ -571,7 +576,7 @@ struct ProfilerImpl {
     uint64_t producer_drop_count() const
     {
         uint64_t result = 0;
-        std::lock_guard lock(thread_mutex);
+        std::lock_guard lock(ingress_mutex);
         for (const auto& thread : threads)
             result += thread->drop_count.load(std::memory_order_relaxed)
                 + thread->stack_overflow_count.load(std::memory_order_relaxed);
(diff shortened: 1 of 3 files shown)


jhelferty-nv:
Closing and reopening to re-trigger CI

tdavidovicNV (approved):

PR #1151, opened 2026-09-12
Update LMDB to 0.9.36

tdavidovicNV:
## Summary

- replace the vendored `mdb.master` development snapshot with the exact library sources from the stable [`LMDB_0.9.36`](https://github.com/openldap/openldap/releases/tag/LMDB_0.9.36) tag
- keep the change limited to `lmdb.h`, `mdb.c`, `midl.c`, and `midl.h`; no SGL integration or cache policy changes

## Motivation

A downstream Windows CI run hit rare access violations in `mdb_page_search_root` while several processes shared a shader cache. This does not prove that LMDB itself caused the corruption, but the currently vendored snapshot is from `mdb.master` at `e13f05211f8c0869fe1475cd632ba59d4f957656`, identifies itself with the development placeholder version 0.9.70, and predates the fix for [ITS#9564](https://lists.openldap.org/hyperkitty/list/openldap-bugs%40openldap.org/thread/EI3DFZERCS6OQEHAF7WHAYTGUHFXG5JZ/), a real race when freeing spilled pages after a write transaction. LMDB 0.9.36 contains that fix.

There is also a [Windows corruption report against the development line](https://lists.openldap.org/hyperkitty/list/openldap-bugs%40openldap.org/thread/ECXK36FNUBYW6M7656MNJSQWUEN3JL4A/) where upstream explicitly recommends using the stable release branch. Moving to 0.9.36 removes both the known race and the unsupported development-only Windows code path. It is a plausible mitigation for the observed crash, not a claim that its root cause is proven.

## Compatibility

Both snapshots use LMDB data format version 1, so existing `data.mdb` files remain compatible. Their lock-file formats differ; LMDB reinitializes the lock file when startup has exclusive ownership and returns `MDB_VERSION_MISMATCH` if old and new binaries overlap. No cache-path migration is required.

## Validation

- Falcor2 Windows Release build passed
- synchronized cache stress passed: 10 waves x 4 processes x 2 device threads x 2 reopen cycles x 4 pipelines, all sharing one cache
- all four vendored file blob hashes match the 0.9.36 tag
- SlangPy pre-commit passed (vendored sources are excluded by configuration)


skallweitNV (approved):
LGTM

PR #1159, opened 2026-09-14
Add explicit luminance handling to texture loading

ccummingsNV:
Add `TextureLoader.Options.y_handling` with `YHandling.preserve_as_r` (the default) and `YHandling.expand_to_rgba`, following the existing `YAHandling` API.

The default preserves historical single-channel Y loading, including when `load_as_srgb` is true. Explicit expansion replicates luminance into RGB and fills alpha while preserving component type and gamma encoding; normal sRGB/normalized format selection then applies. Explicit R/RG/RGB/RGBA inputs and the separate `YAHandling` policy retain their existing behavior.

Merged upstream main at `2e882fe7`. The sole added test is `test_expand_luminance_to_rgba`, covering one-bit and 8-bit PNGs through bitmap and file loading. All six cases pass across D3D12, Vulkan and CUDA using the validated Windows Release build. Full pre-commit checks pass.


src/slangpy_ext/utils/texture_loader.cpp
```diff
@@ -46,6 +47,12 @@ SGL_PY_EXPORT(utils_texture_loader)
             D(TextureLoader, Options, load_as_normalized)
         )
         .def_rw("load_as_srgb", &TextureLoader::Options::load_as_srgb, D(TextureLoader, Options, load_as_srgb))
+        .def_rw(
+            "load_as_rgb",
+            &TextureLoader::Options::load_as_rgb,
+            "Expand luminance-only bitmaps to RGBA without changing component type or gamma encoding. "
```
skallweitNV, line 53: Should use the `D` macro

src/sgl/utils/texture_loader.h
```diff
@@ -36,8 +36,12 @@ class SGL_API TextureLoader : public sgl::Object {
     struct SGL_API Options {
         /// Load 8/16-bit integer data as normalized resource format.
         bool load_as_normalized{true};
-        /// Use \c Format::rgba8_unorm_srgb format if bitmap is 8-bit RGBA with sRGB gamma.
+        /// Use \c Format::rgba8_unorm_srgb for 8-bit color bitmaps with sRGB gamma.
+        /// Luminance-only color bitmaps are expanded to RGBA.
         bool load_as_srgb{true};
+        /// Expand luminance-only bitmaps to RGBA for consumers that require RGB channels.
+        /// Preserves the component type and gamma encoding; explicit R/RG bitmaps are unchanged.
+        bool load_as_rgb{false};
```
skallweitNV, line 44: I find this confusing. Should this be named `load_luminance_as_rgba` or `load_y_as_rgba` or similar? Does it convert to RGB or RGBA? It looks like this also changes the meaning of `load_as_srgb`?

skallweitNV (changes requested):

src/sgl/utils/texture_loader.h
ccummingsNV: yup - wip :)

tdavidovicNV (dismissed):

src/slangpy_ext/utils/texture_loader.cpp
```diff
@@ -50,6 +55,7 @@ SGL_PY_EXPORT(utils_texture_loader)
         .def_rw("allocate_mips", &TextureLoader::Options::allocate_mips, D(TextureLoader, Options, allocate_mips))
         .def_rw("generate_mips", &TextureLoader::Options::generate_mips, D(TextureLoader, Options, generate_mips))
         .def_rw("usage", &TextureLoader::Options::usage)
+        .def_rw("y_handling", &TextureLoader::Options::y_handling, D(TextureLoader, Options, y_handling))
         .def_rw("ya_handling", &TextureLoader::Options::ya_handling);
```
skallweitNV: we could add docstrings for all of them now

src/sgl/utils/texture_loader.h
skallweitNV, line 38: nitpick: the ordering is reverse for the two enums

skallweitNV: would be easier on the eyes with a trailing `,` which `clang-format` will handle by having one item per line. same applies to the existing enum

skallweitNV (approved):

PR #1164, opened 2026-09-16
Let another project reuse the BenchView submission path

expipiplus1:
allow specifying the project name etc, so other pytest projects can reuse the benchview functionality here


slangpy/testing/benchmark/benchview.py
```diff
@@ -28,12 +29,34 @@
...
+@dataclass(frozen=True)
+class BenchViewProject:
+    """Identity BenchView files a submission under.
+
+    BenchView keys a project's history by these values, so another repository
+    reusing this module must supply its own. Leaving them at the defaults would
+    file its results under SlangPy.
+    """
+
+    id: str = BENCHVIEW_PROJECT_ID
```
ccummingsNV: I'd advise against defaulting class members to values for a specific project - this is a recipe for accidently submitting 1 set of data to the wrong place. Generally if this is something a project needs to specify to get correct, it should be an error if it is not specified.

slangpy/testing/benchmark/benchview.py
```diff
@@ -257,6 +315,7 @@ def _build_submission_base(
     execution_id: str,
     project_info: dict[str, Any],
     commit_info: dict[str, Any],
+    project: BenchViewProject = SLANGPY_PROJECT,
```
ccummingsNV: Same point as above - defaults are good when you genuinely want 'forgot to specify' to mean 'use the default' - that's not what you want here

ccummingsNV (changes requested):
I'd suggest changing the 'defaults' before merging this or we'll regret it at some point in the future :)

PR #1168, opened 2026-09-18
Optional GLFW

skallweitNV:
Allows building SlangPy without GLFW or X11 development headers while retaining compute and offscreen rendering.

Adds `SGL_ENABLE_GLFW` with three modes:

- `AUTO` (default): enables GLFW when build dependencies are available.
- `ON`: requires GLFW support; missing dependencies fail configuration.
- `OFF`: omits GLFW entirely.

Without GLFW, the `Window` API remains available but construction raises a clear error, and the display scale defaults to `1.0`. The build option is documented in the compiling guide.

Updates slang-rhi to include Vulkan Xlib declarations that remove its dependency on X11 headers while preserving native-handle surface support. Also adds mappings for its new cooperative matrix features.

Validation:
- Windows builds with GLFW enabled and disabled.
- D3D12 and Vulkan compute/offscreen rendering checks.
- Linux GLFW dependency detection and option validation.
- Linux slang-rhi build without X11 headers and Vulkan/CUDA compute tests.


ccummingsNV (approved):
LGTM

